PR #2, opened 2015-11-21
Add authorization and parametrize nick names.

Rotwang:
I'm trying to use mattermost as a replacement for znc.

### DIFF
--- a/config.go
+++ b/config.go
@@ -13,7 +13,9 @@ type Config struct {
 		Server        string
 		Port          int
 		Nick          string
+		Password      string
 		Channel       string
+		SendMUserName bool
 	}
 	Mattermost struct {
 		URL           string
@@ -22,6 +24,7 @@ type Config struct {
 		Token         string
 		IconURL       string
 		SkipTLSVerify bool
+		IrcNickPrefix string
 	}
 	General struct {
 		GiphyAPIKey string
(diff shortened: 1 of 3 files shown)


config.go
```diff
@@ -13,7 +13,9 @@ type Config struct {
 		Server        string
 		Port          int
 		Nick          string
+		Password      string
 		Channel       string
+		SendMUserName bool
```
42wim: A bool is false if uninitialized, so the keep backwards compatibility with existing configs, better make this the inverse. Like DisableMmUserName

config.go
```diff
@@ -22,6 +24,7 @@ type Config struct {
 		Token         string
 		IconURL       string
 		SkipTLSVerify bool
+		IrcNickPrefix string
```
42wim: Same here, for compatibility, could you make the default "irc-" if IrcNickPrefix is ""

matterbridge.go
```diff
@@ -52,11 +53,11 @@ func (b *Bridge) handlePrivMsg(event *irc.Event) {
 		msg = event.Nick + " "
 	}
 	msg += event.Message()
-	b.Send("irc-"+event.Nick, msg)
+	b.Send(b.Config.Mattermost.IrcNickPrefix+event.Nick, "PRIV: "+msg)
```
42wim: "PRIV: " should not be send when doing CTCP_ACTION (/me)

matterbridge.go
```diff
@@ -89,10 +90,19 @@ func (b *Bridge) handleMatter() {
 		case "!gif":
 			message.Text = b.giphyRandom(strings.Fields(strings.Replace(message.Text, "!gif ", "", 1)))
 			b.Send(b.Config.IRC.Nick, message.Text)
+		case "!priv":
+			who := strings.Fields(message.Text)[1]
```
42wim: Need to check if len(result) of strings.Fields == 2 otherwise you can get nil pointer panics

42wim:
Thanks for the PR! 
I added some comments, could you check those and rebase ?

PR #8, opened 2016-01-27
Add PASS support as per RFC1459

daysofwineandroses:
Provide a connection password via the protocol's PASS command.

Imported irc.go supports it as a simple parameter:
https://github.com/thoj/go-ircevent/blob/master/irc.go#L381

See https://tools.ietf.org/html/rfc1459#section-4.1 for full details.


42wim:
Thanks!

PR #13, opened 2016-03-18
Enable !users command and add a table formatter

fdevibe:
* Listen to wildcarded events, currently handle RPL_NAMREPLY.
* Add a formatter to present nick lists from IRC as a table.


42wim:
Thanks for your work. I'd like to merge it, but I think the config fits better under [mattermost] instead of [irc]
(Also the case with your previous PR)

fdevibe:
@42wim sure, if that's all, it should be quickly done. :smile:

42wim:
Ok, have been testing it now ;-) 
The header also shows NicksPerRow columns
image

Is prettier if that's not the case ;-)
image

fdevibe:
Closed in favor of #14.

PR #15, opened 2016-03-21
Handle connects asynchroneously.

fdevibe:
On connect events, update the nick with what's actually used and
set up channels as configured.


42wim:
Thanks! Much appreciated

PR #16, opened 2016-03-21
Add Freenode style NickServ-based authentication.

fdevibe:
If the server sends a NOTICE requesting identification, identify.


42wim:
Thanks again :+1:

PR #17, opened 2016-03-22
Gather all names before printing them.

fdevibe:
When issuing a NAMES request, the reponse is chunked. Cache names
until all are received, then print them out.

### DIFF
--- a/matterbridge.go
+++ b/matterbridge.go
@@ -9,6 +9,7 @@ import (
 	"github.com/thoj/go-ircevent"
 	"strconv"
 	"strings"
+	"sort"
 )
 
 type Bridge struct {
@@ -17,6 +18,7 @@ type Bridge struct {
 	cmap        map[string]string
 	ircNick     string
 	ircNickPass string
+	names       []string
 	*Config
 }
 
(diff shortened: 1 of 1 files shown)


fdevibe:
My Go experience is pretty much limited to my patches to this project. My array / slice handling here seems somewhat dysfunctional to me, any hints on how to improve this (or just fixing it) would be much appreciated.

matterbridge.go
```diff
@@ -25,6 +27,7 @@ func NewBridge(name string, config *Config) *Bridge {
 	b.Config = config
 	b.cmap = make(map[string]string)
 	b.ircNick = b.Config.IRC.Nick
+	b.names = make([]string, 0, 10000)
```
42wim: This is not necessary for our use case, we don't need to allocate this much. We can just use the append function to grow dynamically.

matterbridge.go
```diff
@@ -131,13 +133,24 @@ func (b *Bridge) formatnicks(nicks string) string {
...
 }
 
+func (b *Bridge) storeNames(event *irc.Event) {
+	b.names = append(b.names, strings.Split(event.Message(), " ")...)
+}
+
+func (b *Bridge) endNames(event *irc.Event) {
+	sort.Strings(b.names)
+	b.Send(b.ircNick, b.formatnicks(b.names), b.getMMChannel(event.Arguments[0]))
+	b.names = make([]string, 0, 10000)
```
42wim: This can be replace by b.names = nil

42wim:
Looks pretty good, I'm no expert myself though.
Gave you some notes how I would change those 2 lines :-)

fdevibe:
@42wim thanks for the input. From the Go docs, it wasn't clear to me it could be done this way. Pushed now. :-)

PR #81, opened 2016-11-15
Add SkipTLSVerify in xmpp

jadjay:
Thanks


42wim:
Thanks!
But this disable TLSverify for everyone, it should be configurable like the rest. 
Could you fix this and also update matterbridge.toml.sample with the new parameter ?

PR #93, opened 2016-11-25
Allow PrefixMessagesWithNick on Telegram

zoni:
This makes the PrefixMessagesWithNick setting work for messages relayed
to Telegram, similar to how they work for messages relayed to mattermost
(it is in fact copied from the mattermost bridge implementation).

I believe this logic should be moved up the stack and not left to
individual bridge implementations, but adding it here like this made it
quickly work for my use-case (I don't have the time to dig in deeper and
refactor everything).

### DIFF
--- a/bridge/telegram/telegram.go
+++ b/bridge/telegram/telegram.go
@@ -57,7 +57,12 @@ func (b *Btelegram) Send(msg config.Message) error {
 	if err != nil {
 		return err
 	}
-	m := tgbotapi.NewMessage(chatid, msg.Text)
+
+	message := msg.Text
+	if b.Config.PrefixMessagesWithNick {
+		message = msg.Username + " " + message
+	}
+	m := tgbotapi.NewMessage(chatid, message)
 	_, err = b.c.Send(m)
 	return err
 }


42wim:
Thanks, but ```RemoteNickFormat="<{NICK}> "``` should just work. (this logic is up the stack). 
Could you let me know if this works for you ?

```PrefixMessagesWithNick``` is specially for mattermost and slack where there is a webhook username override option and you still want to prefix the messages with the nick.

**Comment by zoni** [comment]
That's odd. I'll take another look at it somewhere over the next few days when I have time again, but, I dove into the code and lifted that from elsewhere specifically because `RemoteNickFormat="{NICK} says: "` wasn't working for me.

My config (which failed to work as expected without the changes on this PR) consists of:

```
# Note: Some sensitive items have been replaced, but this config is otherwise identical to what is really used

[mattermost]
[mattermost.myteamname]
useAPI=true
Server="mattermost.nick.groenen.me"
Team="myteamname"
Login="secret"
Password="verysecret"
PrefixMessagesWithNick=true
RemoteNickFormat="{NICK} says: "

[telegram]
[telegram.myteamname]
Token="verysecret"
PrefixMessagesWithNick=true
RemoteNickFormat="{NICK} says: "

[[gateway]]
enable=true

    [[gateway.inout]]
    account="telegram.myteamname"
    channel="-14488271"

    [[gateway.inout]]
    account="mattermost.myteamname"
    channel="telegrams"
```

**Comment by 42wim** [comment]
For ```mattermost.myteamname``` the ```PrefixMessagesWithNick=true``` shouldn't be necessary because you have ```useAPI=true```. You should see the nick double with that config.

The telegram config looks fine, should work. I don't have a telegram account, but I've got a bot token to test, and I see the username getting echoed with the modification.

Could you run with -debug ? 
You should see something like this:

```
DEBU[2016-11-25T23:48:16+01:00] Sending config.Message{Text:"a", Channel:"-123456789123", Username:"wim", Avatar:"", Account:"irc.test", Event:""} from irc.test (#telegramtest2) to telegram.test (-123456789123)
DEBU[2016-11-25T23:48:16+01:00] Receiving config.Message{Text:"a", Channel:"-123456789123", Username:"wim says: ", Avatar:"", Account:"irc.test", Event:""}  module=telegram
```

ishitatsuyuki:
Noticed a flaw that you have never took the username variable into use. Go does make an error on unused local ones, but not struct ones :stuck_out_tongue:

zoni:
@42wim said:
> For mattermost.myteamname the PrefixMessagesWithNick=true shouldn't be necessary because you have useAPI=true. You should see the nick double with that config.

I actually don't. It's not mimicking the user on Telegram, just using the username of the account it's associated with (`telegram` in my case):

screenshot from 2016-12-14 16-05-33

@ishitatsuyuki's comment likely explains that though:
> Noticed a flaw that you have never took the username variable into use. Go does make an error on unused local ones, but not struct ones

As for the rest of the issue, I still haven't had the time to properly sit down and try some variations, I hope to be able to have the time for it soon.

ishitatsuyuki:
#102, this was simply a bug, and doesn't require an option addition.

PR #102, opened 2016-12-17
Telegram: Fix the missing username

ishitatsuyuki:


ishitatsuyuki:
Added markdown support which depends on this fix.

42wim:
Thanks!
The username fix is a no-brainer for merge, but I'm not so sure about the markdown addition.
I think this should be fixed by regexp, so we are more certain we're not changing random * or _ in sentences ?
Would you care to split this up in 2 PR's ? one for the username and one for the markdown support ?
Thanks again.

ishitatsuyuki:
Done. Don't forget to close #93.

PR #103, opened 2016-12-17
Telegram: add markdown

ishitatsuyuki:
Telegram has the non-standard "markdown" style. This ugly replacement logic is hard to understand but works properly. Tested.

### DIFF
--- a/bridge/telegram/telegram.go
+++ b/bridge/telegram/telegram.go
@@ -1,10 +1,14 @@
 package btelegram
 
 import (
+	"bytes"
+	"html"
+	"strconv"
+
 	"github.com/42wim/matterbridge/bridge/config"
 	log "github.com/Sirupsen/logrus"
 	"github.com/go-telegram-bot-api/telegram-bot-api"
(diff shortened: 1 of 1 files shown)


42wim:
Thanks, but I think this should be fixed by regexp, so we are more certain we're not changing random * or _ in sentences ?

ishitatsuyuki:
Oh. Good point, will find another method with regexp.

ishitatsuyuki:
I'm very lazy, so I decided to use a mature parser. This works perfectly, with many syntax extensions ported.

42wim:
Thanks for the work, I've got one question though, why not using the default BlockCode and Paragraph ?

ishitatsuyuki:
I'm blacklisting the tags that Telegram doesn't parse. `<p>` and `<code>` will cause an error resulting in the message not delivered.

ishitatsuyuki:
Please wait for an update to also blacklist "quote".

> this is a quote

42wim:
Thanks!

PR #106, opened 2017-01-04
Handle SkipTLSVerify for XMPP client

fgrsnau:
This PR is similiar to #81 but makes the SkipTLSVerify option configurable at runtime.

This makes it possible to use matterbridge with self-signed certificates. Actually it would be better to specify custom trust anchor, but in the meantime this works if used in a trusted network.


42wim:
Thanks! 
Could you put the documentation changes in ```matterbridge.toml.sample``` ? 
matterbridge.conf.sample is obsolete (and I just removed it). 
Please put it after ```Nick``` and before ```IgnoreNicks``` please :)

fgrsnau:
Okay, should be at the right location now.

42wim:
Thanks!

PR #114, opened 2017-01-18
Fix tls in xmpp by setting ServerName

janza:
Fixes error message shown by tls: "either ServerName or InsecureSkipVerify must be specified in the tls.Config"


42wim:
Thanks for the fix!

PR #157, opened 2017-05-04
[TO CLOSE] Add http.ProxyFromEnvironment to use default env var like in net/http…

ryarnyah:
Add standard http client proxy possibility to matterircd like DefaultTransport in net/http/transport.go.

Ref PR https://github.com/42wim/matterircd/pull/95


ryarnyah:
Edit: I'm sorry, i wasn't testing with the right env var.... this PR and https://github.com/42wim/matterircd/pull/95 can be closed.

PR #162, opened 2017-05-06
Add http.ProxyFromEnvironment to use default env var like in net/http…

ryarnyah:
Cf https://github.com/42wim/matterircd/pull/95


42wim:
Thanks!

PR #172, opened 2017-05-21
Fix JoinChannel argument to use IRC channel key

siinus:
JoinChannel is called with channel.Name, although mychannel is already defined and IRC channel needs key.


42wim:
Thanks, good catch!

Issue #116, opened 2017-01-20
Status monitoring


jurgenhaas:
I was wondering what's the best way to monitor the status of matterbridge. The docker container is one thing to look at but I'd also be interested about the status of the services that matterbridge is configured for.

Is it feasible to get some sort of an end-point in matterbridge that responds with a json summary of the current status which could then be used by monitoring tools?

42wim:
I've been thinking about this.

Almost all of the libraries I use reconnect automatically on failures and most of them can't be asked about the current state. So this would only be possible to implement in my own (mattermost) library. 

In the last release, irc and mattermost reconnects should be more stable. So services should always be connected or being busy trying to connect.

So unless every underlying library add a way to ask about their state, this isn't feasible.

jurgenhaas:
Understood, this is not a simple one but I doubt that the other services would come up with status APIs without letting them know about the requirement.

What I've done in the meantime is to use the fluentd log driver of docker and now monitor the logs instead. I do get an error like this when Mattermost goes down:

```
time="2017-02-20T08:38:14Z" level=error msg="error:websocket: close 1006 (abnormal closure): unexpected EOF" module=matterclient
```

This is great and helps me to identify downtime. What's missing however is a log entry when Mattermost comes back and matterbridge reconnects. Would you mind putting something into the log even when not in debug mode?

Infernoman:
Hey just wanted to post an update on this as well. Let me know if i should open a new issue. matterbridge has been crashing on me recently. and I've been looking into ways to monitor the application. I tried supervisor. Which i had no luck with. And found that forever.js isn't supported either. So I looked into docker. and had some issues as well.

I'm using the same configuration file for running it inside the go/bin folder but when i run it inside docker messages from mattermost don't get relayed to slack/irc/discord. but messages from slack/irc/discord get relayed to mattermost just fine.

When matterbridge is run by itself. messages relay between all 4 of the applications correctly.

42wim:
@Infernoman Please reopen a new issue if you experience crashes. I'll try to fix those :)
Be sure you're running the latest version and put the stack trace after a panic in the issue.

For restarting systemd is a good option, you could do something like this example service below.

```
[Service]
ExecStart=/opt/matterbridge/matterbridge-linux64 -conf /opt/matterbridge/matterbridge.conf  
Restart=always  
StandardOutput=syslog  
StandardError=syslog  
SyslogIdentifier=matterbridge  
User=matterbridge  
[...]

PR #176, opened 2017-05-23
Matterbridge as a linux systemd service

realrolfje:
Add scripts to have matterbridge run as a systemd daemon in Linux.

### DIFF
--- a/init.d/README.md
+++ b/init.d/README.md
@@ -1,0 +1,24 @@
+Matterbridge as a Service in Linux
+==================================
+
+To install Matterbridge as a service in Linux, place matterbridge and the config file in the `/opt/matterbridge` directory, like so:
+
+```
+sudo mkdir -p /opt/matterbridge
+sudo cp matterbridge-linux64 /opt/matterbridge
+sudo cp matterbridge.toml /opt/matterbridge
+```
+
+Then, install the [matterbridge deamon script](matterbridge) service into `/etc/init.d/` like so (description for Debian like installations):
+
+```
+sudo cp matterbridge /etc/init.d/
+sudo chmod +x /etc/init.d/matterbridge
+sudo systemctl daemon-reload
+sudo systemctl enable matterbridge
+sudo systemctl start matterbridge
+```
+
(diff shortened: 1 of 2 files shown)


42wim:
Hi, thanks for the PR!
It seems like a mix between a sysv initscript and systemd though?

realrolfje:
Could be. I cobbled it together based on the mattermost init.d script, I'm no expert. We needed it because when a connection drops, de bridge seemed to stop working. I decided that having it as a service would be nice, so we can do an easy restart, while also having the benefit of automatic starting at boot.

If you have tips or pointers on how to improve the script I'd be happy to improve this pull request.

42wim:
An example for a true systemd service unit can be found here
https://github.com/42wim/matterbridge/issues/116#issuecomment-299642840

Normally you put those files in /etc/systemd/system 

The daemon-reload, enable and start commands remain the same.

heinrich5991:
I have this:

``` 
[Unit]
Description=matterbridge
After=network.target

[Service]
ExecStart=/usr/bin/matterbridge -conf /etc/matterbridge/bridge.toml
User=matterbridge
Group=matterbridge

[Install]
WantedBy=multi-user.target
```

Should I make a pull request? It's helpful when projects have some standard service files. :)

42wim:
@heinrich5991 looks good, sure, go ahead :)

patcon:
FYI, this is something related, which you might consider directing folks toward instead of including init scripts :)
https://github.com/jordansissel/pleaserun

patcon:
So this would mean:

```
sudo mkdir -p /opt/matterbridge/bin
sudo cp matterbridge-$ARCH /opt/matterbridge/bin
gem install pleaserun
pleaserun --install  /opt/matterbridge/bin/matterbridge
```

Puts it in the users hands to install ruby, but then it works on any platform `pleaserun` supports :)

42wim:
@patcon feel free to edit stuff about this on the wiki, I'm going to close this PR now as @heinrich5991 #291 is merged.

PR #196, opened 2017-06-14
remove second flag.Parse()

syamgk:
flag.Parse() is already being called on line 28 https://github.com/42wim/matterbridge/blob/master/matterbridge.go#L28
and there is no need for calling it again


42wim:
thanks!

PR #204, opened 2017-06-23
Add webhook posting mode for discord.

saury07:
Using it implies to configure a Webhook on discord and set the parameter :
- WebhookURL (New parameter, discord-specific)

Discord API does not allow to change the name of the user posting, but webhooks does.
This makes the relay much more elegant, even if we might lose some more advanced features.

Discord API does not allow to change the name of the user posting, but webhooks does.
This makes the relay much more elegant, even if we might lose some more advanced features.

Resulting relays looks like this :
capture d ecran 2017-06-23 a 20 32 44
Instead of looking like this :
capture d ecran 2017-06-23 a 20 36 22

Signed-off-by: saury07 <sacha.aury@gmail.com>

### DIFF
--- a/bridge/discord/discord.go
+++ b/bridge/discord/discord.go
@@ -19,6 +19,8 @@ type bdiscord struct {
 	UseChannelID  bool
 	userMemberMap map[string]*discordgo.Member
 	guildID       string
+	webhookID     string
+	webhookToken  string
 	sync.RWMutex
 }
 
@@ -35,6 +37,12 @@ func New(cfg config.Protocol, account string, c chan config.Message) *bdiscord {
 	b.Remote = c
 	b.Account = account
 	b.userMemberMap = make(map[string]*discordgo.Member)
+	if b.Config.WebhookURL != "" {
+		flog.Debug("Configuring Discord Incoming Webhook")
+		webhookURLSplit := strings.Split(b.Config.WebhookURL, "/")
+		b.webhookToken = webhookURLSplit[len(webhookURLSplit)-1]
+		b.webhookID = webhookURLSplit[len(webhookURLSplit)-2]
+	}
 	return b
 }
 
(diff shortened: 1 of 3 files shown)


bridge/discord/discord.go
```diff
@@ -101,7 +101,21 @@ func (b *bdiscord) Send(msg config.Message) error {
 		flog.Errorf("Could not find channelID for %v", msg.Channel)
 		return nil
 	}
-	b.c.ChannelMessageSend(channelID, msg.Username+msg.Text)
+	if b.Config.UseAPI {
```
saury07: Not really sure of the default value here.

42wim:
Hi, thanks for the great PR. I didn't know discord had webhook support :-)

I'd like to merge this, but there are a couple of things I'm pondering about which are impacting your patch.  

1) It seems like it's not a fully fledged webhook on discord, only allows for incoming webhooks. (which of course is your usecase)

2) I'm personally not a big fan of webhooks (the fully fledged ones with outgoing, it's harder to setup for users. (not the case here))

I'm thinking about making some breaking changes in the configuration.
* changing the```URL``` key to ```WebhookURL``` (clearer for the enduser)
* changing ```BindAddress``` to ```WebhookBindAddress```
* make the "api" setup default and if you want to use webhooks, you'll have to set ```Webhooks``` to
   *  ```send``` (post using webhooks like discord (or maybe slack)),
   *  ```sendreceive```?  ```full```? ```complete```? (not sure about the value yet) for posting and receiving using webhooks
   *  empty the default which doesn't use webhooks at all
This way you can use the api for receiving and still use webhooks for sending only.

Coming back to your patch, I tried configuring webhooks on my discord server, and I got a webhook url, not a ```login``` and ```token```. This will throw users ;-) So I suggest making this one key ```WebhookURL``` which contains the full url and split it up ourselves in the needed id/token for the discordgo api. 

Any thoughts ? :)

saury07:
Hi,
Thanks for your review.
I updated the PR according to your review. This is much simpler now for the end user, you were right.
I am not a big fan myself of webhooks either, but I did not find any other way to achieve what I wanted (A really smooth bridge between Slack and Discord), so I went with this solution. Having to configure both the Bot and the Webhook is sad, but no choice.
I also considered that if the user specifies a `WebhookURL`, this is because he wants to post using the webhooks. This remove the need of the `useAPI` setting, which I personally find unclear.

42wim:
Yes, maybe going with your example, if WebhookURL is specified, then webhook will be used, is better.
It certainly is for the discord setup.

Your changes look good, thanks a lot for your work!

Issue #217, opened 2017-07-16
Slack threaded message parents resent on each child


jheiselman:
If you have a configuration problem, please first try to create a basic configuration following the instructions on [the wiki](https://github.com/42wim/matterbridge/wiki/How-to-create-your-config) before filing an issue.

Please answer the following questions. 

### Which version of matterbridge are you using?
0.16.1

### Please describe the expected behavior.
When a user starts a thread on Slack, child messages should cross the bridge while the parent message should only be sent the first time.

#### Use logs from running ```matterbridge -debug``` if possible.
On each child message received, the parent message also gets sent across the bridge.

### Any steps to reproduce the behavior?
1. On Slack, in a channel that is synced, start a thread by responding to a message using the "Start thread" option.
2. Type a message.
3. On the other side of the bridge (tested with Discord) you will see the original parent message, the child message, and then the parent message again.

#### (be sure to exclude or anonymize private data (tokens/passwords))
[discord.testserver]
Token="<redacted>"
Server="Test Server"
RemoteNickFormat="{NICK}: "

[slack.testserver]
Token="<redacted>"
RemoteNickFormat="{NICK}"

[[samechannelgateway]]
  name="bot-testing"
  enable=true
  accounts = ["discord.testserver","slack.testserver"]
  channels = ["bot-testing"]

PR #218, opened 2017-07-16
Suppress parent message when child message is received

jheiselman:
When a thread is started in Slack and a user makes a comment on the thread, matterbridge sends the original parent message again on each child comment. This change suppresses that.

This is addresses #217.

### DIFF
--- a/bridge/slack/slack.go
+++ b/bridge/slack/slack.go
@@ -14,11 +14,13 @@ import (
 )
 
 type MMMessage struct {
-	Text     string
-	Channel  string
-	Username string
-	UserID   string
-	Raw      *slack.MessageEvent
+	Text      string
+	Channel   string
+	Username  string
+	UserID    string
+	Thread_ts string
+	Ts	  string
+	Raw       *slack.MessageEvent
 }
 
 type Bslack struct {
@@ -219,6 +221,9 @@ func (b *Bslack) handleSlack() {
 	for message := range mchan {
 		// do not send messages from ourself
 		if b.Config.WebhookURL == "" && b.Config.WebhookBindAddress == "" && message.Username == b.si.User.Name {
+			continue
+		}
+		if message.Ts == message.Thread_ts {
 			continue
 		}
 		texts := strings.Split(message.Text, "\n")
--- a/bridge/slack/slack.go
+++ b/bridge/slack/slack.go
@@ -14,11 +14,11 @@ import (
 )
 
 type MMMessage struct {
-	Text      string
-	Channel   string
-	Username  string
-	UserID    string
-	Raw       *slack.MessageEvent
+	Text     string
+	Channel  string
+	Username string
+	UserID   string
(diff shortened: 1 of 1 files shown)


bridge/slack/slack.go
```diff
@@ -221,6 +223,9 @@ func (b *Bslack) handleSlack() {
 		if b.Config.WebhookURL == "" && b.Config.WebhookBindAddress == "" && message.Username == b.si.User.Name {
 			continue
 		}
+		if message.Ts == message.Thread_ts {
```
42wim: This will always be true, because they are both empty.

42wim:
Hi thanks for your contribution!
Did you test this ? The 2 new strings you added aren't getting any values.

42wim:
A newer version of the nlopes/slack library I'm using has support for threads, it looks like I have to update this library.

jheiselman:
I did test this. I'm running it right now with my changes, though, admittedly, I didn't build it before making the changes as I assumed that the binary I got from github was already built against master on this repo. I will revert my changes and build again and see if it is already resolved. Will update later today.

jheiselman:
Reverting my changes and building with master from this repo exhibits the broken behavior.

42wim:
Running it just doesn't show any messages anymore for me.
Could you explain me how it works ?

jheiselman:
Interesting. It was just a quick five minute patch and test, so I didn't put a whole lot of thought into it. According to the documentation on threaded conversations at https://api.slack.com/docs/message-threading, a parent message contains a thread_ts == to it ts property while child messages will contain their own unique ts and thread_ts is equal to the parent's.

I had assumed that you were populating MMMessage properties from json dynamically and so just adding the appropriate fields to the struct would allow them to be populated by whatever mechanism that was.

If that's not the case, let me know and I'll actually spend some time on it to get a full understanding.

42wim:
Thanks for explaining, no the MMMessage isn't created dynamically.

So it's strange that this works for you.

I've found the fix by
Changing the line on  https://github.com/42wim/matterbridge/blob/master/bridge/slack/slack.go#L244
from 
``` 
if !b.Config.EditDisable && ev.SubMessage != nil {
```
to
```
if !b.Config.EditDisable && ev.SubMessage != nil && ev.SubMessage.ThreadTimestamp != ev.SubMessage.Timestamp {
```

Could you try this change on your code ? Let me know if it works for you as expected and if it does, please update your pull request with that code :)

Thanks for your time!

jheiselman:
That did indeed work once I updated nlopes/slack in the vendor directory. Pull request updated with second commit

42wim:
Ok, great!
One small thing, Travis failed, it seems because gofmt failed.  (https://travis-ci.org/42wim/matterbridge/jobs/254184021)

Could you just fix that and then I can merge it.

jheiselman:
Alright done.

42wim:
Thanks!

PR #234, opened 2017-08-01
Add UseUserName option (discord)

anon724:
my friends on discord using stupid nick names all the time, this change makes it so i can show only username instead


42wim:
Thanks!

PR #240, opened 2017-08-16
Added new file_ids parameter for Mattermost outgoing webhook

Thracky:
As of version 4.1.0 it seems as if Mattermost is now including a file_ids parameter in the outgoing webhook requests. 

This adds the parameter to the IMessage structure so that schema doesn't complain and outgoing webhook messages function again.


42wim:
Thanks!

PR #279, opened 2017-10-18
[general] settings should be default, not total override (specifically RemoteNickFormat)

patcon:
If you have a configuration problem, please first try to create a basic configuration following the instructions on [the wiki](https://github.com/42wim/matterbridge/wiki/How-to-create-your-config) before filing an issue.

Please answer the following questions. 

v1.3.0

### Please describe the expected behavior.
Expect that setting `RemoteNickFormat` in `[general]` would be default, and that setting in one specific protocol would override that.

#### Use logs from running ```matterbridge -debug``` if possible.

Instead, seems that setting `RemoteNickFormat` in `[general]` supercedes everywhere else.

This would seem to be intended behaviour right now:
https://github.com/42wim/matterbridge/blob/master/matterbridge.toml.sample#L776

#### (be sure to exclude or anonymize private data (tokens/passwords))

```
[general]
RemoteNickFormat="{NICK}@{BRIDGE}"

# ...

[slack.hacklabto]
Token="${SLACK_HACKLABTO_TOKEN}" # bot user
RemoteNickFormat="{NICK}@{PROTOCOL}"

[telegram.hacklabto]
Token="${TELEGRAM_TOKEN}"
RemoteNickFormat="{NICK}@{PROTOCOL}: "
```

telegram screenshot from slack

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -243,9 +243,9 @@ func (gw *Gateway) ignoreMessage(msg *config.Message) bool {
 func (gw *Gateway) modifyUsername(msg config.Message, dest *bridge.Bridge) string {
 	br := gw.Bridges[msg.Account]
 	msg.Protocol = br.Protocol
-	nick := gw.Config.General.RemoteNickFormat
+  nick := dest.Config.RemoteNickFormat
 	if nick == "" {
-		nick = dest.Config.RemoteNickFormat
+    nick = gw.Config.General.RemoteNickFormat
 	}
 	if len(msg.Username) > 0 {
 		// fix utf-8 issue #193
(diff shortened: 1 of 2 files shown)


patcon:
Just went out on a limb and put this through. Happy for a conversation though, if it merits it.

patcon:
Noticed that you use tab style, so added an editorconfig and reverted the changes my vim setup had made

42wim:
Good job finding this bug :)

About the tab style, go has ```go fmt``` which automatically formats the go code (https://blog.golang.org/go-fmt-your-code)

I'd merge it if you could remove the ```.editorconfig``` from the commit and rebase. (if your git-fu isn't strong feel free to reopen a new PR without the .editorconfig)

patcon:
I will remove, but as someone who came here bc i use the tool (not golang), and made a change without ever installing golang, it would appear to be helpful to offer a flat file alternative that it pretty small? :)

Might you consider accepting it in another PR?

patcon:
also, one sec and I'll update the sample toml too

patcon:
Also, what say you to renaming `[general]` to `[default]`?

42wim:
Thanks for the patch.
I'll think about the rename

PR #287, opened 2017-10-25
Remove angle-brackets from urls coming out of slack

patcon:
v1.3.1

Would it be a sensible improvement to stop adding (or rather, to start stripping, if that's the case) the angle brackets from slack-slack bridge messages? It seems to prevent slack for hyperlinking and unfurling.

screenshot

I suspect it's happening here, but please let me know if it might exist elsewhere:
https://github.com/42wim/matterbridge/blob/master/bridge/slack/slack.go#L413


patcon:
Ok ready for review :)

42wim:
Thanks!
LGTM

PR #290, opened 2017-11-03
Fix build on go1.9.2

patcon:
```
~/repos/matterbridge (master ✘)✭ ᐅ go build
# _/home/patcon/repos/matterbridge
./matterbridge.go:30:31: cannot use agent.Options literal (type *agent.Options) as type agent.Options in argument to agent.Listen
~/repos/matterbridge (master ✘)✭ ᐅ go version
go version go1.9.2 linux/amd64
~/repos/matterbridge (master ✘)✭ ᐅ 
```

Not sure what's up, but I get this when i try to build. I've got a fix, but I'm not sure the implications of it. Will submit a PR, and perhaps you might have an idea what's going on. (Would love a pointer in the right direction to better understand what's going on here! :) )

### DIFF
--- a/matterbridge.go
+++ b/matterbridge.go
@@ -27,7 +27,7 @@ func main() {
 	flagGops := flag.Bool("gops", false, "enable gops agent")
 	flag.Parse()
 	if *flagGops {
-		agent.Listen(&agent.Options{})
+		agent.Listen(agent.Options{})
 		defer agent.Close()
 	}
 	if *flagVersion {


patcon:
fwiw, gops docs don't pass by reference either: https://github.com/google/gops#diagnostics

42wim:
Maybe you're not using the right gops vendored version. It seems like you updated the vendor dependencies?
It build just fine on 1.9.2 :) 
Your changes create an error though: see https://travis-ci.org/42wim/matterbridge/builds/296557533#L490

Also see the difference between my vendored version:
https://github.com/42wim/matterbridge/blob/master/vendor/github.com/google/gops/agent/agent.go#L61

and the current gops
https://github.com/google/gops/blob/master/agent/agent.go#L63

patcon:
Hm. yeah, totally. I must have done `go get ./...` at some point and fetched a new one into my actual GOPATH. Is there any way we could make building more foolproof for people like me who don't normally _do_ go? perhaps so that building a binary is just `make build`? (ie. fresh build from any given relative dir someone chooses to place the git repo in, as opposed to one that requires a special nested new spot on the filesystem)

42wim:
I have no idea how you did this, maybe you removed vendor and did a ```go get ./...``` otherwise this can't happen.

Go needs a GOPATH, since 1.8 (or 1.9 not sure), if not specified this is ```~/go```
So if you do a ```go get github.com/42wim/matterbridge``` it will put it in ```~/go/src/github.com/42wim/matterbridge```

The GOPATH thing is annoying for new developers, but it's made easier by using a default path in newer versions.

So to develop, go to ```~/go/src/github.com/42wim/matterbridge``` and do ```go build```

PR #288, opened 2017-11-03
Handle variables (@here, @channel, etc.) in Slack message

patcon:
Docs: https://api.slack.com/docs/message-formatting#variables

Basically almost of dup of `replaceMention`, but can accept a more channel-like syntax with or without the delimiter, and should prepend the variable with "@" like it's a mention.

I'll wait for the other PRs to be merged, as this will prob cause merge conflict. Lemme know if this sounds alright to you :) thanks!

### DIFF
--- a/bridge/slack/slack.go
+++ b/bridge/slack/slack.go
@@ -340,6 +340,7 @@ func (b *Bslack) handleSlackClient(mchan chan *MMMessage) {
 			}
 			m.Raw = ev
 			m.Text = b.replaceMention(m.Text)
+			m.Text = b.replaceVariable(m.Text)
 			m.Text = b.replaceChannel(m.Text)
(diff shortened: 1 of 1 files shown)


patcon:
Add urls to appropriate message formatting docs (`@see` seems to be a convention [accepted by aws](https://godoc.org/github.com/aws/aws-sdk-go/aws), so thought it would be good enough for us, though not officially a go-ism apparently)

patcon:
any insight into that travis failure?

```
bridge/discord/discord.go:158:6: the surrounding loop is unconditionally terminated (SA4004)
```

42wim:
Thanks, looks good.
The travis failure isn't related to your patch, was a commit that megacheck wasn't happy about. (fixed now)

The correct way to document functions can be found at https://golang.org/doc/effective_go.html#commentary but your way is fine for now :)

PR #294, opened 2017-11-05
Add support for bridging to individual steam chats.

rrigby:
Added this because I wanted to be able to bridge to individual people, let me know what you think!


42wim:
Looks good, it's small enough change to be merged.
You've tested it ? :)
How do you find the userid? https://steamidfinder.com/ and steamID64 ?

rrigby:
I tested it with a config that created two gateways, one bridging matrix and a steam group chat, another bridging matrix and a steam individual chat, both worked for me.

I found the user id by pasting their profile URL into steamrep.com (but your website works too) this is only required if they have a custom profile URL. most people just have the standard URL which is their 64bit userId.

42wim:
Thanks!

PR #350, opened 2018-01-29
Enable Long Polling for Telegram. Reduces bandwidth consumption.

### LINKED ISSUE #349: High bandwidth consumption with Telegram servers
Matterbridge exchanges anomalously high amount of data with Telegram servers. Usually about 500 Kbit/s - 1.5 Mbit/s, which consumes up to 20 GB of bandwidth per day.
I'm running only a single gateway with single Telegram account.

Matterbridge v1.7.0

ValdikSS:
Fixes #349.


42wim:
Thanks!

PR #352, opened 2018-01-31
Allow ShowJoinPart config to work for Slack bridges

patcon:
As stated here, this setting only works for IRC and Mattermost right now:

https://github.com/42wim/matterbridge/blob/master/matterbridge.toml.sample#L129

### DIFF
--- a/matterbridge.toml.sample
+++ b/matterbridge.toml.sample
@@ -126,7 +126,7 @@ ReplaceNicks=[ ["user--","user"] ]
 RemoteNickFormat="[{PROTOCOL}] <{NICK}> "
 
 #Enable to show users joins/parts from other bridges 
-#Only works hiding/show messages from irc and mattermost bridge for now
+#Currently works for messages from the following bridges: irc, mattermost, slack
 #OPTIONAL (default false)
 ShowJoinPart=false
 


patcon:
Hm. Seems to not work for Slack?

config: https://github.com/edgi-govdata-archiving/matterbridge-heroku/blob/edgi/config/config-heroku-template.toml#L20

patcon:
And the logs: https://gist.github.com/patcon/2594154df1446fd107ce0cfba5f3b319

patcon:
Actually, seems that it was another bit of misconfig..! sorry

But in the process of getting it working, i'm am fairly certain that trying to config in `[general]` or `[slack]` didn't work, and I had to set it in each `[slack.myorg]` section. Does that sound right?

42wim:
Yes, only the settings specified in `[general]` of matterbridge.toml.sample are general ;-)

42wim:
You should add your change to every `ShowJoinPart` description in the matterbridge.toml.sample
This is probably something this will eventually move to `[general]`, you may make an issue for that :)

```
$ grep ShowJoinPart matterbridge.toml.sample
ShowJoinPart=false
ShowJoinPart=false
ShowJoinPart=false
ShowJoinPart=false
ShowJoinPart=false
ShowJoinPart=false
ShowJoinPart=false
ShowJoinPart=false
ShowJoinPart=false
ShowJoinPart=false
ShowJoinPart=false
```

patcon:
Also, note to self for later spin-off into issues: slack config only shows joins, not "leaves". I should either clarify help text or fix (i'm on the run now)

Wooo! Go team!

42wim:
I've added your change to all bridges in the documentation because I was updating the file anway.

PR #390, opened 2018-03-15
Print list of valid team names when team not found

alecdwm:
When attempting to sign in to a team, this PR provides a more informative list of valid team options to choose from in the event that the user's selected team could not be found.


42wim:
Thanks, will take a few days before it'll show up in matterircd, I'm going to do some other vendor updates too

alecdwm:
No worries, thanks for the bridge!

Issue #391, opened 2018-03-22
Blank URL text from Slack should be ignored


jheiselman:
### Environment

Version: 1.9.0

Operating system: linux/amd64

### Please describe the expected behavior.
When a Google Hangout is bridged to Slack (via Hangoutsbot slackrtm plugin) it adds a URL with an empty description to the messages (i.e. <ho://xxxxx/xxxxx | >). This allows it to track the return path of a message should it need to, but also keeps the tag invisible to Slack users.

When Slack is then bridged to another platform using matterbridge, this tag is exposed on the far end of the bridge. It should be stripped out since it doesn't have any use beyond that bridge.

### Please describe the actual behavior. 

[1318] DEBUG slack:        == Receiving event &slack.MessageEvent{Msg:slack.Msg{Type:"message", Channel:"C699FK1U3", User:"", Text:"No hail damage yay <ho://<redacted>/<redacted>| >", Timestamp:"1521562009.000762", ThreadTimestamp:"", IsStarred:false, PinnedTo:[]string(nil), Attachments:[]slack.Attachment(nil), Edited:(*slack.Edited)(nil), SubType:"bot_message", Hidden:false, DeletedTimestamp:"", EventTimestamp:"1521562009.000762", BotID:"B695YDH96", Username:"Google User (Hangouts)", Icons:(*slack.Icon)(0xc42061a460), Inviter:"", Topic:"", Purpose:"", Name:"", OldName:"", Members:[]string(nil), ReplyCount:0, Replies:[]slack.Reply(nil), ParentUserId:"", File:(*slack.File)(nil), Upload:false, Comment:(*slack.Comment)(nil), ItemType:"", ReplyTo:0, Team:"T67TH8VRQ", Reactions:[]slack.ItemReaction(nil)}, SubMessage:(*slack.Msg)(nil)}
[1318] DEBUG slack:        <= Sending message from Google User (Hangouts) on slack.dudegroup to gateway
[1318] DEBUG slack:        <= Message is &config.Message{Text:"No hail damage yay ho://<redacted>/<redacted>", Channel:"dude_group", Username:"Google User (Hangouts)", UserID:"B695YDH96", Avatar:"", Account:"slack.dudegroup", Event:"", Protocol:"", Gateway:"", Timestamp:time.Time{wall:0x0, ext:0, loc:(*time.Location)(nil)}, ID:"slack 1521562009.000762", Extra:map[string][]interface {}{}}
[...]

PR #392, opened 2018-03-22
Strip markdown URLs with blank text

jheiselman:
This PR is to resolve issue #391. When Slack sends a message that contains a URL in markdown and the text portion of that URL is blank, the URL should be stripped from the message because it would never have shown up in Slack.


42wim:
Thanks for reporting and fixing!

PR #405, opened 2018-04-12
Send mediaserver link to Discord in Webhook mode

ValdikSS:
Right now no media files are sent to Discord in Webhook mode. This patch fixes this issue.


ValdikSS:
I'm not sure why "File" field in struct does not work. Is it possible to upload files directly to Discord in Webhook mode?

42wim:
Thanks!
Discord seems to support uploading files via webhook (https://discordapp.com/developers/docs/resources/webhook#execute-webhook) but the discordgo library doesn't, the file in the struct seems to be a string placeholder.

PR #421, opened 2018-05-08
Send attached files to XMPP in different message with OOB data and without body

ValdikSS:
Conversations can't show inline pictures if there's anything besides URL in the message body.
Workaround this issue by sending one usual message and one message with OOB data and without message body.
The second message should not be shown in the clients without OOB support, so the user won't see the empty message.

go-xmpp modifications are required: https://github.com/matterbridge/go-xmpp/pull/2


42wim:
Thanks!

PR #423, opened 2018-05-09
[WIP] Colorize username sent to IRC using its crc32 IEEE checksum

yuvallanger:


42wim:
Hi, looking good!
Would you mind also adding a config option (and documentation in matterbridge.toml.sample) ?
Like for example `ColorNicks` in https://github.com/42wim/matterbridge/blob/master/bridge/config/config.go#L60 which is by default false.
You can then do a `if b.GetBool("colornicks") { ... }`

yuvallanger:
This is the message I was looking for! Thank you for the guidance!

42wim:
Don't forget to let me know if the WIP is done :-)
Feel free to ask if you need more help/info!

yuvallanger:
It might be close to be done now. Could you test it?

derram:
I compiled the binary and tested it.

Nicks on irc weren't colored and it seemed to reverse the telegram quote option, causing the bridge to drop the message but display the quote.

I don't know much about go, though, so there's always the possibility that I didn't compile it correctly. There was a lot more text in the console on start up than there is on the main branch, for instance.

bridge/irc/irc.go
```diff
@@ -240,11 +241,16 @@ func (b *Birc) doSend() {
 	throttle := time.NewTicker(rate)
 	for msg := range b.Local {
 		<-throttle.C
+		username := msg.Username
+		if b.GetBool("Colornicks") {
+			checksum := crc32.ChecksumIEEE([]byte(msg.Username))
+			username = string(0x03) + string(checksum%0x10) + msg.Username + string(0x03)
```
42wim: Doesn't work, but changing to
`username = fmt.Sprintf("\x03%d%s\x03", checksum%0x10, msg.Username)` works

42wim:
If you fix this small thing and also update the matterbridge.toml.sample with your new feature, I can merge this.

Thanks!

yuvallanger:
Where do I put the setting in `matterbridge.toml.sample`?

42wim:
As it is for IRC only please add on line 94

yuvallanger:
Please have a look now.

42wim:
Thanks! merged

yuvallanger:
The commit message starts with `[WIP]`. Sorry about that. :/

PR #424, opened 2018-05-11
Solve light-on-light and dark-on-dark nickname colors problem

yuvallanger:
If a nickname ends up coloured white or black, this patch will add a background to the nickname coloured opposite to the foreground.


42wim:
Did you test this yourself ?
The last statement is outside the if/else if, which means it'll always be the last statement.
See also the travis build failure.

yuvallanger:
I am unsure how to solve this light-on-light and dark-on-dark problem. Maybe I should just set a background colour for each foreground colour and end it with that?

yuvallanger:
An alternative would be to have a list of (foreground, background) pairs as a configuration value, then have (Golang pseudocode):

```
foregroundColor, backgroundColor := colorList[hashNumber%len(colorList)]
```

derram:
Aside from the occasional unreadable nick, there's also an issue with color being passed on to the message text if the first characters from the bridge are numbers.

selection_307

selection_306

yuvallanger:
@42wim Please consider this suggestion. ( It compiles, but it was not yet tested.

Also, the issue raised by @derram on https://github.com/42wim/matterbridge/pull/424#issuecomment-389017022 was solved by https://github.com/42wim/matterbridge/pull/425.

bridge/irc/irc.go
```diff
@@ -242,15 +242,33 @@ func (b *Birc) Send(msg config.Message) (string, error) {
...
+	return fmt.Sprintf("\x03%02d,%02d%s\x03", foreground, background, s)
+}
+
+func (b *Birc) colorizeUsername(username string) string {
+	ircNickColors := b.GetStringSlice2D("ColorNicksList")
+	checksum := int(crc32.ChecksumIEEE([]byte(username)))
+	selectedColorPair := ircNickColors[checksum%len(ircNickColors)]
+	foreground, err := strconv.ParseInt(selectedColorPair[0], 16, 64)
+	if err == nil {
+		panic(fmt.Sprintf("Bad IRC color code: %v", selectedColorPair[0]))
```
42wim: Panics are not allowed for this, you'll have to handle this gracefully

bridge/irc/irc.go
```diff
@@ -242,15 +242,33 @@ func (b *Birc) Send(msg config.Message) (string, error) {
...
+	}
+	background, err := strconv.ParseInt(selectedColorPair[1], 16, 64)
+	if err == nil {
+		panic(fmt.Sprintf("Bad IRC color code: %v", selectedColorPair[1]))
```
42wim: Same as my previous message

42wim:
Could you also please test it yourself first ?

bridge/irc/irc.go
yuvallanger: What the possible ways of handling this error gracefully?

42wim: Maybe do a `b.Log.Errorf("Bad IRC color code: %v.", selectedColorPair[0])` and use a fixed foreground/background color pair to pass

rreuvekamp:
@yuvallanger this seems like a great idea! Are you continuing development on it (by testing and fixing the points raised by @42wim) or would you like someone else to take over?

yuvallanger:
re: https://github.com/42wim/matterbridge/pull/424#issuecomment-394474453

I would like to work on it.

re: https://github.com/42wim/matterbridge/pull/424#issuecomment-392365472

@42wim Is it not sufficient to let the CI test on push? I will test myself before pushing my commits.

rreuvekamp:
> I would like to work on it.

Yey cool

> I will test myself before pushing my commits.

Please also run go test and fix broken tests, since that's what the CI is complaining about it seems.

42wim:
To be fair the test that fails is my fault, it's testing a map which is random and most of the time works correct, I should fix that test.

yuvallanger:
https://github.com/42wim/matterbridge/pull/424#issuecomment-394474453
Sorry. I would like to stop working on this feature. Please someone take the reigns.

42wim:
@rreuvekamp feel free to take this now @yuvallanger has stopped working on this

rreuvekamp:
Opened

rreuvekamp:
@42wim What would be the preferred way to commit my changes? Would you like me to request write access to @yuvallanger 's repository and push the changes there, or would you like me to create a new pull request?

42wim:
@rreuvekamp please open a new PR, thanks!

PR #428, opened 2018-05-18
Fix format string bug

znz:
- Current behavior: When client says `%v`, bridged message is `%!v(MISSING)`
- After this fix: When client says `%v`, bridged message is `%v`


42wim:
Thanks!

PR #434, opened 2018-05-28
quick fix - prevent white or black IRC nickname foreground colors

yuvallanger:
White and black foregrounds create problems, so let's just prevent them from appearing.


derram:
Tested and working!

Still some issues with colors like dark gray and, I assume, a few of the white hues, but this solves the issue of completely unreadable nicks.

42wim:
Thanks for the patch and testing!

PR #437, opened 2018-05-29
Add message correction support for XMPP

ValdikSS:
It works worse than it could be, since message correction in XMPP works differently compared to other messengers. XMPP replaces old message with old ID with new message with new ID. Matterbridge
remembers only old ID, that's why you can only edit a message from XMPP to gateway once.

Edited messages from other networks to XMPP are handled correctly though.

### DIFF
--- a/bridge/xmpp/xmpp.go
+++ b/bridge/xmpp/xmpp.go
@@ -6,6 +6,7 @@ import (
 	"github.com/42wim/matterbridge/bridge/config"
 	"github.com/42wim/matterbridge/bridge/helper"
 	"github.com/jpillora/backoff"
+	"github.com/rs/xid"
 	"github.com/matterbridge/go-xmpp"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


42wim:
Thanks! 
Delete could also be implemented, by setting the message to "" ? (for another PR if this is possible :)

ValdikSS:
@42wim, not really. XEP-0308 describes only last message correction. According to the specification, other messages might be edited as well, but:

>While it is possible to use this protocol to correct messages older than the most recent received from a full JID, such use is out of scope for this document and support for this SHOULD NOT be assumed without further negotiation.

Nothing about deleting messages. But you got me an interesting idea, I'll try it.

ValdikSS:
@42wim, is it possible to add a function to replace message ID into matterbridge?
Currently, most clients just replace one message to other on edits, but matterbridge requires the same message id to be able to edit it.

PR #443, opened 2018-06-04
Add config option MediaUploadPath

rreuvekamp:
Add option MediaServerPath.
Can be used instead of MediaServerUpload for when your webserver is on the same system as matterbridge and matterbridge has write access to the serve dir.

Instead of uploading (like MediaServerUpload does), MediaServerPath will download it to a location on the current system.

I made this since running a caddy server (or configuring my current nginx installation for uploading) seems kind of overkill for me, when I could just give my matterbridge write access to the /srv/http directory.

### DIFF
--- a/bridge/config/config.go
+++ b/bridge/config/config.go
@@ -77,7 +77,7 @@ type Protocol struct {
 	MediaDownloadSize      int    // all protocols
 	MediaServerDownload    string
 	MediaServerUpload      string
-	MediaServerPath        string     // Basically MediaServerUpload, but instead of uploading it, just write it to a file on the same server.
+	MediaDownloadPath      string     // Basically MediaServerUpload, but instead of uploading it, just write it to a file on the same server.
 	MessageDelay           int        // IRC, time in millisecond to wait between messages
 	MessageFormat          string     // telegram
 	MessageLength          int        // IRC, max length of a message allowed
--- a/bridge/config/config.go
+++ b/bridge/config/config.go
@@ -2,14 +2,15 @@ package config
 
 import (
 	"bytes"
-	"github.com/fsnotify/fsnotify"
-	log "github.com/sirupsen/logrus"
-	"github.com/spf13/viper"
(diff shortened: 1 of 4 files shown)


rreuvekamp:
Please note that I didn't test if MediaServerUpload still works as intended, since it is kind of a pain to get a caddy server up and running. If someone with an already functioning MediaServerUpload could test if it still works as intended with the patch, that would be really great!

rreuvekamp:
Also limited the length of the hash in the URL to 8 characters, since that is unique enough. And even if two uploaded files have the same first 8 characters of the hash, it's still not a problem as long as the filenames are not the same. Including the full hash makes the URL quite long, and on terminal IRC clients, shorter URLs are preferred to make sure they fit on one line. Now that I think of it, this should have been a separate pull request, but I've already pushed the commit :$

42wim:
Thanks for the PR 👍 
It's looking good, I'm on the fence about the `MediaServerPath` name. Maybe `MediaDownloadPath` is a better name, because it's not doing anything on a remote server.

rreuvekamp:
Great!

> I'm on the fence about the MediaServerPath name. [...] because it's not doing anything on a remote server.

Yeah, kind of disagree with it too, but couldn't think of anything better. "MediaDownloadPath" makes me think it is an alternative to "MediaServerDownload" though

42wim:
Sorry, but I would like to keep `MediaDownloadPath` because it's the path where the media gets downloaded. It has nothing to do with the mediaserver, it could also be used for people who wants matterbridge to download attachments for any reason.

I hope you understand my reasoning.

rreuvekamp:
Okay, it's cool of course! Shall I push a commit to rename it?

rreuvekamp:
@42wim fixed!

42wim:
Merged, thanks a lot!

rreuvekamp:
Great, thank you for accepting!

PR #445, opened 2018-06-09
Reconnect on quit 431

lrstanley:
This should help in situations where an IRC proxy (e.g. znc) may send a `QUIT` event, it will ensure matterbridge reconnects, even if the disconnect was not in error.

Also ran the code through gofmt/goreturns to fix some of the inconsistencies.


lrstanley:
Had to force push to fix https://github.com/42wim/matterbridge/pull/445/commits/3b72594a9dafa85a6292403813ee557fccf8fb67#diff-228410cdab7b67ae6ed9a2aa28c657d0 not using proper import names.

Just FYI, when importing, if the package name you use doesn't match the folder name used in the import (e.g. https://github.com/42wim/matterbridge/pull/445/commits/3b72594a9dafa85a6292403813ee557fccf8fb67#diff-8cf1b62ff356263dd5a7c773e800dcc7, actual package being `samechannelgateway` but directory name being `samechannel`), this causes most editors to not properly provide intellisense/code-completion, and isn't standard. It should be concise and clear to the person reading the code, where the package came from.

lrstanley:
Also, I would use something like https://github.com/sqs/goreturns (or at least goimports) in your editor. VS Code has amazing support for these types of things, and it's my favorite editor for Go (even if it is a Microsoft product..).

FWIW, vim-go should also support goreturns/goimports/etc, among other editors, if that's what you use.

42wim:
Thanks for your PR!
I actually use goimports and it doesn't do the stuff you add.
Also it seems like vim-go can do the code-completion without the package matching the foldername.

image

Don't use goreturns though

PR #446, opened 2018-06-13
Add Title from attachment slack message

mvaleev:
Current behavior: Some monitoring systems use Title (like prometheus alertmanager) and this text lost
After this fix: Title text will be added to the message


42wim:
Thanks!

PR #450, opened 2018-06-18
Readme - link fix and minor clarification added. 

ckartchner:
Fixed typo in the GOPATH link.
While there, I added a sentence of clarification to make it clear that first code block in that section is about installing matterbridge and not related to setting up Go. Having no prior experience with Go, I was a bit confused when I go to the Building section. I believe the new line adds minor but useful clarification.


42wim:
Thanks! 
GOPATH is often an issue for go newbies. I've been there :)

PR #451, opened 2018-06-18
Add channel password support for XMPP

0xbb:
It would be great if we could add support for XMPP channel passwords.


0xbb:
I am not sure about the failed test. On my local machine it doesn't seem to fail.

42wim:
Don't worry about that test, it sometimes fails (my mistake). Has to do with the fact that map/hash entries are random.

Thanks a lot for your PR, looks good!

PR #465, opened 2018-07-10
Stop numbers being stripped after non-color control codes

qtaf:
Currently numbers are stripped not just after the color control code (\x03) but also after other formatting such as bold (\x02) and italic (\x1D), which is both unnecessary and leads to missing text from irc. This fixes that by only stripping numbers after the color control code.


42wim:
Thanks!

Issue #460, opened 2018-06-28
Discord mentions


NikkyAI:
**Is your feature request related to a problem? Please describe.**

People want to mention / ping people on discord

**Describe the solution you'd like**

incoming text to discord should be scanned for mentions eg: `\b(@[^\s]+)\b`
and find matching nickanmes or username,
which one is prefered might be determined by the `UseUserName` option
so that it matches the username people see on ther bridges first

**Describe alternatives you've considered**

**Additional context**

this feature should be disabled by default to keep spam by default to a minimum,
i'd suggest `EnableUserMentions` as config key

and `EnableRoleMentions`, obviously restrictedby mentionable roles

`@here` and `@everyone` should probably be on a seperate config key as well

Some more about name matching:
Since spaces are allowed in names it should probably try with more words if the current word did match multiple users
gven the users `@user name abc` and `@user name abcdef` 
wit hthe query `@user name a` it would try `@user`, `@user name` and finally `user name a`
which should match the first user
hopefully nobody ever ends up with such a situaltion, but usernames with spaces are sadly too common

TheHolyRoger:
https://github.com/reactiflux/discord-irc/blob/master/lib/bot.js#L248

This is implemented in this bot, linking for reference in case useful!

NikkyAI:
a very naive implementation, excuse my lacking go skills, i touched the language for the first time today

```go
func (b *Bdiscord) replaceUserMentions(text string) string {
	members, err := b.c.GuildMembers(b.guildID, "", 1000)
	if(err != nil) {
		b.Log.Error("Error obtaining guild members", err)
		return text
	}
	b.Log.Debug("found users: %s", members)

	re := regexp.MustCompile("@[^@]+")
	text = re.ReplaceAllStringFunc(text, func(m string) string {
		mention := m[1:]
		b.Log.Debugf("Testing mention: '%s'", mention)
		//iterate over members
		for i, _ := range mention {
			b.Log.Debug("Testing Mention substring", mention[:i])
			for _, member := range members {
				nick := b.getNick(member.User)
				if nick == mention[:i] {
					return member.User.Mention()
				}
			}
		}
		return m
	})
	return text
}
```

surely the guild members should be cached
i am not sure if my solution to iterate the string slices is smart.. it seems to work

PR #492, opened 2018-08-31
Replace @... string with user mention if match found

jheiselman:
Added check for @-mention pattern and replacing it with a user with a matching Nick on incoming messages

This is in reference to issue #460

### DIFF
--- a/bridge/discord/discord.go
+++ b/bridge/discord/discord.go
@@ -22,6 +22,7 @@ type Bdiscord struct {
 	Nick           string
 	UseChannelID   bool
 	userMemberMap  map[string]*discordgo.Member
+	nickMemberMap  map[string]*discordgo.Member
 	guildID        string
 	webhookID      string
 	webhookToken   string
@@ -319,6 +320,7 @@ func (b *Bdiscord) getNick(user *discordgo.User) string {
 		return user.Username
 	}
 	b.userMemberMap[user.ID] = member
+	b.nickMemberMap[member.Nick] = member
 	// only return if nick is set
 	if b.userMemberMap[user.ID].Nick != "" {
 		return b.userMemberMap[user.ID].Nick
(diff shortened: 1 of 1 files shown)


NikkyAI:
it is notable that  think all things done by `replaceChannelMentions` is handled by the api as well in `m.ContentWithMoreMentionsReplaced(b.c)`

its just something i remmembered while looking at it

unless i misunderstand and the api builtin handling does not work in all situations and has permission restrictions ?

NikkyAI:
what about nicknames containing spaces ?
i would try to match against `@[^@]+` and then try to match `m[1:2]` till `m[1:]` to match against pings

bridge/discord/discord.go
```diff
@@ -364,6 +398,21 @@ func (b *Bdiscord) replaceChannelMentions(text string) string {
 	return text
 }
 
+func (b *Bdiscord) replaceUserMentions(text string) string {
+	re := regexp.MustCompile("@[^ ]+")
```
NikkyAI: see somment about usernames or nicknames containing spaces, this is possible in discord iirc

jheiselman: Alright, turns out that Discord usernames and nicks are limited to 32 characters. That means that the regex `(@[^@]{1,32})\b` would work to limit it to a max of 32 characters while automatically providing a word boundary to reduce the space a little further. From here, we can test each variation by popping off entire words at a time until we have a match or we've exhausted the testing space.

jheiselman: That will give us a maximum of 16 tests in the worst case scenario.

bridge/discord/discord.go
```diff
@@ -323,6 +326,37 @@ func (b *Bdiscord) getNick(user *discordgo.User) string {
 	return user.Username
 }
 
+func (b *Bdiscord) getGuildMemberByNick(nick string) (*discordgo.Member, error) {
```
42wim: This function will have locking issues, you can't range over the userMemberMap without locking.
I think it's better to just make a new map called nickMemberMap which has the nick as key. 

So this can be rewritten analogue to the existing `getNick` function.

Also update the `memberUpdate` function to add the new nick to `nickMemberMap`

bridge/discord/discord.go
```diff
@@ -364,6 +382,34 @@ func (b *Bdiscord) replaceChannelMentions(text string) string {
 	return text
 }
 
+func (b *Bdiscord) replaceUserMentions(text string) string {
+	re := regexp.MustCompile("@[^@]{1,32}")
+	text = re.ReplaceAllStringFunc(text, func(m string) string {
+		mention := strings.TrimSpace(m[1:])
+		var member *discordgo.Member
+		var err error
+		for {
```
42wim: I don't think we need the for loop here? Also the `var member` and `var err` can be removed.
something like below should work (not tested)
```
        b.Log.Debugf("Testing mention: '%s'", mention)
	member, err := b.getGuildMemberByNick(mention)
	if err != nil {
		lastSpace := strings.LastIndex(mention, " ")
		if lastSpace != -1 {
			mention = strings.TrimSpace(mention[0:lastSpace])
		}
                return m
        }
	return member.User.Mention()
```

42wim:
Sorry for the delay on replying.
Looks good, some small nitpicks and it can be merged.
(also try to get the README.md changes out of your commits)

PR #498, opened 2018-09-14
fix Discord mentions by populating the nickMemberMap at connect

NikkyAI:
in my local testing the memberUpdates never arrived realibly after connecting the bot, i waited ~5 minutes
i think it is reasonable to initialize the mapping and then update it via the events


NikkyAI:
i updated and promptly discovered more things, discord mentions do not work on webhooks (i only tested wit ha very basic config in dev)

and the nickname updates seem to affect all connected guilds and not just the one that is affected

PR #501, opened 2018-09-20
Add Gateway's name to RemoteNickFormat

Helcaraxan:
In order to support extra use cases we should add the `{GATEWAY}` tag to the `RemoteNickFormat` string which would be replaced by the value of the `name=` field from a gateway's configuration.

This is _very_ useful when you are forwarding, for example, multiple channels from one chat to a single channel on another one (one-way). It will help you identify the source channel of a message on the target chat.


42wim:
Thanks for your contribution!
Looks fine to me, but I think the matterclient/matterclient.go change shouldn't be there?

Helcaraxan:
Hey. Sorry for the late reply. Was off on holiday last week. The change in `matterclient/matterclient/go` is just a drive-by fix that my linters were complaining about. The formatting directive is not correct. On second look my fix is not even enough as there are two formatting directives but only one input argument.

Should the whole thing be:
```go
	if m.GetUser(data.UserId) == nil {
		m.log.Infof("User %s is not known, ignoring message %s", data.UserId, data.Message)
		return
	}
```
? In that case I would indeed not need to change the formatting directive.

Helcaraxan:
Will also add the necessary documentation changes for the new `{GATEWAY}` tag in the `RemoteNickId` format.

PR #508, opened 2018-09-28
various cleanups

dajohi:
- please go vet
- adding missing mutex Unlock

### DIFF
--- a/bridge/api/api.go
+++ b/bridge/api/api.go
@@ -110,8 +110,8 @@ func (b *Api) handleStream(c echo.Context) error {
 	c.Response().Header().Set(echo.HeaderContentType, echo.MIMEApplicationJSON)
 	c.Response().WriteHeader(http.StatusOK)
 	greet := config.Message{
-		Event:config.EVENT_API_CONNECTED,
-		Timestamp:time.Now(),
+		Event:     config.EVENT_API_CONNECTED,
+		Timestamp: time.Now(),
 	}
 	if err := json.NewEncoder(c.Response()).Encode(greet); err != nil {
 		return err
--- a/bridge/discord/discord.go
+++ b/bridge/discord/discord.go
@@ -91,6 +91,7 @@ func (b *Bdiscord) Connect() error {
 	}
 	// obtaining guild members and initializing nickname mapping
 	b.Lock()
+	defer b.Unlock()
 	members, err := b.c.GuildMembers(b.guildID, "", 1000)
 	if err != nil {
 		b.Log.Error("Error obtaining guild members", err)
@@ -103,7 +104,6 @@ func (b *Bdiscord) Connect() error {
 			b.nickMemberMap[member.Nick] = member
 		}
 	}
-	b.Unlock()
 	return nil
 }
 
@@ -488,9 +488,16 @@ func (b *Bdiscord) handleUploadFile(msg *config.Message, channelID string) (stri
 	var err error
 	for _, f := range msg.Extra["file"] {
 		fi := f.(config.FileInfo)
-		files := []*discordgo.File{}
-		files = append(files, &discordgo.File{fi.Name, "", bytes.NewReader(*fi.Data)})
-		_, err = b.c.ChannelMessageSendComplex(channelID, &discordgo.MessageSend{Content: msg.Username + fi.Comment, Files: files})
+		file := discordgo.File{
+			Name:        fi.Name,
+			ContentType: "",
+			Reader:      bytes.NewReader(*fi.Data),
+		}
+		m := discordgo.MessageSend{
+			Content: msg.Username + fi.Comment,
+			Files:   []*discordgo.File{&file},
+		}
+		_, err = b.c.ChannelMessageSendComplex(channelID, &m)
 		if err != nil {
 			return "", fmt.Errorf("file upload failed: %#v", err)
 		}
--- a/bridge/slack/helpers.go
+++ b/bridge/slack/helpers.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/nlopes/slack"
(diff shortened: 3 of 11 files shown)


bridge/api/api.go
```diff
@@ -93,9 +93,10 @@ func (b *Api) handlePostMessage(c echo.Context) error {
 
 func (b *Api) handleMessages(c echo.Context) error {
 	b.Lock()
-	defer b.Unlock()
```
42wim: why? what's wrong with the defer?

bridge/discord/discord.go
```diff
@@ -93,6 +93,7 @@ func (b *Bdiscord) Connect() error {
 	b.Lock()
 	members, err := b.c.GuildMembers(b.guildID, "", 1000)
 	if err != nil {
+		b.Unlock()
```
42wim: good catch, but should be a defer b.Unlock() after the Lock()

bridge/steam/steam.go
```diff
@@ -98,11 +98,13 @@ func (b *Bsteam) Send(msg config.Message) (string, error) {
 
 func (b *Bsteam) getNick(id steamid.SteamId) string {
 	b.RLock()
-	defer b.RUnlock()
-	if name, ok := b.userMap[id]; ok {
-		return name
+	name, ok := b.userMap[id]
```
42wim: why those changes ? what's wrong with the current code ?

42wim:
Thanks for this PR! 
I do have some questions about some changes I don't really understand though :)

bridge/api/api.go
dajohi: defer is slower.  figure it is not needed since there are no early returns.

bridge/slack/slack.go
```diff
@@ -47,8 +45,6 @@ func (b *Bslack) Command(cmd string) string {
 }
 
 func (b *Bslack) Connect() error {
-	b.RLock()
```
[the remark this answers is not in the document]
dajohi: it doesn't protect anything.

bridge/steam/steam.go
dajohi: this is just style to ditch the defer.

bridge/api/api.go
42wim: Okay, but I don't really think speed matters here.
I prefer using locking and defer unlocking where possible (unless it gets locked too long) so that unlocks don't get forgotten and it's more readable.

dajohi:
cleaned up.

bridge/slack/helpers.go
```diff
@@ -61,22 +60,16 @@ func (b *Bslack) getChannelByID(ID string) (*slack.Channel, error) {
...
-	earliestUserRefresh    = time.Now()
-)
-
 func (b *Bslack) populateUsers() {
-	refreshMutex.Lock()
-	if time.Now().Before(earliestUserRefresh) || refreshInProgress {
-		b.Log.Debugf("Not refreshing user list as it was done less than %d seconds ago.", int(minimumRefreshInterval.Seconds()))
-		refreshMutex.Unlock()
+	b.refreshMutex.Lock()
+	defer b.refreshMutex.Unlock()
```
Helcaraxan: We can't use a `defer` here. The purpose of manually unlocking it early, after the refresh timer and in-progress check is to enable other calls to this function not to block on it.

bridge/slack/helpers.go
```diff
@@ -95,19 +88,20 @@ func (b *Bslack) populateUsers() {
...
+	b.refreshInProgress = false
 }
 
 func (b *Bslack) populateChannels() {
-	refreshMutex.Lock()
-	if time.Now().Before(earliestChannelRefresh) || refreshInProgress {
-		b.Log.Debugf("Not refreshing channel list as it was done less than %d seconds ago.", int(minimumRefreshInterval.Seconds()))
-		refreshMutex.Unlock()
+	b.refreshMutex.Lock()
+	defer b.refreshMutex.Unlock()
```
Helcaraxan: Same comments as above about the need for manual unlocking.

dajohi:
@Helcaraxan How is the latest code?

bridge/slack/helpers.go
```diff
@@ -61,22 +60,17 @@ func (b *Bslack) getChannelByID(ID string) (*slack.Channel, error) {
...
-
 func (b *Bslack) populateUsers() {
-	refreshMutex.Lock()
-	if time.Now().Before(earliestUserRefresh) || refreshInProgress {
-		b.Log.Debugf("Not refreshing user list as it was done less than %d seconds ago.", int(minimumRefreshInterval.Seconds()))
-		refreshMutex.Unlock()
+	b.refreshMutex.Lock()
+	if time.Now().Before(b.earliestUserRefresh) || b.refreshInProgress {
+		b.Log.Debugf("Not refreshing user list as it was done less than %v ago.",
+			minimumRefreshInterval)
```
Helcaraxan: Nit: Please don't split this across multiple lines. Go does not have a line length. 🙂 

And if there were need to split it please use the: one-arg-per-line style like a struct declaration:
```go
{
  b.Log.Debugf(
    "Not refreshing user list as it was done less than %v ago.",
    minimumRefreshInterval,
  )
}
```

matterclient/matterclient.go
```diff
@@ -365,7 +365,8 @@ func (m *MMClient) parseActionPost(rmsg *Message) {
 	data := model.PostFromJson(strings.NewReader(rmsg.Raw.Data["post"].(string)))
 	// we don't have the user, refresh the userlist
 	if m.GetUser(data.UserId) == nil {
-		m.log.Infof("User %s is not known, ignoring message %s", data.UserId, data.Message)
+		m.log.Infof("User '%v' is not known, ignoring message '%#v'",
+			data.UserId, data)
```
Helcaraxan: Actually same comment about line-splitting here.

bridge/slack/helpers.go
Helcaraxan: @42wim I know these things tend to be like viper pits but as owner of the project what is your position on code-style? As much as Go is not usually the most troublesome of languages on that point do you want to (slowly) start enforcing a particular style?

Currently I am mostly going by my own references and the things that I have seen around in other projects. Would not want in any way to cross a line with my suggestions.

dajohi: not a fan of one-arg-per-line myself, but I am a fan of 80-char wide.  I leave it to @42wim to choose between current code or reverting to all-on-one-line

42wim:
Thanks @dajohi and @Helcaraxan !

Issue #510, opened 2018-10-07
Move to use Slack's new APIs and tokens?


Helcaraxan:
# Context & problem
Somewhat. Almost filed this as a "bug" rather than an FR but thought that as it is not a blocker issue I'd go this way to open up the discussion. I have been in the process of setting up some bridging between, mostly, Discord & Slack. Matterbridge is great for this (love, kudos and many thanks to this great repo and its contributors!).

The Discord-side has been entirely painless, up and running within minutes (if not seconds). However the Slack-side has been (and still is) a painful experience. Everything worked fine on my own test-Slack workspace but when migrating to an Enterprise Slack things got a little bit spicier. The obvious failure in the room was the continuous inability for Matterbridge to forward incoming Discord messages with the error `channel_not_found`.

# Solution
Debugging brought me into the depths of the Slack API and the `nlopes/slack` (which are both old friends o' mine). Long story short, Matterbridge relies (if one follows the documentation) on Slack legacy tokens. There have however been significant changes with regards to how Slack deals with Apps and their bots and it is nowadays perfectly possible to use non-legacy tokens. Another, separate, change has been their ["Conversations API"](https://api.slack.com/docs/conversations-api) which aims at simplifying the management of all things chat.

The easiest, albeit not necessarily best, solution would be to switch entirely to using the new Apps+bot approach and rewrite the code in `bridge/slack/slack.go` accordingly. This would remove the possibility to use legacy tokens. A clean-up of the semantics of the Slack-section in the configuration file could be done at the same time.

# Alternatives
If the removal of legacy tokens is considered to be too much of burden and there is a preference for a more gradual migration there is an alternative:

Maintaining the status quo that is adapted to legacy token and write separate new code-paths for the Apps+bots path. These paths could be auto-chosen based on the specified token value (as the new bot-tokens all start with `xoxb-`, unlike the legacy tokens).

# Further notes
I'd be happy to write the code for this. I'll be writing a variant of it anyway given that I will need to use the new Apps+bot approach myself. However thought I'd ask for input here first to decide which of both ways would be preferred when the time comes for a PR.

42wim:
Hi, thanks for looking into this!

I've noticed the conversation API changes (matterircd has added support), but haven't got time yet to add this to matterbridge (and I also don't have access to the paid version) so your help and a PR is more than welcome!

I've looked into the bot tokens before and one issue was that matterbridge couldn't join the channels by itself, it needs to be invited (I'm not sure if this is still the case).

[...]

Issue #481, opened 2018-08-08
Matterbridge incorrectly posts image caption when bridging Telegram <=> Slack


genotrance:
**Describe the bug**
I'm using matterbridge to bridge a Telegram channel to Slack. Everything works great except when someone other than me posts an image with a caption. As soon as that happens, my bot will repost the caption from that image as a plain text message to Telegram.

**To Reproduce**
Bridge Telegram to Slack, wait for someone to post an image with caption. Your bot will post the caption back to the channel.

**Screenshots/debug logs**
I've attached a sceenshot from Telegram and Slack for reference.

**Environment (please complete the following information):**
 - OS: Ubuntu
 - Matterbridge version: Latest #head as of today

genotrance:
![slack](https://user-images.githubusercontent.com/192231/43858395-88956552-9b13-11e8-8eee-74775c8bba28.JPG)

Note how the caption doesn't show up as from Huy but from my account.

genotrance:
![screenshot_20180808-140247](https://user-images.githubusercontent.com/192231/43858675-42cfda6a-9b14-11e8-9f29-7b64064caefe.png)

Attached is the Telegram side. Note how the caption is echoed by the bot.

42wim:
Is it possible you're using both a slack token and slack webhook ?
Could you paste your config ? (don't forget to obfuscate the tokens)

genotrance:
Just a token, no webhook.

```
[slack]
  [slack.name]
  Token="xyz"
  RemoteNickFormat="{NICK}"

[telegram]
  [telegram.name]
  Token="xyz"
  UseFirstName=true

[[gateway]]
name="slack-telegram"
enable=true
  [[gateway.inout]]
  account="slack.name"
  channel="cname"

  [[gateway.inout]]
  account="telegram.name"
  channel="-123123123"
```

42wim:
@genotrance could you update to latest master and let me know if it's still an issue ? the slack library has been updated, and some stuff regarding to image handling has been fixed.

If the issue persists, could you possibly run matterbridge with `-debug` and reproduce the issue.
And paste the debug logs here or on pastebin.

genotrance:
I used to run with `go get` and now just download the latest binary release posted. What's the easiest way to build master? `git clone` and then /?

Thanks in advance.

42wim:
A couple of ways:
* You can run `go get -u github.com/matterbridge/42wim` to build the latest binary yourself.
[...]

PR #519, opened 2018-10-12
Code clean-up and strengthening

Helcaraxan:
Further prefactor work for #510.

Changes include:
- Refactor of strings into package-wide constants.
- Predeclaration of regexps to be instantiated at package load time.
- Checking of unchecked errors.
- Structural changes:
  - Adding verifications to type-casting code.
  - Remove unnecessary 'len(X) > 0' checks before iterating over X.
  - Remove unnecessary 'else' clause after 'if' with 'return'.
  - Unexporting of public fields of Bridge struct.
- Formatting:
  - One-field-per-line struct definitions.

### DIFF
--- a/bridge/slack/handlers.go
+++ b/bridge/slack/handlers.go
@@ -14,7 +14,7 @@ import (
 
 func (b *Bslack) handleSlack() {
 	messages := make(chan *config.Message)
-	if b.GetString("WebhookBindAddress") != "" {
+	if b.GetString(incomingWebhookConfig) != "" {
 		b.Log.Debugf("Choosing webhooks based receiving")
 		go b.handleMatterHook(messages)
 	} else {
@@ -43,7 +43,7 @@ func (b *Bslack) handleSlack() {
 
 func (b *Bslack) handleSlackClient(messages chan *config.Message) {
 	for msg := range b.rtm.IncomingEvents {
-		if msg.Type != "user_typing" && msg.Type != "latency_report" {
+		if msg.Type != sUserTyping && msg.Type != sLatencyReport {
 			b.Log.Debugf("== Receiving event %#v", msg.Data)
 		}
 		switch ev := msg.Data.(type) {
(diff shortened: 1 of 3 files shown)


gateway/gateway_test.go
```diff
@@ -265,11 +338,41 @@ func TestGetDestChannelAdvanced(t *testing.T) {
 					hits[gw.Name]++
 					switch br.Account {
 					case "irc.zzz":
-						assert.Equal(t, []config.ChannelInfo{{Name: "#main", Account: "irc.zzz", Direction: "out", ID: "#mainirc.zzz", SameChannel: map[string]bool{"announcements": false}, Options: config.ChannelOptions{Key: ""}}, {Name: "#main-help", Account: "irc.zzz", Direction: "out", ID: "#main-helpirc.zzz", SameChannel: map[string]bool{"announcements": false}, Options: config.ChannelOptions{Key: ""}}}, channels)
```
Helcaraxan: This change was required as test could fail because of wrong ordering of slice elements.

bridge/slack/helpers.go
```diff
@@ -58,58 +61,52 @@ func (b *Bslack) getChannelByID(ID string) (*slack.Channel, error) {
...
-		}
-	}
-	return ""
-}
+var (
+	mentionRE  = regexp.MustCompile(`<@([a-zA-Z0-9]+)>`)
+	channelRE  = regexp.MustCompile(`<#[a-zA-Z0-9]+\|(.+?)>`)
+	variableRE = regexp.MustCompile(`<!((?:subteam\^)?[a-zA-Z0-9]+)(?:\|@?(.+?))?>`)
+	urlRE      = regexp.MustCompile(`<(.*?)(\|.*?)?>`)
+)
```
Helcaraxan: The best-practice is to instantiate regexps at package-load time as package-wide variables. This way any issue in a regular expression will provoke an immediate crash and can't creep in silently because of lack of testing.

bridge/slack/slack.go
```diff
@@ -18,22 +19,52 @@ type Bslack struct {
 	mh           *matterhook.Client
 	sc           *slack.Client
 	rtm          *slack.RTM
```
Helcaraxan: The `Usergroups` fields is used nowhere. There are also not public uses of the `Users` and `UseChannelID` as this struct implements a generic `bridge.Bridge` interface and has no use for publicly-exported fields.

bridge/slack/handlers.go
```diff
@@ -308,29 +329,24 @@ func (b *Bslack) skipMessageEvent(ev *slack.MessageEvent) bool {
...
-				} else {
-					b.Log.Debugf("Not skipping %s %s", f.Name, time.Now().String())
-				}
-			}
+	for _, f := range ev.Files {
+		// if the file is in the cache and isn't older then a minute, skip it
+		if ts, ok := b.cache.Get("file" + f.ID); ok && time.Since(ts.(time.Time)) < time.Minute {
+			b.Log.Debugf("Not downloading file id %s which we uploaded", f.ID)
+			return true
+		} else if ts, ok := b.cache.Get("filename" + f.Name); ok && time.Since(ts.(time.Time)) < time.Minute {
```
Helcaraxan: I've moved this delay back to `time.Minute` instead of `10 * time.Second` as it seemed contradictory with the comment and the other delay.

42wim:
Thanks for going through this all! 
Looks good, only 2 small nitpicks
- attachments instead attachements
- I'd like the gateway_test change in a separate PR if possible

Helcaraxan:
Have split out the gateway_test changes in #520.

bridge/slack/handlers.go
Helcaraxan: Can revert this back indeed if you want.

Looks like it might be interesting to make this a little bit more solid by rather checking for something like an MD5 of the uploaded file? WDYT?

42wim: Yes, better to revert this for now. We can't check for md5, because that would mean actually downloading the file. (ofcourse we can but I rather avoid this if possible)
Code is related to https://github.com/42wim/matterbridge/issues/481

bridge/slack/handlers.go
```diff
@@ -61,17 +61,25 @@ func (b *Bslack) handleSlackClient(messages chan *config.Message) {
...
+			b.users, err = b.sc.GetUsers()
+			if err != nil {
+				b.Log.Errorf("Could not reload users: %#v", err)
+			}
 		case *slack.ConnectedEvent:
 			var err error
-			b.channels, _, err = b.sc.GetConversations(&slack.GetConversationsParameters{Limit: 1000, Types: []string{"public_channel,private_channel,mpim,im"}})
+			b.channels, _, err = b.sc.GetConversations(&slack.GetConversationsParameters{
+				Limit: 1000,
+				Types: []string{"public_channel,private_channel,mpim,im"},
```
patcon: @42wim re: `mpim,im`: Do DM's (group or one-to-one) get used anywhere?

PR #520, opened 2018-10-12
Gateway test race fix.

Helcaraxan:
The gateway unit-tests could fail as it relies on a non-guaranteed ordering
within a computed slice. We fix it by only testing for the presence of the
necessary elements and not their order.


42wim:
Thanks!

PR #521, opened 2018-10-12
Clean-up user and channel information management

Helcaraxan:
More prefactoring for #510 

- Channel and users are no longer exported as raw fields in the Bridge struct.
- Channel and users are now stored as maps for easy & constant-time access.
- Refactored the management and retrieval of channels and users into independent functions.

### DIFF
--- a/bridge/slack/helpers.go
+++ b/bridge/slack/helpers.go
@@ -38,27 +38,72 @@ func (b *Bslack) getChannel(channel string) (*slack.Channel, error) {
 }
 
 func (b *Bslack) getChannelByName(name string) (*slack.Channel, error) {
-	if b.channels == nil {
-		return nil, fmt.Errorf("%s: channel %s not found (no channels found)", b.Account, name)
-	}
-	for _, channel := range b.channels {
-		if channel.Name == name {
-			return &channel, nil
-		}
+	b.channelsMutex.RLock()
+	defer b.channelsMutex.RUnlock()
+
+	if channel, ok := b.channelsByName[name]; ok {
+		return channel, nil
 	}
 	return nil, fmt.Errorf("%s: channel %s not found", b.Account, name)
 }
 
 func (b *Bslack) getChannelByID(ID string) (*slack.Channel, error) {
-	if b.channels == nil {
-		return nil, fmt.Errorf("%s: channel %s not found (no channels found)", b.Account, ID)
-	}
-	for _, channel := range b.channels {
-		if channel.ID == ID {
-			return &channel, nil
-		}
+	b.channelsMutex.RLock()
+	defer b.channelsMutex.RUnlock()
+
+	if channel, ok := b.channelsByID[ID]; ok {
+		return channel, nil
 	}
 	return nil, fmt.Errorf("%s: channel %s not found", b.Account, ID)
+}
+
+func (b *Bslack) populateUsers() {
+	users, err := b.sc.GetUsers()
+	if err != nil {
+		b.Log.Errorf("Could not reload users: %#v", err)
+		return
+	}
+
+	newUsers := map[string]*slack.User{}
+	for _, user := range users {
+		newUsers[user.ID] = &user
+	}
+
+	b.usersMutex.Lock()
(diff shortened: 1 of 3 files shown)


bridge/slack/helpers.go
```diff
@@ -38,27 +38,54 @@ func (b *Bslack) getChannel(channel string) (*slack.Channel, error) {
...
 	}
-	for _, channel := range b.channels {
-		if channel.ID == ID {
-			return &channel, nil
+}
+
+func (b *Bslack) populateChannels() {
+	b.channelsByID = map[string]*slack.Channel{}
+	b.channelsByName = map[string]*slack.Channel{}
+
```
42wim: Maybe put a `Lock()` here as it's called from multiple routines? or use `sync.Map`?
Same in `getChannelByName` and  `getChannelByID` (they're getting called a lot by the `Send()` function)

Also maybe better to not empty them (`b.channelsByID` and  `b.channelsByName`) here, but create new temporary ones and swap them at the end after they're filled in to keep the locking to a minimum?

Helcaraxan: Good call indeed. Have kept the locking to a minimum and added RW locks to keep parallel read calls to the maps possible.

bridge/slack/helpers.go
```diff
@@ -38,27 +38,70 @@ func (b *Bslack) getChannel(channel string) (*slack.Channel, error) {
...
+		if err != nil {
+			b.Log.Errorf("Could not reload channels: %#v", err)
+			return
 		}
+		for _, channel := range channels {
+			newChannelsByID[channel.ID] = &channel
+			newChannelsByName[channel.Name] = &channel
+		}
+		if nextCursor == "" {
+			return
```
42wim: This should be `break`

42wim:
I'm trying your code now, and my bridge is failing to start on slack.
The `break` was one issue, debugging further now

bridge/slack/helpers.go
```diff
@@ -38,27 +38,70 @@ func (b *Bslack) getChannel(channel string) (*slack.Channel, error) {
...
+	}
+	for {
+		channels, nextCursor, err := b.sc.GetConversations(queryParams)
+		if err != nil {
+			b.Log.Errorf("Could not reload channels: %#v", err)
+			return
 		}
+		for _, channel := range channels {
+			newChannelsByID[channel.ID] = &channel
+			newChannelsByName[channel.Name] = &channel
```
42wim: Seems that `IsIM:true` channels all have the same `""` name

bridge/slack/helpers.go
Helcaraxan: 😱 who did I miss that. Anyway... corrected 😅

bridge/slack/helpers.go
Helcaraxan: Ok. Will look into that tomorrow. 👍

42wim: also channel gets reused in this loop (this is why my test bridge failed), it's always the same pointer
possible solution
``` 
for _, channel := range channels {
                lchannel := channel
		newChannelsByID[channel.ID] = &lchannel
		newChannelsByName[channel.Name] = &lchannel
}
```

Helcaraxan: Had taken your suggestion but updated it to what I propose now:
```go
		for i := 0; i < len(channels); i++ {
			newChannelsByID[channels[i].ID] = &channels[i]
			newChannelsByName[channels[i].Name] = &channels[i]
		}
```

This has the advantage over the one proposed above of not having any copies, instead of two copies per channel that is iterated over (first copy into the `channel` variable, second into the iterations instance of `lchannel`).

42wim: Yes, good thinking, but it comes with a loss of readability. (You could also do `for i := range channels { ... }` )
But I'm not going to nitpick over this, it's not like my code is that readable either :)
Looks good to merge now, thanks!

42wim: Oops, not yet, the issue with the `IsIm:true` channels still exist

PR #517, opened 2018-10-12
Slack bridge file shuffle

Helcaraxan:
# Summary
Distributing the source of the Slack bridge across multiple files to
increase readability and as a prelude to various refactors and
clean-ups.

This is part of work for #510.

# Functionality / semantics changes
There are no code changes in this PR, just copy-paste of existing code.

PR #511, opened 2018-10-07
Update of nlopes/slack dependency

Helcaraxan:
Result of running:
> vgo get -u github.com/nlopes/slack
> vgo mod tidy
> vgo mod vendor

Prerequisite for work on #510

PR #531, opened 2018-10-22
Clean-up message send logic.

Helcaraxan:
Four down (#511, #517, #519, #521), three more to go for #510 and it's accompanying (p)refactor.

This is fairly easy one that deals mostly with breaking up the single `Send()` method into a main `Send()` "selector" function and two implementation ones (the existing `sendWebhook()` and a new `sendRTM()` (with a further lower-level utility function as well).

The only other significant change is part of this new `sendRTM()` method: instead of using the slack client of the bridge to do REST calls to Slack we actually use the RTM websocket.

The two other remaining PRs after this one will be:
- Large-scale rewrite of the incoming event handling: splitting out the massive `handleMessageEvent()` method into multiple smaller ones with distinct purposes.
- Rewrite of the `Connect()` method for simplification and with warning for use of legacy-tokens.

### DIFF
--- a/bridge/slack/slack.go
+++ b/bridge/slack/slack.go
@@ -165,6 +165,10 @@ func (b *Bslack) JoinChannel(channel config.ChannelInfo) error {
 	return nil
 }
 
+func (b *Bslack) Reload(cfg *bridge.Config) (string, error) {
+	return "", nil
+}
+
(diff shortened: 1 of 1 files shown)


42wim:
👍

Issue #522, opened 2018-10-14
Upgrade blackfriday markdown-parsing library to v2


patcon:
v1 has the markdown parser as private, but in v2 it's public.

This would allow me to customize the markdown parser for slackdown (slack's simplified flavor of markdown). Parsing slack's messages into HTML with this parser would allow Google Translate to work better (since it can understand and translate HTML, but not markdown). The HTML message can be translated, then rendered back into slackdown format for rendering on the other side of bridge.

I would expect to do this work myself, just hoping to a +1 before I tackle it.

(I will close out #499 first)

Related: https://github.com/42wim/matterbridge/issues/499

Helcaraxan:
@patcon have created the PR for this.

PR #532, opened 2018-10-22
Update Blackfriday dependency

Helcaraxan:
Addresses #522.

Includes fixup to the Telegram bridge implementation to use the updated dependency.


bridge/telegram/html.go
```diff
@@ -53,13 +53,16 @@ func (options *customHTML) ListItem(out *bytes.Buffer, text []byte, flags int) {
 }
 
 func makeHTML(input string) string {
```
Helcaraxan: This is the only code change part of Matterbridge. The rest is just changes in `vendor/`.

42wim:
Thanks!

Issue #527, opened 2018-10-16
Slack bridge relays ephemeral messages sent to bridge user


patcon:
**Describe the bug**
When using a full user account as the bridging Slack account, some ephemeral messages that are sent to this user (ostensibly by Slackbot) are relayed through the bridge.

**To Reproduce**
1. Create bridge user
2. Create gateway between two slack channels on the same slack team.
3. upload a file to one channel
4. The other channel will get a message that show as from slackbot in the logs

**Expected behavior**
Would expect messages from slackbot to be discarded.

**Screenshots/debug logs**
Sorry, forgot to grab these. Will try to reproduce later

**Environment (please complete the following information):**
 - OS: linux
 - Matterbridge version: `1.11.4-dev`
 - If self compiled: output of `git rev-parse HEAD` Sorry, I am technically way down on my own feature branch. There is a small chance that it's not reproducible, but I don't think I touched any code related to this.

**Additional context**
Please add your configuration file  (be sure to exclude or anonymize private data (tokens/passwords))

Helcaraxan:
@patcon are you still seeing this happen? Just checking to see whether we should close this.

patcon:
Yep, it addressed perfectly fine with your commit that ignores slackbot by name :)

though I think you use display name instead of more authoritative user ID, which I believe is consistently "USLACKBOT". Also, we should perhaps later investigate if the larger superset of messages that should be ignored are all ephemeral messages from any source, not just slackbot. Some other apps might somehow send the not user an ephemeral message (prob only affects legacy tokens)

So yes good to close and i could reticket those things :)

Helcaraxan:
I'll let you indeed reticket things that you think are worth it. 🙂

patcon:
Oh! Wait. Slackbot does autoreplies, so ignoring all messages from them might be bad. We can check this by setting up an auto-reply in the "Customise Slack" settings page, and see if it gets through the gateway. Will do this before closing

PR #533, opened 2018-10-22
Refactor and clean-up handlers.

Helcaraxan:
Five down (#511, #517, #519, #521, #531), two more to go for #510 and it's accompanying (p)refactor. This PR also has a proposed one-line fix for #527 (see comment on the given line).

# Disclaimer
This is the penultimate and probably largest, most complex PR of the bunch. The obvious goal is to preserve most, if not all, of the existing behaviour. Unfortunately I do not have the setup to do the necessary testing that would make me 100% sure that I am not breaking anything (_note to myself: introducing more solid unit-testing for the Slack bridge is a good next project ™️_). Current confidence factor is 80-90% given how many times I've analysed the code-path changes that I made and only external third-party review will be able to spot other things from here.

# Summary
The goal of this PR is to break up the massive `handleEventMessage()` and `skipMessageEvent()` method which currently have [cyclomatic complexities](https://en.wikipedia.org/wiki/Cyclomatic_complexity) of respectively 40+ and 20+ whereas best practices and linting (`gocyclo`) usually start to scream at anything above 10. This PR leaves only the `skipMessageEvent()` method with a complexity of 11 and all the rest of the methods <=10.

# Methodology
To accomplish these goals I have sliced up the existing code in independent logical units that each accomplish a part of the handling and filtering process. I've analysed the existing flow and the interdependence (or lack there of) of the various lines in the current function. This has brought me to the following pipeline and the corresponding implementation changes: each step now has its own sub-method (and potentially even sub-sub-methods for lower-level logical partitioning).

1. Check if the message should be ignored (_i.e._ the `skipMessageEvent()` method.
   - The `skipMessageEvent()` sub-method has been simplified by finding shorter equivalents for each existing logical expression.
   - The move of the skipping of particular attached files to step 5 of the pipeline in the `handleDownloadFile()` sub-sub-method.
2. Populate the Matterbridge message that will be sent to the router based on the received event and logic that is common to all events that are not skipped.
   - The `populateReceivedMessage()` sub-method.
   - The `populateMessageWithUserInfo()` sub-sub-method.
3. Detect and handle any message that is "status" related (think join channel, etc.). This might result in an early exit from the pipeline and passing of the pre-populated message to the Matterbridge router.
   - The `handleStatusEvent()` sub-method which has its logic condensed into a single switch statement.
4. Handle the specific case of messages that edit existing messages depending on configuration.
   - Straightforward and contained in the top-level `handleMessageEvent()` method.
5. Handle any attachments of the received event.
   - Separate `handleAttachements()` sub-method which calls the existing `handleDownloadFile()` sub-sub-method.
   - Attachments are now ignored at the sub-sub-method level if they are in the cache and recent.
6. Check that the Matterbridge message that we end up with after at the end of the pipeline is valid.
   - Straightforward and contained in the top-level `handleMessageEvent()` method.

NB: A digest of the above has been added to the code in the form of a comment on the `handleMessageEvent()` function.

### DIFF
--- a/bridge/slack/handlers.go
+++ b/bridge/slack/handlers.go
@@ -1,7 +1,6 @@
 package bslack
 
 import (
-	"bytes"
 	"fmt"
 	"html"
 	"regexp"
@@ -61,7 +60,15 @@ func (b *Bslack) handleSlackClient(messages chan *config.Message) {
 		case *slack.OutgoingErrorEvent:
 			b.Log.Debugf("%#v", ev.Error())
 		case *slack.ChannelJoinedEvent:
+			// When we join a channel we update the full list of users as
+			// well as the information for the channel that we joined as this
+			// should now tell that we are a member of it.
 			b.populateUsers()
+
+			b.channelsMutex.Lock()
+			b.channelsByID[ev.Channel.ID] = &ev.Channel
+			b.channelsByName[ev.Channel.Name] = &ev.Channel
+			b.channelsMutex.Unlock()
 		case *slack.ConnectedEvent:
 			b.si = ev.Info
 			b.populateChannels()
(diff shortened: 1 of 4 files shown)


bridge/slack/handlers.go
```diff
@@ -90,248 +98,170 @@ func (b *Bslack) handleMatterHook(messages chan *config.Message) {
...
+	case sChannelLeave, sChannelJoin:
+		return b.GetBool(noSendJoinConfig)
+	case sPinnedItem, sUnpinnedItem:
+		return true
 	}
-	// actually download the file
-	data, err := helper.DownloadFileAuth(file.URLPrivateDownload, "Bearer "+b.GetString(tokenConfig))
-	if err != nil {
-		return fmt.Errorf("download %s failed %#v", file.URLPrivateDownload, err)
+
```
Helcaraxan: This additional `ev.Username == sSlackBotUser` check below should be a fix for #527 (cc @patcon).

bridge/slack/handlers.go
```diff
@@ -90,248 +98,170 @@ func (b *Bslack) handleMatterHook(messages chan *config.Message) {
...
-		if ev.SubMessage.Edited == nil {
-			return true
-		}
+	// Actually download the file.
+	data, err := helper.DownloadFileAuth(file.URLPrivateDownload, "Bearer "+b.GetString(tokenConfig))
+	if err != nil {
+		return fmt.Errorf("download %s failed %#v", file.URLPrivateDownload, err)
 	}
 
-	for _, f := range ev.Files {
```
Helcaraxan: This piece of logic has been removed from the `skipMessageEvent()` method (which as moved higher-up in the file).

As far as my understanding goes the old logic below could have resulted in a message with multiple attachments being skipped if only one of the files was present and fresh in the cache although others were not.

Hence I've moved the verification to be done within `handleDownloadFile()` which is called separately for each file.

bridge/slack/slack.go
```diff
@@ -302,6 +307,39 @@ func (b *Bslack) sendRTM(msg config.Message) (string, error) {
 	return "slack " + id, nil
 }
 
+// handleUploadFile handles native upload of files
```
Helcaraxan: This file was copy-pasted without changes from the `handlers.go` file as it is actually not a handler and part of the outgoing pipeline, not the incoming one.

bridge/slack/handlers.go
patcon: Thanks! Should all ephemeral messages from bots be ignored? Or even all ephemeral messages? to the bot/slack user?

bridge/slack/handlers.go
```diff
@@ -61,7 +60,16 @@ func (b *Bslack) handleSlackClient(messages chan *config.Message) {
 		case *slack.OutgoingErrorEvent:
 			b.Log.Debugf("%#v", ev.Error())
 		case *slack.ChannelJoinedEvent:
+			// When we join a channel we update the full list of users as
+			// well as the information for the channel that we joined as this
+			// should now tell that we are a member of it.
 			b.populateUsers()
+
+			b.channelsMutex.Lock()
+			defer b.channelsMutex.Unlock()
```
42wim: This will never get unlocked?
Should be unlock() after you add the channel

bridge/slack/handlers.go
```diff
@@ -90,248 +98,177 @@ func (b *Bslack) handleMatterHook(messages chan *config.Message) {
...
 func (b *Bslack) handleMessageEvent(ev *slack.MessageEvent) (*config.Message, error) {
-	var err error
-
-	// update the userlist on a channel_join
-	if ev.SubType == sChannelJoin {
-		b.populateUsers()
-	}
-
-	// Edit message
-	if !b.GetBool(editDisableConfig) && ev.SubMessage != nil && ev.SubMessage.ThreadTimestamp != ev.SubMessage.Timestamp {
```
42wim: Any reason why you removed this `!b.GetBool(editDisableConfig)` in your diff ? We can drop this here if we aren't going to do anything with edits

bridge/slack/helpers.go
```diff
@@ -106,6 +107,65 @@ func (b *Bslack) populateChannels() {
...
+	}
+	return rmsg, err
+}
+
+func (b *Bslack) populateMessageWithUserInfo(ev *slack.MessageEvent, rmsg *config.Message) error {
+	if ev.SubType == sMessageDeleted || ev.SubType == sFileComment {
+		return nil
+	}
+
+	if ev.BotID != "" && b.GetString(tokenConfig) != "" {
```
42wim: We should check `b.GetString(outgoingWebhookConfig) == ""` like in the original code, because people can (and do) combine outgoing webhook for sending messages and token for receiving.
Don't do any of this stuff when we have a `outgoingWebhookConfig` because we're not sure if it's our webhook or not

bridge/slack/helpers.go
```diff
@@ -106,6 +107,65 @@ func (b *Bslack) populateChannels() {
...
+		if bot.Name != "" && bot.Name != "Slack API Tester" {
+			rmsg.Username = bot.Name
+			if ev.Username != "" {
+				rmsg.Username = ev.Username
+			}
+			rmsg.UserID = bot.ID
+		}
+	}
+
+	if rmsg.Username == "" && ev.User != "" {
```
42wim: same comment here as above, I'm not 100% sure anymore but there was a reason :)

42wim:
Nice work! 👍 
Some small comments, and then we can test it.
Are you a willing victim @patcon to help test this, you're an avid user of the slack bridge ?

bridge/slack/handlers.go
Helcaraxan: 🤦‍♂️

bridge/slack/helpers.go
Helcaraxan: What do you specifically mean here by "our webhook"?

----

May I ask what the reasons is for using webhooks for sending when already having an RTM connection available?

Context: In [my last follow-up commit](/Helcaraxan/matterbridge/commit/a7967527ceb436d082204e8d42884f849c65feb6) I was planning on simplifying the connection logic to use RTM for both sending and receiving when a token is available and only use webhooks if specified and if there is no token. This is also the reason why, maybe too early I flipped over the logic to check for the presence of the RTM token.

bridge/slack/helpers.go
Helcaraxan: I am trying to find a possible explanation but I really can't come up with anything. According to the old code's comment:
```go
  // when using webhookURL we can't check if it's our webhook or not for now
```

However there is no later point either where we check this either so I don't understand the "not for now" part.

That said according to https://api.slack.com/custom-integrations/outgoing-webhooks we should definitely be able to verify the user's identity as the `ev.Username` field will be set instead of the `ev.User` one. I've updated the accordingly.

bridge/slack/helpers.go
42wim: Users want to have this option because with the webhooks they can "impersonate" a user, (changing the IconURL (avatar) and changing the username) with the RTM, you can't change your username or the icon. That's why we allow the use of both options (in a lot of other bridges too)

bridge/slack/handlers.go
42wim: The `not been dropped` link doesn't point to anything :(

bridge/slack/helpers.go
42wim: Seems to be added in commit bd97357f8d30322dc9c9745c47b9461d862844ba
IIRC, the problem is that if we see messages in slack where `ev.BotID` is set and we also send messages to that channel via webhooks, we couldn't figure out if it was our own webhook that send the message or if it was another bot. So we just skip those message from other bots if we're using webhooks ourselves.

bridge/slack/handlers.go
Helcaraxan: Sorry don't know why it messed up... basically it's at line 154 in the new version of the file.

Helcaraxan:
@42wim I have reworked the `populateMessageWithUserInfo()` based on your feedback. And it made me realise that there was some stuff wrong with user info management which I also fixed.

42wim:
I just also tried editing a slack message and the username is getting replaced by the slackuser ID (on the receiving bridge)
image
(haven't got the time right now to debug this further)

@patcon can you also test this PR in your setup ? because you have probably a lot of edge cases :)

patcon:
Haha tons. I really want to learn more about got testing so I'm not the only one hitting these :)

bridge/slack/helpers.go
```diff
@@ -106,6 +110,65 @@ func (b *Bslack) populateChannels() {
...
+			if ev.Username != "" {
+				rmsg.Username = ev.Username
+			}
+			rmsg.UserID = bot.ID
+		}
+	}
+
+	if ev.User != "" {
+		user := b.getUser(ev.User)
+		if user != nil {
```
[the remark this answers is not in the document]
Helcaraxan: I definitely _thought_ I had tested it... but this would tend to prove the contrary. 😱 . Fixing.

Helcaraxan:
@42wim don't worry about debugging it further yourself. This is my mess (and fault for coding with my brain on automatic). Will address it and ping you when ready.

Helcaraxan:
@42wim have had a bit of a further look and I think it works now as expected w.r.t the users, etc.

Here's some screenshots of my (little bit of) testing between Discord & Slack:
screenshot 2018-11-04 at 16 33 34
screenshot 2018-11-04 at 16 33 47

NB: This is for Slack while only using a bot token associated with an app. As you can see it can perfectly impersonate the user that is sending the message when posting to the target Slack channel.

42wim:
Just tried to run with `webhookURL` in slack (without token) and it panics.
Did some tests and it seems to be broken since commit 498377a230b8712792a47a61ad63ab746f7379d1

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x9d3a40]

goroutine 1 [running]:
github.com/nlopes/slack.(*Client).GetConversationsContext(0x0, 0x10999c0, 0xc00003a4f0, 0xc0000301c0, 0xe37940, 0xc0008e96b0, 0x40e1d8, 0x10, 0xcf6580, 0xe37901, ...)
        /nasfs/users/u0022146/go/pkg/mod/github.com/nlopes/slack@v0.4.0/conversation.go:462 +0xb0
github.com/nlopes/slack.(*Client).GetConversations(0x0, 0xc0000301c0, 0x17d1820, 0xbef044aeda696381, 0x641b74, 0x17d1820, 0x0, 0x0, 0xc000776360)
        /nasfs/users/u0022146/go/pkg/mod/github.com/nlopes/slack@v0.4.0/conversation.go:456 +0x51
github.com/42wim/matterbridge/bridge/slack.(*Bslack).populateChannels(0xc0002328f0)
        /home/u0022146/git/go/src/github.com/42wim/matterbridge/bridge/slack/helpers.go:126 +0x352
github.com/42wim/matterbridge/bridge/slack.(*Bslack).JoinChannel(0xc0002328f0, 0xc0004624c6, 0x7, 0xc000462480, 0xa, 0xe8452d, 0x5, 0xc0005651a0, 0x11, 0xc0005744b0, ...)
        /home/u0022146/git/go/src/github.com/42wim/matterbridge/bridge/slack/slack.go:155 +0x2f
github.com/42wim/matterbridge/bridge.(*Bridge).joinChannels(0xc0001733b0, 0xc000575380, 0xc0005753b0, 0xceb4e0, 0xc000764230)
        /home/u0022146/git/go/src/github.com/42wim/matterbridge/bridge/bridge.go:61 +0x2f0
github.com/42wim/matterbridge/bridge.(*Bridge).JoinChannels(0xc0001733b0, 0x0, 0x0)
        /home/u0022146/git/go/src/github.com/42wim/matterbridge/bridge/bridge.go:53 +0x3d
github.com/42wim/matterbridge/gateway.(*Router).Start(0xc00056b360, 0xc00056b360, 0x0)
        /home/u0022146/git/go/src/github.com/42wim/matterbridge/gateway/router.go:53 +0x392
main.main()
        /home/u0022146/git/go/src/github.com/42wim/matterbridge/matterbridge.go:52 +0x5b0

Helcaraxan:
@42wim have indeed reproduced as well myself now (you need indeed to remove the token). Will make a separate fix PR to fast-track it.

patcon:
I rebased way back and ran into some issues (from other PR, I recall), but I'll give this another shot with the translation branch.

Helcaraxan:
Now rebased on top of the fixed `master`.

42wim:
Hm looks like it still needs a rebase, not sure how this can be if you rebased 3 hours ago

Helcaraxan:
Probably due to the merge you just did of #508. But I'll rebase it in a moment.

Helcaraxan:
Well. It wasn't that merge... Don't really know what confused the merge but it's fixed now. 😆

42wim:
Thanks! Did some small tests and looks good 🎉 
@patcon could you try master for a while to see if there isn't something that got overlooked?

patcon:
Hitting this on my crazy translation branch: https://gist.github.com/patcon/1cc9983db575b3c27de4d9f8cdd84948

     time="2018-11-08T01:13:04Z" level=error msg="&errors.errorString{s:"could not find information for user with id U4MCXJKNC"}" prefix=sla

Will investigating later. Was working with master merged in right up until the last few PRs were merged.

Issue #487, opened 2018-08-26
{USERNAME} placeholder in config


realrolfje:
**Is your feature request related to a problem? Please describe.**
After upgrading to mattermost 5.2.1 and the latest matterbridge, we notice that the behavior of {NICK} in the config file is changed.  It used to display the username of the mattermost user, and now it displays the nickname of the mattermost user. This feels like a bugfix and correct, but we'd like to display the username of the mattermost user. 

**Describe the solution you'd like**
For that, I'd like to ask for a {USERNAME} placeholder in the config file so we can reinstate our original setup.

realrolfje:
Possibly related to https://github.com/42wim/matterbridge/issues/441

42wim:
Related to https://github.com/42wim/matterbridge/issues/452 ;-)
But a `{USERNAME}` placeholder is a good idea

PR #515, opened 2018-10-23
Allow origin CHANNEL to be used in RemoteNickFormat

patcon:
Super simple. Will reference this from the commit to be cherry-picked in a sec.

### DIFF
--- a/matterbridge.toml.sample
+++ b/matterbridge.toml.sample
@@ -1245,6 +1245,7 @@ RemoteNickFormat="{NICK}"
 #The string "{BRIDGE}" (case sensitive) will be replaced by the sending bridge
 #The string "{LABEL}" (case sensitive) will be replaced by label= field of the sending bridge
 #The string "{PROTOCOL}" (case sensitive) will be replaced by the protocol used by the bridge
+#The string "{GATEWAY}" (case sensitive) will be replaced by the origin gateway name that is replicating the message.
 #The string "{CHANNEL}" (case sensitive) will be replaced by the origin channel name used by the bridge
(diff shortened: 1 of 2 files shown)


patcon:
Hopefully not an issue that I tried to reduce doc duplication too

matterbridge.toml.sample
```diff
@@ -1298,6 +1245,7 @@ RemoteNickFormat="{NICK}"
 #The string "{BRIDGE}" (case sensitive) will be replaced by the sending bridge
 #The string "{LABEL}" (case sensitive) will be replaced by label= field of the sending bridge
 #The string "{PROTOCOL}" (case sensitive) will be replaced by the protocol used by the bridge
+#The string "{CHANNEL}" (case sensitive) will be replaced by the origin channel name used by the bridge
```
Helcaraxan: Had forgot to document my previous addition of `GATEWAY` in #501.

```suggestion
#The string "{GATEWAY}" (case sensitive) will be replaced by the origin gateway name that is 
replicating the message.
#The string "{CHANNEL}" (case sensitive) will be replaced by the origin channel name used by the bridge
```

Helcaraxan:
Nice work on the deduplication!

Helcaraxan:
Might actually be worth going the extra mile and solving #487 in the same commit.

patcon:
@Helcaraxan heh, thanks. As for the other thing, it's not immediately clear to me how USERNAME / NICK differs across all bridge types, so I think I'll treat that as out-of-scope for me :)

42wim:
👍

PR #537, opened 2018-10-23
Add editorconfig support

patcon:
Just makes it so that IDEs automatically use the right indentation.

https://editorconfig.org/

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,2 @@
+[*.go]
+indent_style=tab


42wim:
Sorry, strong opinions about this :)
Using go you should use an editor that runs gofmt automatically after each save, no editorconfig needed.

patcon:
I can respect that. I've now moved this gitconfig to user-level, and so it resolves it for me :) it's easy enough, but I submitted this in consideration that i worked for months without having this preferred golang workflow you mention ;) and I don't think that would be uncommon for newcomer golang contributors

But totally minor, and happy to defer

PR #539, opened 2018-10-25
Prevent Slack API rate-limit overflow

Helcaraxan:
I noticed that the current implementation of the Slack bridge is actually likely to exceed Slack's API rate-limit.

When starting up and joining multiple channels multiple JoinChannel events are received which all result in the users and channels being refreshed.

### DIFF
--- a/bridge/slack/helpers.go
+++ b/bridge/slack/helpers.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/nlopes/slack"
 )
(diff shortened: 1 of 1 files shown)


bridge/slack/handlers.go
```diff
@@ -64,8 +64,12 @@ func (b *Bslack) handleSlackClient(messages chan *config.Message) {
...
 			b.si = ev.Info
-			b.populateChannels()
-			b.populateUsers()
+
+			// Populating channels and users can take a while so do it async.
+			go func() {
+				b.populateChannels()
+				b.populateUsers()
+			}()
 		case *slack.InvalidAuthEvent:
```
42wim: This was sync with a reason, so that we kept blocking here and don't start receiving messsages where don't have the info about the channel or users. Maybe for big channels (because that takes a long time with the new cursor API) we can do ad-hoc lookup of a user instead (or combined with) loading them all at once. 
The populateChannels() shouldn't take long and can be kept sync.

WDYT ?

Helcaraxan: You are quite right and in retrospective I am removing the async stuff entirely. Just limiting this PR to the scope of not running multiple `populateChannels()` and `populateUsers()` in parallel.

42wim:
Wouldn't it be better to have one function that gets called from populateUsers and populateChannels ? with just one mutex / time ?

Helcaraxan:
In the `handleSlackClient()` method we call the `populateUsers()` without calling the `populateChannels()` hence they are split out in two functions. This in turn means that they both need to rely on separate mutex / time couples.

42wim:
Maybe I'm overlooking something, but as the slack API rate is global, why not something like this ?
```
func refreshAllowed(what string) bool {
	if time.Now().Before(earliestRefresh) {
		b.Log.Debugf("Not refreshing %s list as there was a refresh less than %d seconds ago.",what, int(minimumRefreshInterval.Seconds()))
		return false
	}
	earliestRefresh = time.Now().Add(minimumRefreshInterval)
	return true
}

populateUsers() {
	if !refreshAllowed() { return }
	...
}
	
populateChannels() {
	if !refreshAllowed() { return }
	...
}
```

42wim:
Thanks!

Issue #343, opened 2018-01-18
Slack: When unthreading messages, quote previous message


patcon:
Currently, when two Slack teams are connected, threaded message responses are unthreaded into regular messages. This results in lost context.

**I wonder whether we could use quotes in messages to give these messages more context.** I'm unclear if this would be simple to do, or if it would require extra calls to slack API.

This quote could be either the original message of thread, or the one just prior. (The original message might make more sense, and would be possible to CTRL+F through logs and find all the responses.)

This would be particularly useful for responses to older messagemiis, for which unthreading is quite confusing.

Currently, mitigating this issue is only possible with social conventions that are hard to enforce: for example, 1) people avoiding threading, or 2) being explicit about which message you're responding to. That isn't ideal :)

42wim:
Looked into it, would need lots of changes, extra state keeping. 
This is nothing that can/will be added short-term

patcon:
Thanks for investigating :)

Just to bump a half-formed idea in here -- what if:

1. there was a flag that would cause slack messages to do lookups in the ring cache for original messages on thread replies
    - successful lookup would cause slack messages to be prefixed with

      ```
      > The first ten words of the original thread message so...

      The unthreaded message that currently gets relayed on its own.
      ```
2. as is currently possible, ring cache buffer could be bumped with a protocol-level config for `Buffer=<big number>` in the toml

Would this be a possible path forward, when the time comes. The slight but of context would very helpfully hint the reader to the originated message, and they could use built-in search of the backscrool if they don't recall the rest :)

42wim:
Well, the ring cache is only used by the api bridge.
The other cache doesn't keep messages text, only IDs. 
Extending it to keep messages could be possible though.

42wim:
Has been fixed by supporting threading in slack I guess?

patcon:
This was to avoid confusion when cache is lost and unthreads, but fine to close, since I'm not working on it right now :)

PR #529, opened 2018-10-27
Opportunistically preserve threading between Slack instances

patcon:
Even with clear documentation and encouragement of work-arounds, unthreading of messages between slacks leads to confusion and misunderstanding.

Would be great to preserve threading between Slack instances, when possible. Presumably this would be when the parent message is still in cache.

- Clear documentation
- https://github.com/42wim/matterbridge/issues/343

**Additional context**
Downside: There would be inconsistency: sometimes a channel-level message would have originated as a channel-level message, but sometimes it would be a thread-level message where there wasn't enough detail in cache to rethread. And the oldest replies (not in cache) would be the most common to unthread, which are the ones most likely to cause confusion anyhow,

### DIFF
--- a/bridge/slack/slack.go
+++ b/bridge/slack/slack.go
@@ -310,7 +310,7 @@ func (b *Bslack) prepareMessageParameters(msg *config.Message) *slack.PostMessag
 	params.Username = msg.Username
 	params.LinkNames = 1 // replace mentions
 	params.IconURL = config.GetIconURL(msg, b.GetString(iconURLConfig))
-	params.ThreadTimestamp = msg.ThreadTs
+	params.ThreadTimestamp = msg.ParentID
 	if msg.Avatar != "" {
 		params.IconURL = msg.Avatar
 	}
--- a/bridge/slack/slack.go
+++ b/bridge/slack/slack.go
@@ -310,6 +310,7 @@ func (b *Bslack) prepareMessageParameters(msg *config.Message) *slack.PostMessag
 	params.Username = msg.Username
 	params.LinkNames = 1 // replace mentions
 	params.IconURL = config.GetIconURL(msg, b.GetString(iconURLConfig))
+	params.ThreadTimestamp = msg.ThreadTs
 	if msg.Avatar != "" {
 		params.IconURL = msg.Avatar
 	}
--- a/bridge/slack/slack.go
+++ b/bridge/slack/slack.go
@@ -312,10 +312,7 @@ func (b *Bslack) prepareMessageParameters(msg *config.Message) *slack.PostMessag
 	params.IconURL = config.GetIconURL(msg, b.GetString(iconURLConfig))
 	msgFields := strings.Fields(msg.ParentID)
 	if len(msgFields) >= 2 {
-		msgProtocol, msgID := msgFields[0], msgFields[1]
-		if msgProtocol == "slack" {
-			params.ThreadTimestamp = msgID
-		}
+		params.ThreadTimestamp = msgFields[1]
 	}
 	if msg.Avatar != "" {
 		params.IconURL = msg.Avatar
--- a/bridge/slack/slack.go
+++ b/bridge/slack/slack.go
@@ -310,7 +310,13 @@ func (b *Bslack) prepareMessageParameters(msg *config.Message) *slack.PostMessag
 	params.Username = msg.Username
 	params.LinkNames = 1 // replace mentions
 	params.IconURL = config.GetIconURL(msg, b.GetString(iconURLConfig))
-	params.ThreadTimestamp = msg.ParentID
+	msgFields := strings.Fields(msg.ParentID)
+	if len(msgFields) >= 2 {
+		msgProtocol, msgID := msgFields[0], msgFields[1]
+		if msgProtocol == "slack" {
+			params.ThreadTimestamp = msgID
+		}
+	}
 	if msg.Avatar != "" {
 		params.IconURL = msg.Avatar
 	}
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -312,11 +312,9 @@ func (gw *Gateway) handleMessage(msg config.Message, dest *bridge.Bridge) []*BrM
 			msg.Channel = originchannel
 		}
 
-		if canonicalParentMsgID != "" {
-			msg.ParentID =  gw.getDestMsgID(canonicalParentMsgID, dest, channel)
-			if msg.ParentID == "" {
-			   msg.ParentID = canonicalParentMsgID
-			}
+		msg.ParentID =  gw.getDestMsgID(canonicalParentMsgID, dest, channel)
+		if msg.ParentID == "" {
+		   msg.ParentID = canonicalParentMsgID
 		}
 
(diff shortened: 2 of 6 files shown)


patcon:
Confirmed that's it's messages without attachments that give the error on mobile. This was masked for me since all our translations have attachments with the original text. Hmm....

patcon:
Sent a support request for feedback@slack.com (will post any reply)

gateway/gateway.go
```diff
@@ -276,10 +305,31 @@ func (gw *Gateway) handleMessage(msg config.Message, dest *bridge.Bridge) []*BrM
...
+					// for people that reuse the same bridge multiple times. see #342
+					if dest.Protocol == id.br.Protocol && dest.Name == id.br.Name && channel.ID == id.ChannelID {
+						parentMsgID = IDs[i].ID
+						break
+					}
+				}
+				if parentMsgID == "" {
+					parentMsgID = canonicalParentMsgID
+				}
+				msg.ParentID = strings.Fields(parentMsgID)[1]
```
42wim: Is bridge specific, split should be inside slack (see above comment)

bridge/slack/slack.go
```diff
@@ -310,6 +310,7 @@ func (b *Bslack) prepareMessageParameters(msg *config.Message) *slack.PostMessag
 	params.Username = msg.Username
 	params.LinkNames = 1 // replace mentions
 	params.IconURL = config.GetIconURL(msg, b.GetString(iconURLConfig))
+	params.ThreadTimestamp = msg.ParentID
```
42wim: Do the split here
```
params.ThreadTimestamp = strings.Fields(msg.ParentID)[1]
```

gateway/gateway.go
```diff
@@ -276,10 +305,31 @@ func (gw *Gateway) handleMessage(msg config.Message, dest *bridge.Bridge) []*BrM
 		if dest.Protocol == "api" {
 			msg.Channel = originchannel
 		}
+
+		if canonicalParentMsgID != "" {
+			parentMsgID := ""
+			if res, ok := gw.Messages.Get(canonicalParentMsgID); ok {
```
42wim: This is code duplication, better to create a new function eg `getDestMsgID` which can also replace the code on line 265
```
msg.ParentID = getDestMsgID(dest.Name,dest.Protocol,channel.ID,canonicalParentMsgID)
if msg.ParentID == "" {
   msg.ParentID = canonicalParentMsgID
}
```

```
func getDestMsgID(name string, protocol string, channelID string, origID string) string {
	if res, ok := gw.Messages.Get(origmsg.ID); ok {
		IDs := res.([]*BrMsgID)
		for _, id := range IDs {
			// check protocol, bridge name and channelname
			// for people that reuse the same bridge multiple times. see #342
			if protocol == id.br.Protocol && name == id.br.Name && channelID == id.ChannelID {
				return id.ID
			}
		}
	}
	return ""
}
```

patcon:
Re: slack bug. heard back from Slack within 12 hours, and they seem very interested in resolving this. I dropped the ball with my lack of reply, but will today send them the details they asked for

gateway/gateway.go
```diff
@@ -258,28 +298,32 @@ func (gw *Gateway) handleMessage(msg config.Message, dest *bridge.Bridge) []*BrM
...
-			IDs := res.([]*BrMsgID)
-			for _, id := range IDs {
-				// check protocol, bridge name and channelname
-				// for people that reuse the same bridge multiple times. see #342
-				if dest.Protocol == id.br.Protocol && dest.Name == id.br.Name && channel.ID == id.ChannelID {
-					msg.ID = id.ID
-				}
-			}
-		}
+		msg.ParentID = ""
```
42wim:
```suggestion
		msg.ParentID = gw.getDestMsgID(canonicalParentMsgID, dest, channel)
```
If canonicalParentMsgID is empty, msg.ParentID will be empty, otherwise it will be set

gateway/gateway.go
```diff
@@ -258,28 +298,32 @@ func (gw *Gateway) handleMessage(msg config.Message, dest *bridge.Bridge) []*BrM
...
 			msg.Channel = originchannel
 		}
+
+		if canonicalParentMsgID != "" {
+			msg.ParentID =  gw.getDestMsgID(canonicalParentMsgID, dest, channel)
+			if msg.ParentID == "" {
+			   msg.ParentID = canonicalParentMsgID
+			}
+		}
+
```
42wim: As in my suggestion above, msg.ParentID will be empty or have the correct id, so above can be changed to

```
if msg.ParentID == "" {
      msg.ParentID = canonicalParentMsgID
}
```

bridge/slack/slack.go
```diff
@@ -310,6 +310,13 @@ func (b *Bslack) prepareMessageParameters(msg *config.Message) *slack.PostMessag
 	params.Username = msg.Username
 	params.LinkNames = 1 // replace mentions
 	params.IconURL = config.GetIconURL(msg, b.GetString(iconURLConfig))
+	msgFields := strings.Fields(msg.ParentID)
+	if len(msgFields) >= 2 {
+		msgProtocol, msgID := msgFields[0], msgFields[1]
+		if msgProtocol == "slack" {
+			params.ThreadTimestamp = msgID
+		}
+	}
```
42wim: You can just do  (like you had in your original code), no need to doublecheck here, we should only be getting valid input.
```
params.ThreadTimestamp = msgFields[1]
```

gateway/gateway.go
```diff
@@ -303,7 +303,7 @@ func (gw *Gateway) handleMessage(msg config.Message, dest *bridge.Bridge) []*BrM
 		msg.Avatar = gw.modifyAvatar(origmsg, dest)
 		msg.Username = gw.modifyUsername(origmsg, dest)
 		msg.ID = ""
-		msg.ParentID = ""
```
patcon: Wait, I accepted this, but I'm pretty sure this needs to be "reset" to empty each time, or it persists between for gateways that have more than two bridges. Or is there something I'm missing?

bridge/slack/slack.go
patcon: that was failing, I assumed because some messages are coming through without an ID set in this way

patcon: And if there's an incorrect parentID, it doesn't just post to main channel, but loses the message completely. So resetting to "" at the start is important

patcon:
Ok, I was mistaken with some of my comments (sorry), and ended up taking all your suggestions into account. Let me know if you notice anything else

patcon:
And just so it's not lost in the shuffle, this commit is a significant one that we might want to cherry-pick into another PR: https://github.com/42wim/matterbridge/pull/529/commits/588f04fd01ea4daf44b984734760434b442f7218 (didn't mean to sneak it in, just wanted to get threading working on mobile before merge)

42wim:
As said in the other issue, I'm not going to merge 588f04fd01ea4daf44b984734760434b442f7218 wrt removing callback_id.
Could you please remove that commit from this PR ?

patcon:
Done and done :)

PR #545, opened 2018-10-29
When auto-loading envvars from viper config, remove hyphens

patcon:
See: https://unix.stackexchange.com/questions/23659/can-shell-variable-include-character

### DIFF
--- a/bridge/config/config.go
+++ b/bridge/config/config.go
@@ -185,7 +185,7 @@ func NewConfig(cfgfile string) *Config {
 	viper.SetConfigFile(cfgfile)
 	viper.SetEnvPrefix("matterbridge")
 	viper.AddConfigPath(".")
-	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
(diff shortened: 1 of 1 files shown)


bridge/config/config.go
```diff
@@ -185,7 +185,7 @@ func NewConfig(cfgfile string) *Config {
 	viper.SetConfigFile(cfgfile)
 	viper.SetEnvPrefix("matterbridge")
 	viper.AddConfigPath(".")
-	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
+	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_", "-", ""))
```
42wim: I think it's better to replace dash to underscore 
`viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_", "-", "_"))`

patcon: I guess I liked that the underscores mapped to the keyspace, but I can change it

42wim:
thanks!

PR #525, opened 2018-10-29
Allow toml configuration to be parsed from remote/local URL

patcon:
I'm thinking on how to create a web interface to manage the config, and my current thinking is just ~~the~~ to generate a small ~~all~~ app that access the config via basic auth.

Would you be opposed to this? That would allow me to start building adjacent things that don't need to impinge much on the larger project :)

### DIFF
--- a/bridge/config/config.go
+++ b/bridge/config/config.go
@@ -2,6 +2,8 @@ package config
 
 import (
 	"bytes"
+	"net/http"
+	"net/url"
 	"os"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


patcon:
Ok, so I'm realizing that not all of matterbridge features can work on Heroku atm: heroku has web and worker processes. In short, only web can bind to a port and receive requests. But it also sleeps, and workers are the processes that stay active like matterbridge needs.

So this might not be something you'd be willing to bring into core, but I'm currently investigating using queues. Specifically, queues that are exposed as regular go channels: https://medium.com/@matryer/introducing-vice-go-channels-across-many-machines-bcac1147d7e2

Channels are really new to me, so I'm curious your thoughts on all this. Thanks wim. Sorry for all the noise the last few days ;)

patcon:
heh, i was trying to create an app that could spread across multiple processes, as is advocated by heroku (and pressured by their free tier, due to limitations).

But I can understand that that's a hassle. I'll use the API bridge as you suggest (good idea!), and just pay heroku so that I can run this as a single process

patcon:
Ok, working on scaling this back in a local branch, but quick Q: 

> I think this fits better in the API bridge instead of creating another webhook..

I'm reading this to mean that each API bridge gets an `/api/reload` endpoint, and they trigger the same code path to do a reload of the configuration settings. Am I hearing you correctly?

patcon:
Sounds good to me. This would work for the common ephemeral filesystem setup of most PasS platforms, so long as we grab config as part of the process startup script in `matterbridge-heroku`:

re: https://devcenter.heroku.com/articles/dynos#ephemeral-filesystem

patcon:
Ok, so I'm realizing that I'm confused about our config tactic. Should I store a `ConfigURL` at some deeper base level viper config (outside scope of `General` key), or just store it on `General`?

It seems like it's supposed to be the latter right now, but that might result in wonky effects if someone tried to set it per-gateway (e.g. gateways fighting over what url the config is written from)

patcon:
Another potential issue of storing this url in config file itself (instead of setting as CLI flag or envvar)

It can easily set people up to find themselves in odd territory, because it fetches the file that points to the config that will be used on the next reload. So if someone is using a config file in a gist or github repo, the current config url will not be in the file -- only the last raw url from previous commits. And so trying to let config be managed in a git repo will result in the file "walking back in time" on each reload 😆 

Can be avoided by encouraging setting via envvar, but just a heads up. I'll make sure I document favoured approaches

patcon:
Still curious about the above clarifications though

PR #547, opened 2018-10-29
Use mainline gitter

patcon:
Features seem merged in


42wim:
As there's not much default upstream, I'm going to keep my fork for now.

PR #550, opened 2018-11-01
[API] respond with message on connect

### LINKED ISSUE #549: [API] send initial message to connecting httpClients instantly
it is hard to distinguish between timeout and a successful connection since the first data sent by the api is on message
and i would like to set the timeout until first read, but that is currentlytoo unpredictable

the first bit of datta returned in the stream should be sent instantly
it could be a plain string `connection open` or a message with `event = "api_connected"` or similar

having 2 api endpoints open with a "loopback" gateway allows me to post a message to api_1 to receive it in api_2 to initialize the connection,
but then that message needs to be handled specially and discarded, this also adds a very specific configuration

**Additional context**

NikkyAI:
fix #549

### DIFF
--- a/bridge/api/api.go
+++ b/bridge/api/api.go
@@ -102,6 +102,14 @@ func (b *Api) handleMessages(c echo.Context) error {
 func (b *Api) handleStream(c echo.Context) error {
 	c.Response().Header().Set(echo.HeaderContentType, echo.MIMEApplicationJSON)
 	c.Response().WriteHeader(http.StatusOK)
+	greet := config.Message{
+		Event:config.EVENT_API_CONNECTED,
+		Timestamp:time.Now(),
+	}
+	if err := json.NewEncoder(c.Response()).Encode(greet); err != nil {
+		return err
+	}
+	c.Response().Flush()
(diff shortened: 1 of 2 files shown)


bridge/api/api.go
```diff
@@ -102,6 +102,18 @@ func (b *Api) handleMessages(c echo.Context) error {
 func (b *Api) handleStream(c echo.Context) error {
 	c.Response().Header().Set(echo.HeaderContentType, echo.MIMEApplicationJSON)
 	c.Response().WriteHeader(http.StatusOK)
+	greet := config.Message{
+		Text: "Connected",
+		Channel: "",
+		Username: "system",
+		Event: "api_connect",
```
42wim: Could you add a `EVENT_API_CONNECTED="api_connected"` and change to 
```suggestion
		Event: config.EVENT_API_CONNECTED,
```
Any objections to only having an `event` and `timestamp` like 
`greet := config.Message{Event:config.EVENT_API_CONNECTED,Timestamp:time.Now()}`

NikkyAI: no objections

42wim:
Thanks!

PR #551, opened 2018-11-04
[WIP] Add auto-generated Swagger API docs

patcon:
Not ready for review


patcon:
Seems pretty much ready for review :)

- no way to disable "try it out button" on the stream endpoint, even though it won't work
- need to get clarity on which message object keys need to be set, and have good example values that will actually create a new message (can't get a message posting yet)
- perhaps need to find a way to disable the "auth" button if no token is set. It won't hurt to have it there when not needed, but might be misleading
- might be nice to have it hosted at `/api` or `/api/swagger` rather than exclusively `/api/swagger/index.html`, but not sure how to make this work yet. (I know swagger allows it, bc it's done here: http://petstore.swagger.io/)

patcon:
Also, I recall you saying not to merge go.* files, but seems like they're needed to have CI work...?

patcon:
- [ ] have CI run and compare against the `docs/` dir. If different, assume it wasn't rebuilt, and fail the CI run

patcon:
@42wim heyo, so travis is still failing due to the scripts not knowing how to deal with new packages. I really don't totally understand the vendoring process, but the main question is whether you support modifying the travis scripts so that they have a chance of passing when someone has added new packages or libraries :)

If you could either give that the thumbs up, or offer any advice, then I'd be grateful!

patcon:
One more question, @42wim: **Where does your demo matterbridge server live**, so that I can point the README docs to the live ones that will exist once merged?

patcon:
Ok, this is ready for review :)

patcon:
No no I totally get that. Was surprised at the deps too!

Technically, at minimum, this just needs a hosted json/yaml file, so for other projects, I've just served that, and used the publicly hosted spec renderer to make it explorable.

We would need to enable CORS requests for the UI to be functional. We could set up redirects as exist here, but pointing to external URL like this:
https://petstore.swagger.io/?url=https://raw.githubusercontent.com/patcon/matterbridge/swagger-api-docs/docs/swagger/swagger.yaml

(Has some errors due to assumptions of relative URLs)

This would have essentially no deps, but updating yaml is manual. I might suggest we leave in the doc comments, so that when someone does need to update, it's not tedious (this need not require any additional dep, as only the updater would install the package. Simple updates could be done manually by tweaking the yaml directly)

patcon:
Also, TODO for myself: confirm that hosted UI will work for localhost files for sane local development.

One caveat if this whole approach is that it will require the app to know its hosted domain or rely on it from HOST request headers

Helcaraxan:
Hey @patcon. Also weighing in with my 2c. What is the exact purpose of creating these docs? What is the problem that you see as being solved by this approach? And what are the alternative solutions that you have considered for that problem?

Without clarity on that I would personally not go for this approach as it will induce a non-negligible extra maintenance burden.

PR #554, opened 2018-11-05
Add a ping keepalive endpoint to API

patcon:
~~Depends on this: https://github.com/42wim/matterbridge/pull/551~~

Also, will be used for https://github.com/42wim/matterbridge/pull/525 in keeping the service from sleeping on Heroku

### DIFF
--- a/bridge/api/api.go
+++ b/bridge/api/api.go
@@ -39,6 +39,7 @@ func New(cfg *bridge.Config) bridge.Bridger {
 			return key == b.GetString("Token"), nil
 		}))
 	}
+	e.GET("/api/ping", b.handlePing)
 	e.GET("/api/messages", b.handleMessages)
 	e.GET("/api/stream", b.handleStream)
 	e.POST("/api/message", b.handlePostMessage)
@@ -73,6 +74,10 @@ func (b *Api) Send(msg config.Message) (string, error) {
 	}
 	b.Messages.Enqueue(&msg)
 	return "", nil
+}
+
+func (b *Api) handlePing(c echo.Context) error {
+	return c.String(http.StatusOK, "pong")
 }
 
 func (b *Api) handlePostMessage(c echo.Context) error {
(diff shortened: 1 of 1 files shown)


patcon:
Sorry, it was a little subtle, but I said that this depends on #551

patcon:
Ok, I've just rebased it, since I feel like the other PR might take some discussion :)

patcon:
:)

> If you're polling `/api/messages` or streaming `/api/stream` it shouldn't sleep

I don't expect to be polling them at all -- [as per your great suggestion](https://github.com/42wim/matterbridge/pull/525#issuecomment-434858085), I'm only using the API to kick off config reloads, but I don't plan to use it for messaging :)

> Wondering if it's not better to send a ping message to `/api/message`...

How were you thinking this might work? Like a special payload where a plaintext `pong` response comes back instead of a json response? Or using GET instead of POST?

> that's how slack/mattermost do it (with websockets on one endpoint though)

tbh I don't feel we should model on a websocket api, since RESTful API's aspire for semantic endpoints with homogenous resource objects returned (returning a consistent object per VERB), and websockets APIs don't necessarily abide by that (if I'm understanding correctly)

I'm inclined to vouch for having a ping endpoint, unless you feel strongly :)

patcon:
Not that this was my intention, but now that I'm thinking about it, some sort of ping endpoint is a good candidate for [turning into a healthcheck endpoint](https://blog.runscope.com/posts/build-a-health-check-api-for-services-underneath-your-api). I feel it conceptually makes sense to not conflate the message reading/writing endpoint with the utility purpose of a keepalive/healthcheck

42wim:
Agree for a healthcheck handler, rename it to `/health` and let it return `OK` instead `pong` for now.

Websockets can do whatever you want :-) mattermost for example streams all their stuff over websockets, and uses ping as a keepalive over the websocket.

patcon:
Done and done!

Issue #516, opened 2018-10-11
Allow sending isTyping events to other slack channels


patcon:
Related: https://github.com/nlopes/slack/blob/0afac90b917f77d7c59d638420d5d40fe65eee61/websocket_misc.go#L51-L56

While this would show the "bot" typing, I could be an indicator to show whether a reply might be expected soon.

patcon:
Use case here is when using the Google Translation feature (upcoming) to translate between channels. Would be great to know when someone is writing back. That's one of the only parts that doesn't feel seamless about the experience :)

42wim:
Fixed in #559

PR #559, opened 2018-11-06
Add ability to show when user is typing across Slack bridges

patcon:
Addresses \#516

Should be possible to get this working for Mattermost as well, just not sure how, so leaving it for another PR.

If your "Real Name" of your bot/user is set appropriately, the typing message even sounds semantic :)

screen shot 2018-11-06 at 11 58 42 pm

### DIFF
--- a/bridge/slack/handlers.go
+++ b/bridge/slack/handlers.go
@@ -23,7 +23,9 @@ func (b *Bslack) handleSlack() {
 	time.Sleep(time.Second)
 	b.Log.Debug("Start listening for Slack messages")
 	for message := range messages {
-		b.Log.Debugf("<= Sending message from %s on %s to gateway", message.Username, b.Account)
+		if message.Event != config.EVENT_USER_TYPING {
+			b.Log.Debugf("<= Sending message from %s on %s to gateway", message.Username, b.Account)
+		}
 
 		// cleanup the message
 		message.Text = b.replaceMention(message.Text)
--- a/bridge/slack/handlers.go
+++ b/bridge/slack/handlers.go
@@ -256,8 +256,6 @@ func (b *Bslack) handleTypingEvent(ev *slack.UserTypingEvent) (*config.Message, 
 		Channel:  channelInfo.Name,
 		Account:  b.Account,
 		Event:    config.EVENT_USER_TYPING,
-		ID:       "slack " + time.Now().String(),
-		Extra:    map[string][]interface{}{},
 	}
 
 	return &rmsg, nil
--- a/bridge/slack/handlers.go
+++ b/bridge/slack/handlers.go
@@ -254,10 +254,9 @@ func (b *Bslack) handleTypingEvent(ev *slack.UserTypingEvent) (*config.Message, 
 	}
 
 	rmsg := config.Message{
-		Username: b.getUsername(ev.User),
-		Channel:  channelInfo.Name,
-		Account:  b.Account,
-		Event:    config.EVENT_USER_TYPING,
+		Channel: channelInfo.Name,
+		Account: b.Account,
+		Event:   config.EVENT_USER_TYPING,
 	}
 
 	return &rmsg, nil
--- a/bridge/slack/handlers.go
+++ b/bridge/slack/handlers.go
@@ -46,6 +46,14 @@ func (b *Bslack) handleSlackClient(messages chan *config.Message) {
 			b.Log.Debugf("== Receiving event %#v", msg.Data)
 		}
 		switch ev := msg.Data.(type) {
+		case *slack.UserTypingEvent:
+			rmsg, err := b.handleTypingEvent(ev)
+			if err != nil {
+				b.Log.Errorf("%#v", err)
+				continue
+			}
+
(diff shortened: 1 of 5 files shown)


patcon:
Ready for review

bridge/slack/slack.go
```diff
@@ -245,6 +248,13 @@ func (b *Bslack) sendWebhook(msg config.Message) (string, error) {
 }
 
 func (b *Bslack) sendRTM(msg config.Message) (string, error) {
+	if msg.Event == config.EVENT_USER_TYPING {
```
42wim:
```suggestion
	if msg.Event == config.EVENT_USER_TYPING && b.GetBool("ShowUserTyping") {
```

bridge/slack/slack.go
```diff
@@ -245,6 +248,13 @@ func (b *Bslack) sendWebhook(msg config.Message) (string, error) {
...
+		if b.GetBool("ShowUserTyping") {
+			chanID := b.channelsByName[msg.Channel].ID
```
42wim: Don't read the map directly, locking issues. Use `getChannelByName()` function

bridge/slack/handlers.go
```diff
@@ -290,6 +298,27 @@ func (b *Bslack) handleMessageEvent(ev *slack.MessageEvent) (*config.Message, er
...
+	if err != nil {
+		return nil, err
+	}
+
+	rmsg := config.Message{
+		Username: b.getUsername(ev.User),
+		Channel:  channelInfo.Name,
+		Account:  b.Account,
+		Event:    config.EVENT_USER_TYPING,
+		ID:       "slack " + time.Now().String(),
```
42wim: Why setting an ID ?

bridge/slack/handlers.go
```diff
@@ -290,6 +298,27 @@ func (b *Bslack) handleMessageEvent(ev *slack.MessageEvent) (*config.Message, er
...
+		Extra:    map[string][]interface{}{},
```
42wim: Not needed, can be excluded

bridge/slack/handlers.go
```diff
@@ -290,6 +298,27 @@ func (b *Bslack) handleMessageEvent(ev *slack.MessageEvent) (*config.Message, er
...
+func (b *Bslack) handleTypingEvent(ev *slack.UserTypingEvent) (*config.Message, error) {
+	var err error
+	// use our own func because rtm.GetChannelInfo doesn't work for private channels
+	channelInfo, err := b.getChannelByID(ev.Channel)
+	if err != nil {
+		return nil, err
+	}
+
+	rmsg := config.Message{
+		Username: b.getUsername(ev.User),
```
42wim: Username not needed, not getting used anywhere ?

42wim:
* In go (and also matterbridge config) booleans are default always `false`, so `ShowUserTyping` should be too (also to not break default behaviour).
* I'm going to stop merging slack stuff at the moment until @Helcaraxan clean-up PR's are done, so that we don't make it more difficult for him and me :-)

bridge/slack/handlers.go
patcon: heh happy to leave out, just thought it was perhaps convention. I think I saw it in some debug output, but I'll double-check that.

We probably don't want to store these msgs in the cache, right?  (If I understand correctly, it could theoretically really inflate the cache size, and mean less history for edits/threading)

Related to that, would having a static ID or empty string help here, or should we have logic elsewhere to keep it out?

patcon:
haha yeah, i was thinking last night about how maybe i'm trying to send too much, too quick, and _then_ realized that big refactor was still in queue. I totally get it :)

bridge/slack/slack.go
patcon: heh just realized we fetched this right below

patcon:
> In go (and also matterbridge config) booleans are default always false, so ShowUserTyping should be too (also to not break default behaviour)

ok good call. Realized that the default I was setting was overriding things set on bridge, and in conflict with how `b.GetString()` etc work

I would be nice to have a clear way to set non-false defaults, as sometimes the safest option is otherwise. so maybe that's a task for later :)

42wim:
I usually rephrase like if you want usertyping to be true by default to DisableUserTyping.
But in this case it should be false

bridge/slack/handlers.go
42wim: I think you should also exclude the USER_TYPING event there from getting logged as it's too noisy (like you did on the other places), so it's not needed I think

42wim:
Looks good, but needs a rebase

patcon:
Oops. Missed this verbose one: +1 if you'd like me to silence it as well

https://github.com/42wim/matterbridge/blob/0d64cd8babcec1a54817325adfcc66b4453b3ed1/bridge/slack/handlers.go#L40

patcon:
TODO for me: Update docs to suggest a better bot name: "Someone (over chat bridge)"

This will make the username less confusing when the bot name is invariably mentioned elsewhere (like file uploads)

Issue #560, opened 2018-11-07
Improve the setting of config defaults


patcon:
Reticketed from https://github.com/42wim/matterbridge/pull/559#issuecomment-436509967

Improve the setting of defaults

Right now, was can only easily default to false and empty strings, etc. Sometimes a false or empty string value might not be the most user-friendly default. And altering with the semantics of the key (EnableThing vs DisableThing) to make false _be_ safe is perhaps not ideal.

Here are some settings that could be migrated to a clearer default-setting tactic:
- [`MediaDownloadSize`](https://github.com/42wim/matterbridge/blob/5132d8f097a7eef2cb3d43c1c4cd297bc8254df5/bridge/config/config.go#L206)
- [`Buffer`](https://github.com/42wim/matterbridge/blob/5132d8f097a7eef2cb3d43c1c4cd297bc8254df5/bridge/api/api.go#L37)

KrzysztofMadejski:
+1 

I'd improve the method that is getting the config values to provide defaults. For now we have:

https://github.com/42wim/matterbridge/blob/fb713ed91bfb48c0037e489f80be85c54e69953a/bridge/bridge.go#L99-L105

I'd like to see:
```go
func (b *Bridge) GetStringOrDefault(key string, key string) string {
	val, ok := b.Config.GetString(b.Account + "." + key)
	if !ok {
		val, ok = b.Config.GetString("general." + key)
	}
        if !ok {
                val = default
        }
	return val
}
```

Let me know what you think. It would be 15mins to add such extra methods.

KrzysztofMadejski:
@patcon Now I see that your PR was merged and you addressed something different. Should I create another issue or keep these helper functions here? 
@42wim I wonder what do you think about it.

KrzysztofMadejski:
Also, now I see that I'd like to set default config value for a given bridge (programatically in the bridge). For example, for WhatsApp I'd like to set:
- `PreserveThreading=true` altough the default is false
- `RemoteNickFormat="@{NICK}: "`

42wim:
I'm not for setting hardcoded defaults in the bridges. 

I'd like to all bridges behave the same with their default settings, which are empty strings and false boolean variables.

This gives a consistent experience for the end-user. (and also makes the troubleshooting easier because stuff needs to be enabled explicitly and will be visible in the configuration)

PR #561, opened 2018-11-07
Clean up config loading

patcon:
Started doing this when I was planning to add code for loading remote urls, and
that's no longer needed, but thought I'd still push this upstream, as I think
it's a slight improvement in legibility and DRY?

Also, I've confirmed that even on current master, the default that's trying to be set here `General.MediaDownloadSize` doesn't take affect -- it stays at zero (so if you notice that, it's not an issue with this PR)

Related: https://github.com/42wim/matterbridge/issues/560

### DIFF
--- a/bridge/config/config.go
+++ b/bridge/config/config.go
@@ -2,7 +2,7 @@ package config
 
 import (
 	"bytes"
-	"os"
+	"io/ioutil"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


patcon:
Also, seems we're doing logging in two different ways: directly with `log` and also `flog`. If that's not purposeful, mind if I tidy that too?

42wim:
Sure, you can move everything to `flog`
Will look into the `General.MediaDownloadSize` later.

Issue #244, opened 2017-08-21
Web interface for easy manage matterbridge config file via GUI


MurzNN:
Will be good to create web interface, via that users can manage matterbridge config file via web browser - add new protocols, set passwords and tokens, configure and disable-enable gateways, restart service for apply configuration, etc.
For example, I install matterbridge on server and want give access for add gateways to other company employees, without giving them access to server and ability to broke config file, because they can create rooms and want bridge them, but they are not improved users to do this in file.

jheiselman:
Should this be a separate project that interacts with the matterbridge API? That would require that the API be updated to be able to handle the process's own configuration file and allow for reconfiguring on-the-fly.

It could always start with just a third-party system that updates and then restarts the matterbridge process. Again, would be nice for the API to be able to handle that or for the process to do the "right thing" when it receives a HUP signal.

MurzNN:
This may be simple CLI script, web interface or GUI app that read `matterbridge` config file, accept user changes, update config file and restart matterbridge service.
So we need to parse `toml` file, convert it to array, update data and write back to `toml` file.

42wim:
It would be cool if it could be added to matterbridge itself instead of a separate project.
https://github.com/rakyll/statik could be used to add the files to the binary.

Reconfiguration on the fly would be nice, but not so easy to implement, maybe first just do with a restart.

I'm open to any API changes you need.

patcon:
Not sure whether this is on the critical path, but I've got a simple app that just renders some aspects of the toml config visually:
https://github.com/patcon/matterbridge-heroku-viewer

Contributing functionality to that read-only app might be a step in the right direction, especially if the interface is designed with read-write ability in mind. Then, when we have a feature-set for write-access, we can just take the UI design over to that :)

PR #565, opened 2018-11-07
Allow simple remote configuration files

patcon:
Another run at \#525

Curious about feedback to approach, but still needs work:
- [ ] Debug output says config is reloaded, but channels are not updated. Any advice?

Notes:

- Using PUT instead of POST, since this is replacing an existing resource in full (the config file)
- no public viper methods to get config file location, so had to store it
- not sure best place to store `ConfigURL`, as it's global and never bridge-specific.
- not sure best place to doc in sample config file. Seems like General, but then people will be reading API when they need to know about it, so put it in the latter place :)

If it's alright with you Wim, I\'ll do the rest of the docs inline once Swagger
PR is merged? 

Related: #244

### DIFF
--- a/bridge/api/api.go
+++ b/bridge/api/api.go
@@ -100,11 +100,11 @@ func (b *Api) handleConfigReload(c echo.Context) error {
 		return c.String(http.StatusInternalServerError, "Internal Server Error")
 	}
 	res, err := http.Get(cfgURL)
-	defer res.Body.Close()
 	if err != nil {
 		b.Log.Error("Failed to fetch remote config file: ", err)
 		return c.String(http.StatusInternalServerError, "Internal Server Error")
 	}
+	defer res.Body.Close()
 	content, err := ioutil.ReadAll(res.Body)
 	if err != nil {
 		b.Log.Error("Error reading remote config file: ", err)
--- a/bridge/api/api.go
+++ b/bridge/api/api.go
@@ -2,7 +2,9 @@ package api
 
 import (
 	"encoding/json"
+	"io/ioutil"
 	"net/http"
+	"net/url"
 	"sync"
 	"time"
 
@@ -41,7 +43,9 @@ func New(cfg *bridge.Config) bridge.Bridger {
 			return key == b.GetString("Token"), nil
 		}))
 	}
+
 	e.GET("/api/health", b.handleHealthcheck)
+	e.PUT("/api/reload", b.handleConfigReload)
 	e.GET("/api/messages", b.handleMessages)
 	e.GET("/api/stream", b.handleStream)
 	e.POST("/api/message", b.handlePostMessage)
@@ -80,6 +84,40 @@ func (b *Api) Send(msg config.Message) (string, error) {
 
 func (b *Api) handleHealthcheck(c echo.Context) error {
 	return c.String(http.StatusOK, "OK")
+}
+
+func (b *Api) handleConfigReload(c echo.Context) error {
+	cfgURL := b.GetString("ConfigURL")
+	if cfgURL == "" {
+		b.Log.Warning("Reload API triggered, but no config file url set.")
+		return c.String(http.StatusInternalServerError, "Internal Server Error")
+	}
+
+	b.Log.Debugf("Reloading config from remote file: " + cfgURL)
+	_, err := url.ParseRequestURI(cfgURL)
+	if err != nil {
+		b.Log.Error("Malformed config file url: ", err)
+		return c.String(http.StatusInternalServerError, "Internal Server Error")
+	}
+	res, err := http.Get(cfgURL)
+	defer res.Body.Close()
+	if err != nil {
+		b.Log.Error("Failed to fetch remote config file: ", err)
+		return c.String(http.StatusInternalServerError, "Internal Server Error")
+	}
+	content, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		b.Log.Error("Error reading remote config file: ", err)
+		return c.String(http.StatusInternalServerError, "Internal Server Error")
+	}
+	cfgfile := b.GetConfigFile()
+	err = ioutil.WriteFile(cfgfile, content, 0644)
+	if err != nil {
+		b.Log.Error("Failed to write remote config file: ", err)
+		return c.String(http.StatusInternalServerError, "Internal Server Error")
+	}
+
+	return c.String(http.StatusAccepted, "Accepted")
 }
 
 func (b *Api) handlePostMessage(c echo.Context) error {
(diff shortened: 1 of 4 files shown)


patcon:
Hm. Something about the `config.EVENT_REJOIN_CHANNELS`, but still not clear if every bridge knows how to use that... Sending it from API resulted in a message from "system" in Slack that said "rejoin".

Will investigate more later

patcon:
Also, with the "watching" approach, I don't think it's quick enough in the reload to know when to reload the channels, but we can prob just add in a large wait before sending the reconnect event through the gateways. At least that's my theorizing until I have time to try :)

patcon:
Hm. Or maybe the right approach is to give the router a way to shut down, and just create a new Router object from scratch? I'll admit that I can't figure out how to get everything to reload -- like the viper config is reloading, but it seems there's lots of state in the router/gateways that I don't know how to refresh

Any advice is appreciated!

patcon:
Ok, I'm certain this is wrong, but I'm just pushing that last commit to start a conversation -- I really don't feel like I know a whole lot about how to do this elegantly...!

42wim:
For now don't add any extra features to this PR, only reloading via remote URLs.
If you want to add extra stuff to reload beside the `RELOADABLE settings` open another PR.
All the extra stuff needs to work for *all* bridges and not only slack.

patcon:
> only the reloadable settings are reloaded

So to confirm my understanding, updating of channels bridged to gateways (adding and removing them) is out of scope for this PR?

42wim:
> So to confirm my understanding, updating of channels bridged to gateways (adding and removing them) is out of scope for this PR?

Yes, that's a bigger change with a lot of impact, I'd like to keep the PR's small

bridge/config/config.go
```diff
@@ -229,6 +234,10 @@ func NewConfigFromString(input []byte) *Config {
 	return mycfg
 }
 
+func (c *Config) GetConfigFile() string {
```
42wim: Use https://godoc.org/github.com/spf13/viper#ConfigFileUsed

bridge/config/config.go
```diff
@@ -196,6 +198,9 @@ func NewConfig(cfgfile string) *Config {
 	viper.OnConfigChange(func(e fsnotify.Event) {
 		flog.Println("Config file changed:", e.Name)
 	})
+
+	mycfg.v.Set("ConfigFile", cfgfile)
```
42wim: Not needed, viper already knows which config file

bridge/config/config.go
```diff
@@ -172,6 +173,7 @@ type ConfigValues struct {
 	General            Protocol
 	Gateway            []Gateway
 	SameChannelGateway []SameChannelGateway
+	ConfigFile         string
```
42wim: This is not needed, matterbridge needs to run with `matterbridge -conf /tmp/matterbridge.toml` for example.
The configfile gets set on commandline.

bridge/bridge.go
```diff
@@ -68,6 +68,10 @@ func (b *Bridge) joinChannels(channels map[string]config.ChannelInfo, exists map
 	return nil
 }
 
+func (b *Bridge) GetConfigFile() string {
```
42wim: Not needed

bridge/api/api.go
```diff
@@ -82,6 +86,40 @@ func (b *Api) handleHealthcheck(c echo.Context) error {
...
+		b.Log.Error("Failed to fetch remote config file: ", err)
+		return c.String(http.StatusInternalServerError, "Internal Server Error")
+	}
+	defer res.Body.Close()
+	content, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		b.Log.Error("Error reading remote config file: ", err)
+		return c.String(http.StatusInternalServerError, "Internal Server Error")
+	}
+	cfgfile := b.GetConfigFile()
```
42wim: Use https://godoc.org/github.com/spf13/viper#ConfigFileUsed

patcon:
ok right on. this one's ready for review. Tested it with some reloadable settings and works fine

42wim:
I'm wondering, I think it's better to remove the `ConfigURL` and specify the reload url via the `/api/reload` and a JSON post like below.

`curl --header "Content-Type: application/json" --request POST --data '{"url":"http://youserver/matterbridge.toml"}' http://host/api/reload`

bridge/api/api.go
patcon: Ugh. I don't know how I missed that.

patcon:
I'm down for that, but if we do that, I think we should enforce a Token for the API bridge and fail out if it's not set. Would that be fair?

patcon:
On second thought... I know we're not building anything super-secure or privacy-minded here, but using POST data to set config url feels like unnecessary exposure... Can we allow environment variables too? As in, if someone knows that their config will always be at one spot, there's no need to add that extra level of exposure, if that makes sense :)

bridge/bridge.go
patcon: As in, just load the viper global in bridge/bridge.go?

42wim:
Well, talking about secure or privacy, the URL where you're exposing your configuration on is a greater risk if it's not on https or protected because that contains all your tokens/logins/passwords.

What do you mean by environment variables ?

Another option is that instead of the URL, you just post the whole actual config to the endpoint.

We should also add TLS support to the API (a new PR, not in this one! :), the original use-case was only for localhost and not meant to be available for the whole internet.

patcon:
> URL where you're exposing your configuration on is a greater risk if it's not on https or protected because that contains all your tokens/logins/passwords

Sorry, I don't understand: The tokens and any secrets are stored in environment variables on the matterbridge server (eg. `MATTERBRIDGE_SLACK_MYTEAM_TOKEN`), no one should be allowing their credentials to be fetched this way -- just the data on how the rooms are wired up :)

And good call on TLS!

bridge/config/config.go
42wim: Hm, why marked as resolved?

bridge/bridge.go
```diff
@@ -68,6 +68,10 @@ func (b *Bridge) joinChannels(channels map[string]config.ChannelInfo, exists map
 	return nil
 }
 
+func (b *Bridge) ConfigFileUsed() string {
```
42wim: Why is this function needed ? You have the Config.ConfigFileUsed() function already

42wim:
> Sorry, I don't understand: The tokens and any secrets are stored in environment variables on the matterbridge server (eg. MATTERBRIDGE_SLACK_MYTEAM_TOKEN), no one should be allowing their credentials to be fetched this way -- just the data on how the rooms are wired up :)

Ok that wasn't clear to me, could you please write some examples about this on the wiki too, because this feature will be used wrong by users :)

patcon:
Can do :) And I should test with BasicAuth before we merge.

Also, can we leave the "config file in payload" for anther PR? It's a feature that might be worth putting in, but it doesn't feel necessary to have it this PR, since this is already functional for a set of use-cases

42wim:
I don't think there is going to be any further work on this PR?
If so, please reopen, but closing for now.

Issue #572, opened 2018-11-09
[bug] uploaded files on Slack get sent recursively


patcon:
**Describe the bug**
When uploading a file, the file isn't going into cache, and this seems to be leading to it being sent over and over.

**To Reproduce**
Connect two slack channels (4 channels were wired in env where this surfaced) on same team by a bridge. Send a file.

**Expected behavior**
Would expect a single send.

**Screenshots/debug logs**
![screen shot 2018-11-09 at 11 28 34 am](https://user-images.githubusercontent.com/305339/48241628-a85cd300-e412-11e8-84bf-28fa48cd1608.png)

Logs: https://gist.github.com/patcon/fbf97bc64ed7dd33f7ff49541110f4d4

**Environment (please complete the following information):**
 - OS: darwin
 - Matterbridge version: current HEAD of master

**Additional context**
Please add your configuration file  (be sure to exclude or anonymize private data (tokens/passwords))

cc: @Helcaraxan

patcon:
Actually, I can't reproduce myself. Might have been because i switched branches after matterbridge started running. (Didn't realize it did hot-reload on code?)

patcon:
Turns out this is a real bug once we go back to using web api, so re-opening

New logs that are definitely using the right code (only change was going from `b.rtm` to `b.sc` for `UpdateMessage` and `PostMessage` calls)

https://gist.github.com/patcon/ebe2d279b0551c85f2eb61d3c57ef58b

patcon:
cache.Add returns a bool based on eviction, not success

https://github.com/42wim/matterbridge/blob/0d64cd8babcec1a54817325adfcc66b4453b3ed1/bridge/slack/slack.go#L344-L346

https://godoc.org/github.com/hashicorp/golang-lru#Cache.Add

This resolves the "failed" message, but but there still seems to be something going on that leads to the recursive uploading...

patcon:
Something to do with a skip event that's now missing, i think...

https://github.com/42wim/matterbridge/blob/54adb0509e8f22bdfa70a88834c6499e4078648b/bridge/slack/slack.go#L713-L728

patcon:
@42wim what's the logic of the above? I know it's related to:

https://github.com/42wim/matterbridge/blob/54adb0509e8f22bdfa70a88834c6499e4078648b/bridge/slack/slack.go#L468-L470

[...]

PR #575, opened 2018-11-09
[Slack] Bugfix for file sharings (recursive upload spam)

patcon:
Resolves: https://github.com/42wim/matterbridge/issues/572

Ready for review

cc: @Helcaraxan

### DIFF
--- a/bridge/slack/handlers.go
+++ b/bridge/slack/handlers.go
@@ -132,6 +132,15 @@ func (b *Bslack) skipMessageEvent(ev *slack.MessageEvent) bool {
 		ev.SubMessage.Edited == nil {
 		return true
 	}
+
+	for _, f := range ev.Files {
+		// If the file is in the cache and isn't older then a minute, skip it
+		if b.fileIsRecentlyCached(&f) {
+			b.Log.Debugf("Not downloading file id %s which we uploaded", f.ID)
+			return true
+		}
+	}
+
 	return false
 }
 
@@ -248,10 +257,6 @@ func (b *Bslack) handleTypingEvent(ev *slack.UserTypingEvent) (*config.Message, 
 
 // handleDownloadFile handles file download
 func (b *Bslack) handleDownloadFile(rmsg *config.Message, file *slack.File) error {
-	if b.fileIsAvailable(file) {
-		return nil
-	}
-
 	// Check that the file is neither too large nor blacklisted.
 	if err := helper.HandleDownloadSize(b.Log, rmsg, file.Name, int64(file.Size), b.General); err != nil {
 		b.Log.WithError(err).Infof("Skipping download of incoming file.")
@@ -273,11 +278,10 @@ func (b *Bslack) handleDownloadFile(rmsg *config.Message, file *slack.File) erro
 	return nil
 }
 
-func (b *Bslack) fileIsAvailable(file *slack.File) bool {
-	// Only download a file if it is not in the cache or if it has been entered more than a minute ago.
-	if ts, ok := b.cache.Get("file" + file.ID); ok && time.Since(ts.(time.Time)) > time.Minute {
+func (b *Bslack) fileIsRecentlyCached(file *slack.File) bool {
+	if ts, ok := b.cache.Get("file" + file.ID); ok && time.Since(ts.(time.Time)) < time.Minute {
 		return true
-	} else if ts, ok = b.cache.Get("filename" + file.Name); ok && time.Since(ts.(time.Time)) > 10*time.Second {
+	} else if ts, ok = b.cache.Get("filename" + file.Name); ok && time.Since(ts.(time.Time)) < 10*time.Second {
 		return true
 	}
 	return false
--- a/bridge/slack/handlers.go
+++ b/bridge/slack/handlers.go
@@ -137,13 +137,13 @@ func (b *Bslack) skipMessageEvent(ev *slack.MessageEvent) bool {
 }
 
 func (b *Bslack) filesCached(files []slack.File) bool {
-    for _, f := range files {
-        f := f
-        if !b.fileCached(&f) {
-             return false
-        }
-    }
-    return true
+	for _, f := range files {
+		f := f
+		if !b.fileCached(&f) {
+			return false
+		}
+	}
+	return true
 }
 
 // handleMessageEvent handles the message events. Together with any called sub-methods,
--- a/bridge/slack/handlers.go
+++ b/bridge/slack/handlers.go
@@ -133,15 +133,17 @@ func (b *Bslack) skipMessageEvent(ev *slack.MessageEvent) bool {
 		return true
 	}
 
-	for _, f := range ev.Files {
-		// If the file is in the cache and isn't older then a minute, skip it
-		if b.fileWasRecentlyCached(&f) {
-			b.Log.Debugf("Not downloading file id %s which we uploaded", f.ID)
-			return true
-		}
-	}
+	return b.filesCached(ev.Files)
+}
 
-	return false
+func (b *Bslack) filesCached(files []slack.File) bool {
+    for _, f := range files {
+        f := f
+        if !b.fileCached(&f) {
+             return false
+        }
+    }
+    return true
 }
 
 // handleMessageEvent handles the message events. Together with any called sub-methods,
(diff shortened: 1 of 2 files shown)


bridge/slack/handlers.go
```diff
@@ -293,3 +302,12 @@ func (b *Bslack) fileIsAvailable(file *slack.File) bool {
 	}
 	return false
 }
+
+func (b *Bslack) fileIsRecentlyCached(file *slack.File) bool {
```
Helcaraxan: @patcon the actual fix would be to fix the `fileIsAvailable` method just above that (for whatever reason) has it's comparisons inverted.

patcon: Thanks. I was treading lightly on that as I wasn't clear if it was serving another purpose in handleDownloadFile. But I just confirmed that it was causing another bug (couldn't upload another file of the same name until it expired from cache), and so I just removed it from the download func.

Let me know if you think that's incorrect! Thanks for bearing with me, @Helcaraxan :)

patcon:
Also, the cache.Add issue was also failing each time, because `ok` is not actually ok, but an `evicted` value, which is always false until the cache fills up

patcon:
cache return aside:

Sorry, i was mistaken about flipping operators. Flipping the operators would work too, but would you agree that keeping this in the event skip logic makes more sense? As in, it's not that we're checking for the file being available, and ditching if it's not, but rather that we're ignoring recent ones to avoid flooding back and forth forever (since uploads can't have the callback_id) :)

patcon:
Ugh. I need to get off my computer and sleep more than 4 hours. I have been way too much wrong today. I apologize for the noise, @Helcaraxan (was reading the wrong godoc func 😖 )

bridge/slack/handlers.go
```diff
@@ -133,6 +133,15 @@ func (b *Bslack) skipMessageEvent(ev *slack.MessageEvent) bool {
 		ev.SubMessage.Edited == nil {
 		return true
 	}
+
+	for _, f := range ev.Files {
+		// If the file is in the cache and isn't older then a minute, skip it
+		if b.fileIsRecentlyCached(&f) {
+			b.Log.Debugf("Not downloading file id %s which we uploaded", f.ID)
+			return true
```
Helcaraxan: It would be wrong to skip here. We should only skip (and thus return `true`) if **all** the files have been recently cached.

bridge/slack/handlers.go
```diff
@@ -284,11 +289,10 @@ func (b *Bslack) handleDownloadFile(rmsg *config.Message, file *slack.File) erro
 	return nil
 }
 
-func (b *Bslack) fileIsAvailable(file *slack.File) bool {
-	// Only download a file if it is not in the cache or if it has been entered more than a minute ago.
-	if ts, ok := b.cache.Get("file" + file.ID); ok && time.Since(ts.(time.Time)) > time.Minute {
+func (b *Bslack) fileIsRecentlyCached(file *slack.File) bool {
```
Helcaraxan: Nit: to make this properly named it should be `fileWasRecentlyCached()`.

bridge/slack/handlers.go
```diff
@@ -259,10 +268,6 @@ func (b *Bslack) handleTypingEvent(ev *slack.UserTypingEvent) (*config.Message,
 
 // handleDownloadFile handles file download
 func (b *Bslack) handleDownloadFile(rmsg *config.Message, file *slack.File) error {
-	if b.fileIsAvailable(file) {
```
Helcaraxan: We can't remove this check. If we get here there might be some files that are cached and others are not (if you follow my comment on your addition to the `skipMessageEvent()` method). So we need to filter out (again) the ones that are available in the cache.

bridge/slack/handlers.go
patcon: but isn't this the filter for helping the bridge recognize it's own doing? it's not meant to be the place where the "make sure everything is cached" part of the process happens, is it? Or am I misunderstanding and the "was this us" logic for file_shares happens somewhere else?

bridge/slack/handlers.go
patcon: haha depends if one reads with an implied hyphen or space, but fair enough :)

patcon: Actually, I'm unhappy with my stated rationale, but this captures an explanation for my underlying sensibilities :) 

https://stackoverflow.com/questions/3669433/do-you-use-tense-when-naming-methods-of-boolean-return-type

Changed already , but just food for thought

bridge/slack/handlers.go
Helcaraxan: @patcon this comment still needs addressing. 🙂 

My suggestion would be to keep it as is but modifying the call to correspond to `b.fileWasRecentlyCached()`.

bridge/slack/handlers.go
Helcaraxan: The code I would suggest instead:
```go
	var filesNotSkipped bool
	for _, f := range ev.Files {
		if !b.fileWasRecentlyCached(&f) {
			filesNotSkipped = true
		}
	}
	return !filesNotSkipped
```

bridge/slack/handlers.go
```diff
@@ -284,11 +289,10 @@ func (b *Bslack) handleDownloadFile(rmsg *config.Message, file *slack.File) erro
 	return nil
 }
 
-func (b *Bslack) fileIsAvailable(file *slack.File) bool {
-	// Only download a file if it is not in the cache or if it has been entered more than a minute ago.
-	if ts, ok := b.cache.Get("file" + file.ID); ok && time.Since(ts.(time.Time)) > time.Minute {
+func (b *Bslack) fileWasRecentlyCached(file *slack.File) bool {
```
Helcaraxan: I would add a godoc comment here with the method to keep a trace of the rationale behind this logic.
Suggestion:
```go
// fileWasRecentlyCached implements Matterbridge's caching logic for files
// shared via Slack.
// 
// We consider that a file was cached if its ID was added in the last minute or
// it's name was registered in the last 10 seconds. This ensures that an
// identically named file but with different content will be uploaded correctly
// (the assumption is that such name collisions will not occur within the given
// timeframes).
```

bridge/slack/handlers.go
42wim: I don't find this very readable with all the inverted ifs and naming (lots of mental overhead) 
What if we make a separate function like below (maybe also rename `fileWasRecentlyCached` to `fileCached` maybe?)
```golang
func (b *Bslack) filesCached(files []slack.File) bool {
    for _, f:= range files {
        f := f
        if !b.fileWasRecentlyCached(&f) {
             return false
        }
   }
   return true
}
```
and just call
```golang
return filesCached(ev.Files)
```

Helcaraxan: Sure! I like your suggestion. Indeeds less mental overhead. 👍

patcon:
Ok, this should have all the requested changes :)

42wim:
LGTM 
Can you rebase so it can be merged ? (remove the mattermost stuff, that was my fault which I fixed :)

bridge/slack/handlers.go
```diff
@@ -132,7 +132,18 @@ func (b *Bslack) skipMessageEvent(ev *slack.MessageEvent) bool {
...
 		return true
 	}
-	return false
+
+	return b.filesCached(ev.Files)
+}
+
+func (b *Bslack) filesCached(files []slack.File) bool {
+	for _, f := range files {
+		f := f
```
Helcaraxan: I'd rather that this be:
```go
for i := range files {
	if !b.fileCached(&files[i]) {
		return false
	}
}
```
Copying is not useful here and it's also introducing a shadow variable.

bridge/slack/handlers.go
```diff
@@ -132,7 +132,18 @@ func (b *Bslack) skipMessageEvent(ev *slack.MessageEvent) bool {
 		ev.SubMessage.Edited == nil {
 		return true
 	}
-	return false
+
```
Helcaraxan: Nit: superfluous empty line.

bridge/slack/handlers.go
42wim: It's based on the suggestions on https://github.com/kyoh86/scopelint
But don't mind one way or the other.

patcon:
ok thanks y'all! good to go

42wim:
Thanks! 👍

bridge/slack/handlers.go
patcon: I should caught this, but `return filesCached(ev.Files)` makes every message is skipped. Sorry, I went into no-think mode, without fully testing this last bit :)

I'm thinking this is how we should've ended this func:

```go
if len(ev.Files) > 0 {
    return b.filesCached(ev.Files)
}
return false
```

PR #582, opened 2018-11-10
#510 New Slack connection and forked legacy Slack bridge

Helcaraxan:
Final commit for #510. This moves the existing Slack connection logic to a new `slack-legacy` bridge to maintain the possibility for users to continue to use the former logic as a conscious choice.

The new connection logic requires the use of a Slack bot-token which can be retrieved by installing a SlackApp to the targeted workspace (wiki documentation will be updated at merge of this PR). If a `Token` is specified in the Slack bridge's configuration only that will be used for exchanging messages, disregarding any specification of a `WebhookURL` or `WebhookBindAddress`. This does not preclude the bot messages from being sent with a modified nickname (see screenshot below):

screenshot 2018-11-10 at 18 42 25

### DIFF
--- a/bridge/slack/slack.go
+++ b/bridge/slack/slack.go
@@ -73,6 +73,20 @@ const (
 )
 
 func New(cfg *bridge.Config) bridge.Bridger {
+	// Print a deprecation warning for legacy non-bot tokens (#527).
+	token := cfg.GetString(tokenConfig)
+	if token != "" && !strings.HasPrefix(token, "xoxb") {
+		cfg.Log.Error("Non-bot token detected. It is STRONGLY recommended to use a proper bot-token instead.")
+		cfg.Log.Error("Legacy tokens may be deprecated by Slack at short notice. See the Matterbridge GitHub wiki for a migration guide.")
+		cfg.Log.Error("To continue using a legacy token please move your configuration to a \"slack-legacy\" bridge instead.")
+		cfg.Log.Error("Delaying start of bridge by 30 seconds. Future Matterbridge release will fail here unless you use a \"slack-legacy\" bridge.")
+		time.Sleep(30 * time.Second)
+		return NewLegacy(cfg)
+	}
+	return newBridge(cfg)
+}
+
+func newBridge(cfg *bridge.Config) *Bslack {
 	newCache, err := lru.New(5000)
 	if err != nil {
 		cfg.Log.Fatalf("Could not create LRU cache for Slack bridge: %v", err)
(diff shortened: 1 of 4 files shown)


bridge/slack/slack.go
```diff
@@ -73,6 +73,10 @@ const (
 )
 
 func New(cfg *bridge.Config) bridge.Bridger {
+	return newBridge(cfg)
```
42wim: Maybe we can do some leeway, add deprecation now (matterbridge 1.12.0) and remove it in 1.13.0

```golang
b := &Bslack{Config: cfg}
if !strings.HasPrefix(b.GetString(tokenConfig), "xoxb") {
      b.Log.Error("Non-bot token detected. It is STRONGLY recommended to use a proper bot-token instead.")
      b.Log.Error("Legacy tokens may be deprecated by slack at short notice. See the Matterbridge GitHub wiki for a migration guide.")
      b.Log.Error("To continue using a legacy token please move your configuration to a \"slack-legacy\" bridge instead.")
      b.Log.Error("Delaying for 30 seconds. The next release of matterbridge will exit here.")
      time.Sleep(time.Second * 30)
      return NewLegacy(cfg)
}
return newBridge(cfg)
```

bridge/slack/slack.go
```diff
@@ -97,56 +101,46 @@ func (b *Bslack) Command(cmd string) string {
...
-			b.rtm = b.sc.NewRTM()
-			go b.rtm.ManageConnection()
-			go b.handleSlack()
+
+	// If we have a token we use the Slack websocket-based RTM for both sending and receiving.
+	if token := b.GetString(tokenConfig); token != "" {
+		// Print a warning for legacy non-bot tokens (#527).
+		if !strings.HasPrefix(token, "xoxb") {
+			b.Log.Error("Non-bot token detected. It is STRONGLY recommended to use a proper bot-token instead.")
+			b.Log.Error("Legacy tokens may be deprecated at short notice. See the Matterbridge GitHub wiki for a migration guide.")
```
42wim: ```... be deprecated by slack ...``` so that we're not the one to blame ;-)

bridge/slack/slack.go
42wim: No, the delay should stay. Otherwise users won't see this message and the next version of matterbridge will break their config.

If you don't want the delay and any message, do the migration or move to `slack-legacy` as the message says:  `To continue using a legacy token please move your configuration to a "slack-legacy" bridge instead."` 

You can just edit your config-file to use "slack-legacy" instead of slack.

Helcaraxan: Fair enough. Thank you for the code suggestion @42wim. Have taken it with a slight rewrite of the last "delay" line.

PR #583, opened 2018-11-10
Restore file comments coming from Slack

Helcaraxan:
I noticed during some testing that when I posted a file to Slack with a comment, the comment did not appear on the Discord channel. Parsing through the Matterbridge logs it appears that Slack's way of showing comments has somewhat changed and we thus need to adapt the Slack bridge's logic.

Also using the PR for some minor stylistic fixes as well as adding checks on previously unchecked type-casts.

**Before this PR**
screenshot 2018-11-10 at 23 03 37
screenshot 2018-11-10 at 23 03 45

**After this PR**
screenshot 2018-11-10 at 23 03 51
screenshot 2018-11-10 at 23 03 57

### DIFF
--- a/bridge/slack/handlers.go
+++ b/bridge/slack/handlers.go
@@ -3,7 +3,6 @@ package bslack
 import (
 	"fmt"
 	"html"
-	"regexp"
 	"time"
 
 	"github.com/42wim/matterbridge/bridge/config"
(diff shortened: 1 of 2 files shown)


bridge/slack/handlers.go
```diff
@@ -228,33 +227,23 @@ func (b *Bslack) handleAttachments(ev *slack.MessageEvent, rmsg *config.Message)
...
-		err := b.handleDownloadFile(rmsg, &f)
-		if err != nil {
+	for i := range ev.Files {
+		if err := b.handleDownloadFile(rmsg, &ev.Files[i]); err != nil {
 			b.Log.Errorf("Could not download incoming file: %#v", err)
 		}
 	}
 }
 
-var commentRE = regexp.MustCompile(`.*?commented: (.*)`)
```
Helcaraxan: The regexp is no longer needed as an uploaded file with comment will be a simple Slack message with the file(s) attached in the `Files` field of the message and the comment in the `Text` field.

Issue #353, opened 2018-01-30
Don't send purpose and topic messages through Slack bridge


patcon:
I have a use-case that I expect won't be that uncommon as this tool catches on within communities.

I have created a gateway for connecting the leadership of all civic tech communities in Canada. There are going to be maybe 5 Slack teams in the channel for now.

We try to keep the topic and purpose synchronized between Slack teams. This is where we explain the how the bridge works, and who is present on the other sides.

Right now, when we update the purpose or topic on each slack team, we essentially spam the whole gateway. It would be much more sensible to NOT send these statements through the portal by default, and perhaps allow an option to turn this on (if that's even necessary).

https://imgur.com/WLn7LOb

Thanks! Curious your thoughts!

42wim:
Sounds reasonable (I didn't even know those messages were sent).
I can make a option for this.

![image](https://user-images.githubusercontent.com/1810977/35635869-b77f638e-06af-11e8-95bb-3ea4f1a76d4a.png)

42wim:
So, there is now a `ShowTopicChange` option.
Disabled by default

patcon:
Thanks 42wim! Sorry, I've been lazy lately, and you are doing all the work. I really appreciate it though

Before this gets locked into a release, perhaps a small change I'd be curious your thoughts on:

Seems there are lots of channel action messages that would all be undesirable at the same time:
https://api.slack.com/events/message

Specifically:
```
channel_purpose
channel_name
channel_topic
pinned_item
unpinned_item
```

Would it make sense for the setting to be `ShowMetaMessages` or `ShowChannelMeta`, and enable/disable them all at once?

patcon:
Though actually, "channel_name" would be useful, as that seemingly-innocent change will break the bridge on the next restart :) I'd at least be aware it was about to go down if I same the name-change message...!

But all the others are unhelpful noise

42wim:
oh right, i'll add channel_purpose to the channel_topic switch.
The pinned_items don't deserve a toggle, I'll block those.

The channel_name one are very rare I think? So those can go through.

patcon:
Yep, very rare normally, and should be "never" here :)

You're great!

PR #585, opened 2018-11-13
Sync channel topics over gateway (slack)

patcon:
Keeping topic messages sync'd between Slacks can be burdensome. But without syncing, having different information on each channel can be confusing.

When `SyncChannelTopics` is set to true, updating a topic in one channel would update it in others across gateway, when possible. MVP would just be across Slack teams.

A known pattern would prevent the topic from being set. For example, ending the topic with `[lock]` might prevent it from being modified by changes in other bridges (and reciprocally, updates with that pattern wouldn't be propagated).

Future extensions of feature:
- Apply same logic for channel purpose with `SyncChannelPurposes`.
- Extend feature to work for non-Slack protocols.

- `ShowTopicChange` covers some of this: https://github.com/42wim/matterbridge/issues/353#issuecomment-362693052

**Additional context**
None.

screen shot 2018-11-13 at 5 50 25 pm

### DIFF
--- a/bridge/config/config.go
+++ b/bridge/config/config.go
@@ -117,7 +117,7 @@ type Protocol struct {
 	ShowEmbeds             bool       // discord
 	SkipTLSVerify          bool       // IRC, mattermost
 	StripNick              bool       // all protocols
-	SyncTopicChange        bool       // slack
+	SyncTopic              bool       // slack
 	Team                   string     // mattermost
 	Token                  string     // gitter, slack, discord, api
 	Topic                  string     // zulip
--- a/bridge/config/config.go
+++ b/bridge/config/config.go
@@ -117,6 +117,7 @@ type Protocol struct {
 	ShowEmbeds             bool       // discord
 	SkipTLSVerify          bool       // IRC, mattermost
 	StripNick              bool       // all protocols
+	SyncTopicChange        bool       // slack
 	Team                   string     // mattermost
 	Token                  string     // gitter, slack, discord, api
 	Topic                  string     // zulip
--- a/bridge/slack/handlers.go
+++ b/bridge/slack/handlers.go
@@ -141,7 +141,6 @@ func (b *Bslack) skipMessageEvent(ev *slack.MessageEvent) bool {
 	if len(ev.Files) > 0 {
 		return b.filesCached(ev.Files)
 	}
-
 	return false
 }
 
--- a/bridge/slack/handlers.go
+++ b/bridge/slack/handlers.go
@@ -195,7 +195,7 @@ func (b *Bslack) handleMessageEvent(ev *slack.MessageEvent) (*config.Message, er
 	return rmsg, nil
 }
 
-func (b *Bslack) handleStatusEvent(ev *slack.MessageEvent, rmsg *config.Message) (doneProcessing bool) {
+func (b *Bslack) handleStatusEvent(ev *slack.MessageEvent, rmsg *config.Message) bool {
 	switch ev.SubType {
 	case sChannelJoined, sMemberJoined:
 		b.populateUsers()
--- a/bridge/slack/handlers.go
+++ b/bridge/slack/handlers.go
@@ -190,7 +190,7 @@ func (b *Bslack) handleMessageEvent(ev *slack.MessageEvent) (*config.Message, er
 	return rmsg, nil
 }
 
-func (b *Bslack) handleStatusEvent(ev *slack.MessageEvent, rmsg *config.Message) bool {
+func (b *Bslack) handleStatusEvent(ev *slack.MessageEvent, rmsg *config.Message) (doneProcessing bool) {
 	switch ev.SubType {
 	case sChannelJoined, sMemberJoined:
 		b.populateUsers()
--- a/bridge/slack/handlers.go
+++ b/bridge/slack/handlers.go
@@ -116,6 +116,11 @@ func (b *Bslack) skipMessageEvent(ev *slack.MessageEvent) bool {
 		return b.GetBool(noSendJoinConfig)
 	case sPinnedItem, sUnpinnedItem:
 		return true
+	case sChannelTopic, sChannelPurpose:
+		// Skip the event if our bot/user account changed the topic/purpose
+		if ev.User == b.si.User.ID {
+			return true
+		}
 	}
 
 	// Skip any messages that we made ourselves or from 'slackbot' (see #527).
@@ -201,6 +206,7 @@ func (b *Bslack) handleStatusEvent(ev *slack.MessageEvent, rmsg *config.Message)
 		rmsg.Username = sSystemUser
 		rmsg.Event = config.EventJoinLeave
 	case sChannelTopic, sChannelPurpose:
+		b.populateChannels()
 		rmsg.Event = config.EventTopicChange
 	case sMessageChanged:
 		rmsg.Text = ev.SubMessage.Text
--- a/bridge/slack/helpers.go
+++ b/bridge/slack/helpers.go
@@ -269,7 +269,7 @@ var (
 	topicOrPurposeRE = regexp.MustCompile(`(?s)(@.+) (cleared|set)(?: the)? channel (topic|purpose)(?:: (.*))?`)
 )
 
-func extractTopicOrPurpose(text string) (string, string) {
+func (b *Bslack) extractTopicOrPurpose(text string) (string, string) {
 	r := topicOrPurposeRE.FindStringSubmatch(text)
 	if len(r) == 5 {
 		action, updateType, extracted := r[2], r[3], r[4]
@@ -281,7 +281,7 @@ func extractTopicOrPurpose(text string) (string, string) {
 		}
 	}
 	b.Log.Warnf("Encountered channel topic or purpose change message with unexpected format: %s", text)
-	return unknown, ""
+	return "unknown", ""
 }
 
 // @see https://api.slack.com/docs/message-formatting#linking_to_channels_and_users
--- a/bridge/slack/helpers.go
+++ b/bridge/slack/helpers.go
@@ -271,7 +271,7 @@ var (
 
 func extractTopicOrPurpose(text string) (string, string) {
 	r := topicOrPurposeRE.FindStringSubmatch(text)
-	if len(r) >= 5 {
+	if len(r) == 5 {
 		action, updateType, extracted := r[2], r[3], r[4]
 		switch action {
 		case "set":
@@ -280,7 +280,8 @@ func extractTopicOrPurpose(text string) (string, string) {
 			return updateType, ""
 		}
 	}
-	return "unknown", ""
+	b.Log.Warnf("Encountered channel topic or purpose change message with unexpected format: %s", text)
+	return "unknown, ""
 }
 
 // @see https://api.slack.com/docs/message-formatting#linking_to_channels_and_users
--- a/bridge/slack/helpers.go
+++ b/bridge/slack/helpers.go
@@ -281,7 +281,7 @@ func extractTopicOrPurpose(text string) (string, string) {
 		}
 	}
 	b.Log.Warnf("Encountered channel topic or purpose change message with unexpected format: %s", text)
-	return "unknown, ""
+	return unknown, ""
 }
 
 // @see https://api.slack.com/docs/message-formatting#linking_to_channels_and_users
--- a/bridge/slack/helpers.go
+++ b/bridge/slack/helpers.go
@@ -266,15 +266,12 @@ var (
 	channelRE  = regexp.MustCompile(`<#[a-zA-Z0-9]+\|(.+?)>`)
 	variableRE = regexp.MustCompile(`<!((?:subteam\^)?[a-zA-Z0-9]+)(?:\|@?(.+?))?>`)
 	urlRE      = regexp.MustCompile(`<(.*?)(\|.*?)?>`)
-	topicOrPurposeRE    = regexp.MustCompile(`(?s)^@.+ set the channel (topic|purpose): (.+?)(\[nosync\])?$`)
+	topicOrPurposeRE    = regexp.MustCompile(`(?s)^@.+ set the channel (topic|purpose): (.+?)$`)
 )
 
 func (b *Bslack) extractTopicOrPurpose(text string) (updateType string, extracted string) {
 	r := topicOrPurposeRE.FindStringSubmatch(text)
-	updateType, extracted, noSync := r[1], r[2], r[3]
-	if noSync != "" {
-		return "nosync", ""
-	}
+	updateType, extracted = r[1], r[2]
 	return updateType, extracted
 }
 
--- a/bridge/slack/helpers.go
+++ b/bridge/slack/helpers.go
@@ -266,13 +266,19 @@ var (
 	channelRE        = regexp.MustCompile(`<#[a-zA-Z0-9]+\|(.+?)>`)
 	variableRE       = regexp.MustCompile(`<!((?:subteam\^)?[a-zA-Z0-9]+)(?:\|@?(.+?))?>`)
 	urlRE            = regexp.MustCompile(`<(.*?)(\|.*?)?>`)
-	topicOrPurposeRE = regexp.MustCompile(`(?s)^@.+ set the channel (topic|purpose): (.+?)$`)
+	topicOrPurposeRE = regexp.MustCompile(`(?s)(@.+) (cleared|set)(?: the)? channel (topic|purpose)(?:: (.*))?`)
 )
 
-func (b *Bslack) extractTopicOrPurpose(text string) (updateType string, extracted string) {
+func (b *Bslack) extractTopicOrPurpose(text string) (string, string) {
 	r := topicOrPurposeRE.FindStringSubmatch(text)
-	updateType, extracted = r[1], r[2]
-	return updateType, extracted
+	action, updateType, extracted := r[2], r[3], r[4]
+	switch action {
+	case "set":
+		return updateType, extracted
+	case "cleared":
+		return updateType, ""
+	}
+	return "", ""
 }
 
 // @see https://api.slack.com/docs/message-formatting#linking_to_channels_and_users
--- a/bridge/slack/helpers.go
+++ b/bridge/slack/helpers.go
@@ -262,11 +262,11 @@ func (b *Bslack) populateMessageWithBotInfo(ev *slack.MessageEvent, rmsg *config
 }
 
 var (
-	mentionRE  = regexp.MustCompile(`<@([a-zA-Z0-9]+)>`)
-	channelRE  = regexp.MustCompile(`<#[a-zA-Z0-9]+\|(.+?)>`)
-	variableRE = regexp.MustCompile(`<!((?:subteam\^)?[a-zA-Z0-9]+)(?:\|@?(.+?))?>`)
-	urlRE      = regexp.MustCompile(`<(.*?)(\|.*?)?>`)
-	topicOrPurposeRE    = regexp.MustCompile(`(?s)^@.+ set the channel (topic|purpose): (.+?)$`)
+	mentionRE        = regexp.MustCompile(`<@([a-zA-Z0-9]+)>`)
+	channelRE        = regexp.MustCompile(`<#[a-zA-Z0-9]+\|(.+?)>`)
+	variableRE       = regexp.MustCompile(`<!((?:subteam\^)?[a-zA-Z0-9]+)(?:\|@?(.+?))?>`)
+	urlRE            = regexp.MustCompile(`<(.*?)(\|.*?)?>`)
+	topicOrPurposeRE = regexp.MustCompile(`(?s)^@.+ set the channel (topic|purpose): (.+?)$`)
 )
 
 func (b *Bslack) extractTopicOrPurpose(text string) (updateType string, extracted string) {
--- a/bridge/slack/helpers.go
+++ b/bridge/slack/helpers.go
@@ -266,7 +266,17 @@ var (
 	channelRE  = regexp.MustCompile(`<#[a-zA-Z0-9]+\|(.+?)>`)
 	variableRE = regexp.MustCompile(`<!((?:subteam\^)?[a-zA-Z0-9]+)(?:\|@?(.+?))?>`)
 	urlRE      = regexp.MustCompile(`<(.*?)(\|.*?)?>`)
+	topicOrPurposeRE    = regexp.MustCompile(`(?s)^@.+ set the channel (topic|purpose): (.+?)(\[nosync\])?$`)
 )
+
+func (b *Bslack) extractTopicOrPurpose(text string) (updateType string, extracted string) {
+	r := topicOrPurposeRE.FindStringSubmatch(text)
+	updateType, extracted, noSync := r[1], r[2], r[3]
+	if noSync != "" {
+		return "nosync", ""
+	}
+	return updateType, extracted
+}
 
 // @see https://api.slack.com/docs/message-formatting#linking_to_channels_and_users
 func (b *Bslack) replaceMention(text string) string {
--- a/bridge/slack/helpers.go
+++ b/bridge/slack/helpers.go
@@ -269,16 +269,18 @@ var (
 	topicOrPurposeRE = regexp.MustCompile(`(?s)(@.+) (cleared|set)(?: the)? channel (topic|purpose)(?:: (.*))?`)
 )
 
-func (b *Bslack) extractTopicOrPurpose(text string) (string, string) {
+func extractTopicOrPurpose(text string) (string, string) {
 	r := topicOrPurposeRE.FindStringSubmatch(text)
-	action, updateType, extracted := r[2], r[3], r[4]
-	switch action {
-	case "set":
-		return updateType, extracted
-	case "cleared":
-		return updateType, ""
+	if len(r) >= 5 {
+		action, updateType, extracted := r[2], r[3], r[4]
+		switch action {
+		case "set":
+			return updateType, extracted
+		case "cleared":
+			return updateType, ""
+		}
 	}
-	return "", ""
+	return "unknown", ""
 }
 
 // @see https://api.slack.com/docs/message-formatting#linking_to_channels_and_users
--- a/bridge/slack/helpers_test.go
+++ b/bridge/slack/helpers_test.go
@@ -20,8 +20,9 @@ func TestExtractTopicOrPurpose(t *testing.T) {
 		"error - unhandled":      {"some unmatched message", "unknown", ""},
 	}
 
+	b := &Bslack{}
 	for name, tc := range testcases {
-		gotChangeType, gotOutput := extractTopicOrPurpose(tc.input)
+		gotChangeType, gotOutput := b.extractTopicOrPurpose(tc.input)
 
 		assert.Equalf(t, tc.wantChangeType, gotChangeType, "This testcase failed: %s", name)
 		assert.Equalf(t, tc.wantOutput, gotOutput, "This testcase failed: %s", name)
--- a/bridge/slack/helpers_test.go
+++ b/bridge/slack/helpers_test.go
@@ -1,0 +1,26 @@
+package bslack
+
+import (
+	"testing"
+)
+
+func TestExtractTopicOrPurpose(t *testing.T) {
+	tables := []struct{
+		input string
+		changeType string
+		output string
+	}{
+		{"@someone set channel topic: one liner",    "topic",   "one liner"},
+		{"@someone set channel purpose: one liner",  "purpose", "one liner"},
+		{"@someone set channel topic: multi\nliner", "topic",   "multi\nliner"},
+		{"@someone cleared channel topic",           "topic",   ""},
+		{"some unmatched message",                   "unknown", ""},
+	}
+
+	for _, table := range tables {
+		changeType, output := extractTopicOrPurpose(table.input)
+		if changeType != table.changeType || output != table.output {
+			t.Error()
+		}
+	}
+}
--- a/bridge/slack/helpers_test.go
+++ b/bridge/slack/helpers_test.go
@@ -2,13 +2,15 @@ package bslack
 
 import (
 	"testing"
+
+	"github.com/stretchr/testify/assert"
 )
 
 func TestExtractTopicOrPurpose(t *testing.T) {
-	tables := []struct{
+	testcases := []struct{
 		input string
-		changeType string
-		output string
+		wantChangeType string
+		wantOutput string
 	}{
 		{"@someone set channel topic: one liner",    "topic",   "one liner"},
 		{"@someone set channel purpose: one liner",  "purpose", "one liner"},
@@ -17,10 +19,10 @@ func TestExtractTopicOrPurpose(t *testing.T) {
 		{"some unmatched message",                   "unknown", ""},
 	}
 
-	for _, table := range tables {
-		changeType, output := extractTopicOrPurpose(table.input)
-		if changeType != table.changeType || output != table.output {
-			t.Error()
-		}
+	for _, tc := range testcases {
+		gotChangeType, gotOutput := extractTopicOrPurpose(tc.input)
+
+		assert.Equal(t, tc.wantChangeType , gotChangeType)
+		assert.Equal(t, tc.wantOutput , gotOutput)
 	}
 }
--- a/bridge/slack/helpers_test.go
+++ b/bridge/slack/helpers_test.go
@@ -7,22 +7,23 @@ import (
 )
 
 func TestExtractTopicOrPurpose(t *testing.T) {
-	testcases := []struct {
+	testcases := map[string]struct {
 		input          string
 		wantChangeType string
 		wantOutput     string
 	}{
-		{"@someone set channel topic: one liner", "topic", "one liner"},
-		{"@someone set channel purpose: one liner", "purpose", "one liner"},
-		{"@someone set channel topic: multi\nliner", "topic", "multi\nliner"},
-		{"@someone cleared channel topic", "topic", ""},
-		{"some unmatched message", "unknown", ""},
+		"success - topic type": {"@someone set channel topic: foo bar", "topic", "foo bar"},
+		"success - purpose type": {"@someone set channel purpose: foo bar", "purpose", "foo bar"},
+		"success - one line": {"@someone set channel topic: foo bar", "topic", "foo bar"},
+		"success - multi-line": {"@someone set channel topic: foo\nbar", "topic", "foo\nbar"},
+		"success - cleared": {"@someone cleared channel topic", "topic", ""},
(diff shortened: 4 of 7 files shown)


patcon:
Ok, this is ready for review :)

bridge/config/config.go
```diff
@@ -115,6 +115,7 @@ type Protocol struct {
 	ShowEmbeds             bool       // discord
 	SkipTLSVerify          bool       // IRC, mattermost
 	StripNick              bool       // all protocols
+	SyncTopicChange        bool       // slack
```
patcon: This was kinda discuss before, but should we use something more clear for `SyncTopicChange` and `ShowTopicChange`, since they apply to "purpose" too?

Maybe `SyncChannelMetaChanges` or `ShowChannelMetaChanges`? Kinda long though...

bridge/slack/handlers.go
```diff
@@ -187,6 +192,7 @@ func (b *Bslack) handleStatusEvent(ev *slack.MessageEvent, rmsg *config.Message)
 		rmsg.Username = sSystemUser
 		rmsg.Event = config.EVENT_JOIN_LEAVE
 	case sChannelTopic, sChannelPurpose:
+		b.populateChannels()
```
patcon: This must be up-to-date since we check for "[nosync]" suffix later

bridge/slack/slack.go
```diff
@@ -280,6 +280,24 @@ func (b *Bslack) sendRTM(msg config.Message) (string, error) {
...
+			if strings.HasSuffix(channelInfo.Topic.Value, "[nosync]") {
+				break
+			}
+			b.rtm.SetTopicOfConversation(channelInfo.ID, text)
+		case "purpose":
+			if strings.HasSuffix(channelInfo.Purpose.Value, "[nosync]") {
+				break
+			}
+			b.rtm.SetPurposeOfConversation(channelInfo.ID, text)
+		case "nosync":
```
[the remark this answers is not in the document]
patcon: Actually, working on this now. one sec

bridge/mattermost/mattermost.go
```diff
@@ -23,6 +23,10 @@ type Bmattermost struct {
 	avatarMap map[string]string
 }
 
+const (
+	MATTERMOST_PLUGIN = "mattermost.plugin"
```
Helcaraxan: Please use standard camel-case.

bridge/slack/handlers.go
```diff
@@ -176,7 +181,7 @@ func (b *Bslack) handleMessageEvent(ev *slack.MessageEvent) (*config.Message, er
 	return rmsg, nil
 }
 
-func (b *Bslack) handleStatusEvent(ev *slack.MessageEvent, rmsg *config.Message) bool {
+func (b *Bslack) handleStatusEvent(ev *slack.MessageEvent, rmsg *config.Message) (doneProcessing bool) {
```
Helcaraxan: Why are we introducing a named-return? It does not appear to be set anywhere.

bridge/slack/helpers.go
```diff
@@ -261,12 +261,19 @@ func (b *Bslack) populateMessageWithBotInfo(ev *slack.MessageEvent, rmsg *config
...
 var (
-	mentionRE  = regexp.MustCompile(`<@([a-zA-Z0-9]+)>`)
-	channelRE  = regexp.MustCompile(`<#[a-zA-Z0-9]+\|(.+?)>`)
-	variableRE = regexp.MustCompile(`<!((?:subteam\^)?[a-zA-Z0-9]+)(?:\|@?(.+?))?>`)
-	urlRE      = regexp.MustCompile(`<(.*?)(\|.*?)?>`)
+	mentionRE        = regexp.MustCompile(`<@([a-zA-Z0-9]+)>`)
+	channelRE        = regexp.MustCompile(`<#[a-zA-Z0-9]+\|(.+?)>`)
+	variableRE       = regexp.MustCompile(`<!((?:subteam\^)?[a-zA-Z0-9]+)(?:\|@?(.+?))?>`)
+	urlRE            = regexp.MustCompile(`<(.*?)(\|.*?)?>`)
+	topicOrPurposeRE = regexp.MustCompile(`(?s)^@.+ set the channel (topic|purpose): (.+?)$`)
```
Helcaraxan: I don't really like the magic of the `(?s)` but the alternatives are much less readable... Let's go with it. 😆

bridge/slack/slack.go
```diff
@@ -281,8 +281,14 @@ func (b *Bslack) sendRTM(msg config.Message) (string, error) {
 		return "", nil
 	}
 
-	// Handle message deletions.
 	var handled bool
```
Helcaraxan: Please move this between the comment and the `if` statement. No need to create extra empty lines for nothing.

bridge/slack/slack.go
```diff
@@ -315,6 +321,39 @@ func (b *Bslack) sendRTM(msg config.Message) (string, error) {
 	return b.postMessage(&msg, messageParameters, channelInfo)
 }
 
+func (b *Bslack) updateTopicOrPurpose(msg *config.Message, channelInfo *slack.Channel) (handled bool, err error) {
+	if msg.Event != config.EVENT_TOPIC_CHANGE {
+		return false, nil
+	}
+
+	if !b.GetBool("SyncTopicChange") {
```
Helcaraxan: Please merge the two `if`s into one with an "or".

bridge/slack/helpers.go
```diff
@@ -261,12 +261,19 @@ func (b *Bslack) populateMessageWithBotInfo(ev *slack.MessageEvent, rmsg *config
...
 )
 
+func (b *Bslack) extractTopicOrPurpose(text string) (updateType string, extracted string) {
+	r := topicOrPurposeRE.FindStringSubmatch(text)
+	updateType, extracted = r[1], r[2]
```
Helcaraxan: This will panic if called on a non-matching string. Please check the slice length before referring to its elements.

bridge/slack/slack.go
```diff
@@ -315,6 +321,39 @@ func (b *Bslack) sendRTM(msg config.Message) (string, error) {
 	return b.postMessage(&msg, messageParameters, channelInfo)
 }
 
+func (b *Bslack) updateTopicOrPurpose(msg *config.Message, channelInfo *slack.Channel) (handled bool, err error) {
```
Helcaraxan: Please don't use named returns unless you absolute have to. Especially given that you combine them with only standard returns below.

They should normally only be used for very short functions as they tend to make a function's logic much less readable.

bridge/slack/slack.go
```diff
@@ -315,6 +321,39 @@ func (b *Bslack) sendRTM(msg config.Message) (string, error) {
...
+		return false, nil
+	}
+
+	incomingChangeType, text := b.extractTopicOrPurpose(msg.Text)
+	switch incomingChangeType {
+	case "topic":
+		if strings.HasSuffix(channelInfo.Topic.Value, "[nosync]") {
+			break
+		}
+		_, err = b.rtm.SetTopicOfConversation(channelInfo.ID, text)
```
Helcaraxan: Errors should be handled at this point directly:
```suggestion
		for {
			_, err = b.rtm.SetTopicOfConversation(channelInfo.ID, text)
			if err == nil {
				return true, nil
			}
			if err = b.handleRateLimit(err); err != nil {
				return true, err
			}
		}
```

bridge/slack/slack.go
```diff
@@ -315,6 +321,39 @@ func (b *Bslack) sendRTM(msg config.Message) (string, error) {
...
+	case "purpose":
+		if strings.HasSuffix(channelInfo.Purpose.Value, "[nosync]") {
+			break
+		}
+		_, err = b.rtm.SetPurposeOfConversation(channelInfo.ID, text)
```
Helcaraxan: Same here:
```suggestion
		for {
			_, err = b.rtm.SetTopicOfConversation(channelInfo.ID, text)
			if err == nil {
				return true, nil
			}
			if err = b.handleRateLimit(err); err != nil {
				return true, err
			}
		}
```

bridge/slack/slack.go
```diff
@@ -315,6 +321,39 @@ func (b *Bslack) sendRTM(msg config.Message) (string, error) {
...
+	}
+
+	if err != nil {
```
Helcaraxan: This logic is no longer necessary. If we reach this point (assuming my suggested changes above) we should log an error (as it should never happen). This best actually encoded in the `switch` itself by adding a `default:` case at the bottom. That way we don't need to specify a return after the `switch statement` as all cases are covered.

gateway/gateway.go
```diff
@@ -266,9 +266,15 @@ func (gw *Gateway) handleMessage(msg config.Message, dest *bridge.Bridge) []*BrM
...
 
-	// only relay topic change when configured
-	if msg.Event == config.EVENT_TOPIC_CHANGE && !gw.Bridges[dest.Account].GetBool("ShowTopicChange") {
-		return brMsgIDs
+	if msg.Event == config.EVENT_TOPIC_CHANGE {
+		// don't relay topic/purpose change if disabled
+		if !gw.Bridges[dest.Account].GetBool("ShowTopicChange") {
+			return brMsgIDs
+		}
+		// don't relay topic/purpose update if sync is enabled, but msg marked as nosync
```
Helcaraxan: This logic is not necessary. The Slack receiving handler logic added in this PR already takes care of this case.

bridge/mattermost/mattermost.go
patcon: Shouldn't it be exportable like config.EVENT_BLAH? Sorry, these conventions are new to me :)

bridge/slack/handlers.go
patcon: Sorry, I thought this was how golang documents returns for skimming functions. If this is never done, happy to revert

bridge/slack/slack.go
patcon: Is it possible to have named returns but still explicitly return values? Seems to be perfectly fine as documentation, no? https://tour.golang.org/basics/7

Helcaraxan: It is fine indeed syntax-wise. The core issue with named return variables is not the naming itself. It is that they implicitly allow for "naked returns" and those are the devils' own offspring and the source of many a production bug in systems ([the discussion here is a good read](https://github.com/golang/go/issues/21291)).

Given that the overall code of Matterbridge does not really have policy of using named returns I would at the moment suggest to not start introducing them randomly. We can potentially revisit this at a later point.

gateway/gateway.go
patcon: That receive handler drops based on the _existing_ topic/purpose having [nosync] flag set, as that channel shouldn't be overridden by an incoming topic update msg. This logic, on the other hand, keeps a topic update msg in one channel from propagating to others, as it should just be dropped.

I was torn, as this separation is admittedly confusing but seems the proper way (?) because sync can technically be enabled/disabled per-bridge. Open to other ideas!

patcon:
We could handle it all in the other func, and my previous commit did this, but seemed odd to have the msg stay in the debug logs from hitting each bridge, when it was easy to kill here

bridge/mattermost/mattermost.go
patcon: Ok cool, just so i understand and am learning consistently, then the constants in `config.EVENT_*` are the anti-pattern that we should fix later?

bridge/slack/slack.go
patcon: Yep, I def prefer them as inline documentation as I am sometimes confused whats coming back when skimming, but def happy to defer that convo :) will remove when off mobile!

patcon: Also, yay for this! https://github.com/alexkohler/nakedret

bridge/mattermost/mattermost.go
Helcaraxan: Yes indeed.

bridge/slack/handlers.go
Helcaraxan: I would rather add the checking for ` [nosync]` here.

gateway/gateway.go
Helcaraxan: Have indeed added a comment in the receiving handler. We should be handling the `nosync` there. The idea is to kill the message propagation in the bridge as soon as we can. Such specific logic should not be part of the gateway.

bridge/slack/helpers.go
```diff
@@ -261,12 +261,19 @@ func (b *Bslack) populateMessageWithBotInfo(ev *slack.MessageEvent, rmsg *config
...
-	variableRE = regexp.MustCompile(`<!((?:subteam\^)?[a-zA-Z0-9]+)(?:\|@?(.+?))?>`)
-	urlRE      = regexp.MustCompile(`<(.*?)(\|.*?)?>`)
+	mentionRE        = regexp.MustCompile(`<@([a-zA-Z0-9]+)>`)
+	channelRE        = regexp.MustCompile(`<#[a-zA-Z0-9]+\|(.+?)>`)
+	variableRE       = regexp.MustCompile(`<!((?:subteam\^)?[a-zA-Z0-9]+)(?:\|@?(.+?))?>`)
+	urlRE            = regexp.MustCompile(`<(.*?)(\|.*?)?>`)
+	topicOrPurposeRE = regexp.MustCompile(`(?s)^@.+ set the channel (topic|purpose): (.+?)$`)
 )
 
+func (b *Bslack) extractTopicOrPurpose(text string) (updateType string, extracted string) {
```
Helcaraxan: Another place where I would not add named return values.

bridge/slack/slack.go
42wim: Like @Helcaraxan I'm not a fan of named returns because of exactly the same reason

42wim:
I'm not sure about the "[nosync]" thing and introducing this kind of specific matterbridge metadata.
I think it's better to:
* or sync everything as specified in the general bridge config
* or if you want some more granularity it should be specified in the channel configuration. 
* or allow both options

The ChannelOptions struct could be used for this. 

```golang
type ChannelOptions struct {
        Key        string // irc, xmpp
        WebhookURL string // discord
        SyncTopic  bool // slack
}
```

Also better to just call it SyncTopic, because you probably want to have it always in sync, not only on a change ?

patcon:
I am happy to remove it for the simplicity of this PR, but I would also appreciate a discussion at some point. If you'd prefer to keep this PR without, and move this convo to another issue, please do just say the word :)

In absence of that request:

### Considerations

I've come back to @Helcaraxan's position of keeping it all in the request handler, and so I hope it wouldn't add too much maintenance burden or lines of code.

Core rationale: We can't add UI elements to most of our bridged chat tools, but we can add text interfaces to convenience features that _preserve the spirit of the tools_. Slack very intentionally (and I feel rightfully), let's everyone control the topic/purposes. On large gateways -- with hundreds of participants, communicating between languages and domains of expertise -- It feels quite significant to compel them all to share one context, and to put the admin in the middle of making config file tweaks to let them customize their channel topics/purposes to their needs. 

In my testing this feature works very well, and very smoothly, if you might consider allowing it :)

> introducing this kind of specific matterbridge metadata.

Perhaps we could have slight leeway on things like this ("text-based feature interfaces"), until such a time that we have other options, like figuring out remote reloading of channel config (https://github.com/42wim/matterbridge/issues/589). Once there are more robust and modular options, I'll be in total agreement that this should aspire to live outside in something like [matterbridge-autoconf](https://github.com/patcon/matterbridge-autoconfig) :)

42wim:
@patcon 

> Perhaps we could have slight leeway on things like this ("text-based feature interfaces"), until such a time that we have other options, like figuring out remote reloading of channel config (#589). Once there are more robust and separable options, I'll be in total agreement that this should aspire to live outside in something like matterbridge-autoconf :)

My stance about this is that "meta" `[nosync]` stuff, like putting logic in channel topics is **outside** of matterbridge scope. This is more `bot`-like behaviour and not bridging behaviour. If you want something like this in matterbridge it has to be done be configuration like using per channel configuration like I said in the previous comment.

Also I'd like to have `SyncTopicChange` be `SyncTopic` like mentioned above.

About the `bot`-behaviour, maybe in the future we can add plugin-support that hooks in on the send/receive functions of bridges so you can make your own plugins which can be basically bots, without impact on the core.

patcon:
@42wim What about `SyncChannelMeta` rather than `SyncTopic`? We'll already be syncing topic/purpose on Slack, and IRC calls it a description if anyone adds support for that, and I'm sure it has other names elsewhere.

bridge/slack/handlers.go
patcon: Ok, so we're not going to be doing `nosync` anymore, so technically I don't need this, but doesn't it make sense to keep the state of `ChannelInfo` up-to-date with purpose/topic? So I wonder if we should just run this on changes that knock it out of sync (like these), nevermind whether we're _yet_ deciding to process it.

If we're not updating, the inconsistency will likely hit someone else later who assumed ChannelInfo to be up-to-date.

tl;dr Let's leave this in to keep our internal representations matching reality

bridge/slack/slack.go
patcon: I feel this logically belongs outside, as it sets up a var to be used in many places below (it's not logically tied to that bit the comment refers to)

bridge/slack/slack.go
patcon: Why did we create duplication here? What is inappropriate about doing the repetitive part outside the switch?

Helcaraxan: I should have written the code in an editor, not in GitHub... indeed quite convoluted. However the original version was not good as it did not deal with rate-limiting correctly. Let me use my editor instead and post the suggestion here.

Helcaraxan: So here the suggestion for the whole function. Sorry for the hiccup:
```go
func (b *Bslack) updateTopicOrPurpose(msg *config.Message, channelInfo *slack.Channel) (bool, error) {
	if msg.Event != config.EVENT_TOPIC_CHANGE || !b.GetBool("SyncTopic") {
		return false, nil
	}

	incomingChangeType, text := b.extractTopicOrPurpose(msg.Text)

	var updateFunc func(channelID string, value string) (*slack.Channel, error)
	switch incomingChangeType {
	case "topic":
		updateFunc = b.rtm.SetTopicOfConversation
	case "purpose":
		updateFunc = b.rtm.SetPurposeOfConversation
	default:
		return true, fmt.Errorf("encountered unrecognised channel meta operation: %s", incomingChangeType)
	}

	for {
		_, err = updateFunc(channelInfo.ID, text)
		if err == nil {
			return true, nil
		}
		if err = b.handleRateLimit(err); err != nil {
			return true, err
		}
	}
}
```

bridge/slack/handlers.go
Helcaraxan: Yes. Do keep it.

patcon:
Ok, moved all the logic into the receiving side, as per @Helcaraxan's excellent suggestion. Also caught that we weren't handling a special case around clearing the topic/purpose.

Outstanding questions:
- naming: alt to topic? channelmeta? (Mattermost calls these properties header/purpose, and IRC calls them topic/description tho "description" is a chanserv prop, etc etc)
- do we care about that last codeclimate breakage? we seem to use more than 4 returns all over the place, to shortcircuit, and I'm unclear if we want to move away from that

bridge/slack/handlers.go
Helcaraxan: Please remove the named return.

bridge/slack/helpers.go
```diff
@@ -261,12 +261,25 @@ func (b *Bslack) populateMessageWithBotInfo(ev *slack.MessageEvent, rmsg *config
...
+	mentionRE        = regexp.MustCompile(`<@([a-zA-Z0-9]+)>`)
+	channelRE        = regexp.MustCompile(`<#[a-zA-Z0-9]+\|(.+?)>`)
+	variableRE       = regexp.MustCompile(`<!((?:subteam\^)?[a-zA-Z0-9]+)(?:\|@?(.+?))?>`)
+	urlRE            = regexp.MustCompile(`<(.*?)(\|.*?)?>`)
+	topicOrPurposeRE = regexp.MustCompile(`(?s)(@.+) (cleared|set)(?: the)? channel (topic|purpose)(?:: (.*))?`)
 )
 
+func (b *Bslack) extractTopicOrPurpose(text string) (string, string) {
+	r := topicOrPurposeRE.FindStringSubmatch(text)
+	action, updateType, extracted := r[2], r[3], r[4]
```
Helcaraxan: You are still assigning in a way that might very well panic if there are no matches and even when there is a match but the topic/purpose is being cleared. Please add the necessary checks.

bridge/slack/slack.go
```diff
@@ -265,6 +265,7 @@ func (b *Bslack) sendWebhook(msg config.Message) (string, error) {
 	return "", nil
 }
 
+// Note: returning an empty string skips the message
```
Helcaraxan: Why? This is not true. It would rather mean that the message will not be registered, regardless of whether it was sent or not according to the code [here](https://github.com/42wim/matterbridge/blob/master/gateway/gateway.go#L330-L339).

bridge/slack/slack.go
```diff
@@ -311,6 +318,48 @@ func (b *Bslack) sendRTM(msg config.Message) (string, error) {
...
+		if err == nil {
+			return true, nil
+		}
+		if err = b.handleRateLimit(err); err != nil {
+			return true, err
+		}
+	}
+}
+
+// handles updating topic/purpose and determining whether to further propagate update messages
```
Helcaraxan: Nit: please finish sentences with a dot, same for other comments in the function below.

bridge/slack/slack.go
```diff
@@ -311,6 +318,48 @@ func (b *Bslack) sendRTM(msg config.Message) (string, error) {
...
+func (b *Bslack) handleTopicOrPurpose(msg *config.Message, channelInfo *slack.Channel) (bool, error) {
+	if msg.Event != config.EVENT_TOPIC_CHANGE {
+		return false, nil
+	}
+
+	if b.GetBool("SyncTopic") {
+		return b.updateTopicOrPurpose(msg, channelInfo)
+	}
+
+	// Pass along to normal message handlers
```
Helcaraxan: I am not entirely convinced that it would make sense to send messages about a topic change on another bridge if it is not being reflected on the actual channel's topic. What's your use case?

bridge/slack/handlers.go
```diff
@@ -132,7 +137,12 @@ func (b *Bslack) skipMessageEvent(ev *slack.MessageEvent) bool {
 		ev.SubMessage.Edited == nil {
 		return true
 	}
-	return b.filesCached(ev.Files)
+
+	if len(ev.Files) > 0 {
```
Helcaraxan: Could you please create a separate PR for this fix? That will enable it to be merged ASAP rather than having to wait on all outstanding discussions to concluded. Also: this fix is entirely unrelated to this PR.

bridge/slack/handlers.go
```diff
@@ -132,7 +137,12 @@ func (b *Bslack) skipMessageEvent(ev *slack.MessageEvent) bool {
...
+		return b.filesCached(ev.Files)
+	}
+
```
Helcaraxan: Nit: unnecessary empty line.

bridge/slack/helpers.go
patcon: Ok, thanks, dealt with this and added my first test 🎉

I had to remove it from interface value to add the test, as I'm not clear on how to set up the whole test scaffolding with router, bridger, etc etc

bridge/slack/slack.go
patcon: If we don't do this, there'll be an inconsistent edge-case on Slack, so that when we didn't `SyncTopic`, but had `ShowTopicChange` enabled, messages would essentially get skipped.

bridge/slack/slack.go
patcon: Let's add this to linting pls. I'd rather be told this immediately by a robot if it'd going to block :)

42wim:
> naming: alt to topic? channelmeta? (Mattermost calls these properties header/purpose, and IRC calls them topic/description tho "description" is a chanserv prop, etc etc)

Let's just call it `SyncTopic` for now or are there any other meta stuff that needs syncing besides topic/purpose ? Also I'll bet that some people will request the feature to also have a `SyncPurpose` :-) 

If we make it clear in the documentation that `SyncTopic` syncs topic and the likes (purpose, description) it should be ok for now. 

> do we care about that last codeclimate breakage? we seem to use more than 4 returns all over the place, to shortcircuit, and I'm unclear if we want to move away from that

Well, you specifically wanted the codeclimate analysis, so now you have to deal with it :sunglasses: 

But I'm not sure about this, found a stackoverflow about it too https://stackoverflow.com/questions/36707/should-a-function-have-only-one-return-statement
@helcaraxan you any opinions?

Helcaraxan:
In this particular case the 5 returns is not a significant issue because the function is fairly linear. If you do want to eliminate it in a tricky way is to refactor the check for `len(ev.Files) > 0` into the `filesCached` method. It would end up being:

```go
func (b *Bslack) filesCached(files []slack.File) bool {
	if len(files) == 0 {
		return false
	}
	for i := range files {
		if !b.fileCached(&files[i]) {
			return false
		}
	}
	return true
}
```

patcon:
The 3 code climate comments seem out-of-scope, and this is already long enough, so I'm going to just leave it for either merge or further comment.

I can imagine some further conversation on moving the extractTopicOrPurpose function out of being an interface function. I did this so that it's simpler to test, and I could get some actuals tests in. Until we have time for a refactor (after which it's more clear how to test), this feels like the lesser of evils, but if this is a sticking point, let's discuss

bridge/slack/helpers_test.go
```diff
@@ -0,0 +1,26 @@
...
+		{"@someone set channel topic: one liner",    "topic",   "one liner"},
+		{"@someone set channel purpose: one liner",  "purpose", "one liner"},
+		{"@someone set channel topic: multi\nliner", "topic",   "multi\nliner"},
+		{"@someone cleared channel topic",           "topic",   ""},
+		{"some unmatched message",                   "unknown", ""},
+	}
+
+	for _, table := range tables {
+		changeType, output := extractTopicOrPurpose(table.input)
+		if changeType != table.changeType || output != table.output {
```
Helcaraxan: You should be using the `assert.Equal` methods here from the `assert` package that is already vendored in. You can see an example in the test that I recently added in the `bridge/helper` package. :)

bridge/slack/helpers_test.go
```diff
@@ -0,0 +1,26 @@
...
+		output string
+	}{
+		{"@someone set channel topic: one liner",    "topic",   "one liner"},
+		{"@someone set channel purpose: one liner",  "purpose", "one liner"},
+		{"@someone set channel topic: multi\nliner", "topic",   "multi\nliner"},
+		{"@someone cleared channel topic",           "topic",   ""},
+		{"some unmatched message",                   "unknown", ""},
+	}
+
+	for _, table := range tables {
```
Helcaraxan: Very nit, please ignore if you feel like it: `tables` / `table` variables could be replaced by `testcases` / `testcase` as it corresponds better to what they represent.

bridge/slack/slack.go
```diff
@@ -311,6 +317,48 @@ func (b *Bslack) sendRTM(msg config.Message) (string, error) {
...
+	var updateFunc func(channelID string, value string) (*slack.Channel, error)
+
+	incomingChangeType, text := extractTopicOrPurpose(msg.Text)
+	switch incomingChangeType {
+	case "topic":
+		updateFunc = b.rtm.SetTopicOfConversation
+	case "purpose":
+		updateFunc = b.rtm.SetPurposeOfConversation
+	default:
+		b.Log.Errorf("Unhandled type received from extractTopicOrPurpose: %s", incomingChangeType)
```
Helcaraxan: You need to return after this or else the function will panic below on attempting to call a `nil`-pointer function.

gateway/gateway.go
```diff
@@ -268,7 +268,14 @@ func (gw *Gateway) handleMessage(msg config.Message, dest *bridge.Bridge) []*BrM
 
 	// only relay topic change when configured
 	if msg.Event == config.EventTopicChange && !gw.Bridges[dest.Account].GetBool("ShowTopicChange") {
-		return brMsgIDs
+		// don't relay topic/purpose change if disabled
+		if !gw.Bridges[dest.Account].GetBool("ShowTopicChange") {
+			return brMsgIDs
+		}
+		// don't relay topic/purpose update if sync is enabled, but msg marked as nosync
+		if gw.Bridges[dest.Account].GetBool("SyncTopic") && strings.HasSuffix(msg.Text, "[nosync]") {
```
Helcaraxan: Unless I misunderstood we are not integrating the `[nosync]` functionality? cc @42wim.

gateway/gateway.go
```diff
@@ -268,7 +268,14 @@ func (gw *Gateway) handleMessage(msg config.Message, dest *bridge.Bridge) []*BrM
 
 	// only relay topic change when configured
 	if msg.Event == config.EventTopicChange && !gw.Bridges[dest.Account].GetBool("ShowTopicChange") {
```
Helcaraxan: I am still somewhat confused by the logic here. Mostly due to it not being fully clear to me yet what we are doing, w.r.t. configuration options.  Would it be possible, with the conversation that we have already had to update the PRs description with what the aim is?

In any case the logic here as it is is has an unnecessary duplicate in the check of `ShowTopicChange`.

bridge/slack/helpers_test.go
```diff
@@ -0,0 +1,26 @@
+package bslack
+
+import (
+	"testing"
+)
```
Helcaraxan: If there's only one import our `goimports` linter will want this to be on one line without the parentheses. 🤷‍♂️

gateway/gateway.go
patcon: wtf. You're 100% right. I think this somehow made it back in while I was rebasing out some other changes. omg this PR is the worst. Ugh. I'm so sorry guys!

bridge/slack/slack.go
Helcaraxan: This is not test code. This is production code. So just an appropriate return would be great. :)

patcon: haha oh dear. Ok, addressed in next push.

Say, is our process to let the person "raising" comment to mark as resolved, or is "marking as resolved" like a TODO list for me in this situation? I just realized I've been treating it like the latter, but not sure what our preference is...

cc: @42wim

bridge/slack/slack.go
patcon: > I am not entirely convinced that it would make sense to send messages about a topic change on another bridge if it is not being reflected on the actual channel's topic. What's your use case?

To come back to this: It's still important to share the context, even if I have sync disabled. For instance, I have translation, and on one side they keep a japanese topic. I don't want to clobber that, so I will disable sync, but still want to update to go through -- if I add a link, people can choose to update their own topic to suit their needs.

tl;dr - Difference between alerting of information change, and forcing it. Depends on the fuzzy relationships between each side of bridge :)

gateway/gateway.go
patcon: Ok, I think I simplified this: now it sends if there's any topic-related processing features enabled on the other end.

patcon:
Ok, this is ready for review again.

I know there are some other suggestions in here (like using `ChannelOptions`), but I'd prefer to treat all but the minimum as out-of-scope, and we can open new tickets for anything else :)

bridge/slack/helpers_test.go
```diff
@@ -0,0 +1,28 @@
+package bslack
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestExtractTopicOrPurpose(t *testing.T) {
+	testcases := []struct {
```
Helcaraxan: We should actually make this a `map[string]*[]struct` so that we can name test cases.

bridge/slack/helpers_test.go
```diff
@@ -0,0 +1,28 @@
...
+		{"@someone set channel topic: multi\nliner", "topic", "multi\nliner"},
+		{"@someone cleared channel topic", "topic", ""},
+		{"some unmatched message", "unknown", ""},
+	}
+
+	for _, tc := range testcases {
+		gotChangeType, gotOutput := extractTopicOrPurpose(tc.input)
+
+		assert.Equal(t, tc.wantChangeType, gotChangeType)
+		assert.Equal(t, tc.wantOutput, gotOutput)
```
Helcaraxan: It's quite handy to use the `assert.Equalf` variants here and provide an error message that tells (together with my comment above) the name of the testcase where the assertion failed.

bridge/slack/slack.go
Helcaraxan: @patcon use it as your own TODO checklist. I'll verify anyway regardless as a reviewer and wil re-open where relevant.

bridge/slack/helpers_test.go
patcon: That's a great idea :) after all, won't always fit in the text portion

bridge/slack/helpers_test.go
patcon: Interestingly, Equal and Equalf seem to be funcationally equivalent... But I like using Equalf by convention anyhow

bridge/slack/slack.go
patcon: Does this make sense? (It's the final hold-out, I think, but I don't see any change making sense, as we'd lose functionality.)

It's essentially skip logic, because if this bridge is not set to sync, but IS set to display messages, then we want to let message through to the regular message handling. Dropping this would lead to unexpected behaviour imho, and moving it elsewhere would seem to fragment the logic between places. Open to suggestions, but this feels like the right way

patcon:
Anyone know what that travis failure means?

> File is not `goimports`-ed

screen shot 2018-11-26 at 12 57 30 pm

Helcaraxan:
@patcon could you please clean-up the git history on this PR before we merge? (unless @wim42 is OK with squash-merging).

42wim:
Squash merge is fine for this one. 
Make sure the commit starts with `Sync channels ...` if you squash merge it.

Helcaraxan:
You can edit the squashed commit's subject and body yourself before you hit the merge button @42wim. :)

42wim:
Sure, but was suggesting that it was ok for you to merge it if the requested changes were done. :)

Helcaraxan:
@patcon did a really minor change to add a log-line and am going to merge when CI goes green. Though this would be faster than asking you to change it and then me review it again. Should be quite uncontroversial.

patcon:
Omg I had no idea I could create a Bslack object so easily in the tests! That was why I avoided that fwiw :) thanks!

Helcaraxan:
It's actually a little bit more complicated... sorry for the ongoing hiccups. But it _should_ be fine now. I underestimated the number of changes required for that log message. 🤦‍♂️

Helcaraxan:
@patcon 🎉 💥 congrats on this long-running but eventually fruitful PR!

patcon:
Haha thanks, friend-o :)

Issue #584, opened 2018-11-12
IRC Message-Split loses about 33 characters in the middle of the split


DarkSir23:
**Describe the bug**
When bridging Discord to IRC, Message-Split appears to lose about 33 characters.

Message on Discord Side:
Capheind: Fudge is a generic role-playing game systemfor use in freeform role-playing games.[1] The name "FUDGE" was once an acronym for Freeform Universal Donated (later, Do-it-yourself) Gaming Engine[2] and, though the acronym has since been dropped, that phrase remains a good summation of the game's design goals. Fudge has been nominated for an Origins Award for Best Role-Playing Game System for the Deryni Role-Playing Game

Message on IRC Side:
&lt;Capheind&gt; Fudge is a generic role-playing game systemfor use in freeform role-playing games.[1] The name "FUDGE" was once an acronym for Freeform Universal Donated (later, Do-it-yourself) Gaming Engine[2] and, though the acronym has since been dropped, that phrase remains a good summation of the game's design goals. Fudge has been nominated for an Origins Award for Best Role- <message clipped>
&lt;Capheind&gt; ni Role-Playing Game

**Environment (please complete the following information):**
 - OS: Ubuntu Linux 16.04
 - Matterbridge version: version: 1.11.4-dev

Helcaraxan:
Have confirmed the issue (codewise - don't have an IRC server available for testing). Am creating the PR that should fix it.

DarkSir23:
Awesome, thank you for such a quick response!

PR #586, opened 2018-11-13
Make config.Config more unit-test friendly

Helcaraxan:
## Making config.Config more unit-test friendly
In its current implementation the `config.Config` struct is not very welcoming to easily implement unit-tests that depend on a particular config value being set. This is mostly due to the inability to override values on the fly because the `viper.Viper` object included in `config.Config` is not exported (for good reason).

This PR proposes to solve this UX issue by:
1. Changing `config.Config` to make it an interface and moving the actual implementation into an unexported `config.config` struct. We already have factory methods anyway, limiting the code impact of this change outside of the package.
2. Implementing a `config.TestConfig` exported struct that implements `config.Config` but also allows for providing easy overrides for specific configuration values.

## Moving config-value retrieval to a two-value return model
In the process of doing this and testing its potential applications I also uncovered some strangeness / a potential bug in the implementation of the `Get<Type>()` methods of the current `config.Config` methods. This was due to the fact that these methods did not implement a two-value return but rather a single-value one. This made it impossible to distinguish between a value being set to the retrieved type's default instance or a value not being set, which in turn made the implementation logic quite dodgy and constrained.

The fix for this is luckily not too massive as the `config.Config` is only really accessed through `bridge.Config` which has its own `Get<Type>()` methods. We might want to move those as well to a two-value return model but that would be a much bigger PR and it is not fundamentally necessary at this point in time.

### DIFF
--- a/bridge/bridge.go
+++ b/bridge/bridge.go
@@ -22,7 +22,7 @@ type Bridge struct {
 	Channels map[string]config.ChannelInfo
 	Joined   map[string]bool
 	Log      *log.Entry
-	Config   *config.Config
+	Config   config.Config
 	General  *config.Protocol
 }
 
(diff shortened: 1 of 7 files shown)


42wim:
One small nitpick, maybe use `ConfigValues()` instead of `GetConfigValues()`
I couldn't do it for `GetString()` because it was too confusing, but I think it makes more sense for `ConfigValues()`

Helcaraxan:
@42wim I tried to do that... unfortunately the `config` struct (which implements the `Config` interface) contains an embedded `ConfigValues` struct. This means that `ConfigValues` cannot be reused as method name.

Helcaraxan:
BTW this PR is pre-requisite for code that I've already written to address #584. Once it's merged I'll create the PR to fix that issue.

42wim:
What do you think about making it explicit to `cv *ConfigValues` this only would need changes in `config.go`. (or not worth it?)

Helcaraxan:
Works too. Done.

PR #587, opened 2018-11-13
#584 Fix IRC line splitting

Helcaraxan:
Rewrite of the line-splitting functionality... including unit-tests. 🎉


42wim:
I think it would be useful to make `getSubLines` a `helper.go` function.
So it can be reused for other bridges that also have line length limits.
WDYT ?

Helcaraxan:
Sure. Will do.

42wim:
Great job, thanks! 👍

PR #593, opened 2018-11-15
Fix golint linter issues and enable it in CI

Helcaraxan:
Another linter one. More convoluted in number of lines but all have been done via automatic renaming tooling.


42wim:
You went a bit overboard ;-)
Could you leave the vendored ones out and also `matterclient/matterclient.go` and `matterhook/matterhook.go` as those are used in external projects too and breaks stuff if changed.

According to https://github.com/golangci/golangci-lint#false-positives we can add a `//nolint:golint` or maybe there's another way.

PR #594, opened 2018-11-16
[WIP] add base protocol concept to improve shared state between compatible bridge protocols

patcon:
mIDs are stored in cache as `slack <id>`, but msg.ParentID is generalized and so looks up by looking for `<protocol> <id>`

So with `slack-legacy`, it's looking for `slack-legacy <id>`, when they are instead stored as `slack <id>`.

## Suggestion

We could make a general convention about some specific character in protocol names being used to specify sub-protocols that are cross-compatible aside from connection methods. Then, we could split on that character, and get the root protocol for cache.

I'm happy for that character to remain as "hyphen". The only reason to choose something else might be bc it's so common in namespaces. Another char might "make itself known" more candidly to future explorers. Maybe "/"? Minor though :)

### DIFF
--- a/bridge/bridge.go
+++ b/bridge/bridge.go
@@ -68,6 +68,10 @@ func (b *Bridge) joinChannels(channels map[string]config.ChannelInfo, exists map
 	return nil
 }
 
+func (b *Bridge) BaseProtocol() string {
+	return strings.Split(b.Protocol, "-")[0]
+}
+
 func (b *Bridge) GetBool(key string) bool {
 	val, ok := b.Config.GetBool(b.Account + "." + key)
 	if !ok {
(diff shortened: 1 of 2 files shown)


bridge/bridge.go
```diff
@@ -68,6 +68,10 @@ func (b *Bridge) joinChannels(channels map[string]config.ChannelInfo, exists map
 	return nil
 }
 
+func (b *Bridge) BaseProtocol() string {
```
patcon: Should this be `GetBaseProtocol` so it's not assumed to be a struct key?

patcon:
Ready for review

42wim:
@patcon Instead of asking you to change everthing I made #596 which should fix the same issue as your PR here but in a more generic way for every bridge.

patcon:
Sounds good. I'll wait for that to land and then rejig this one for consideration

patcon:
I think I'm running into a related issue on a gateway that has both slack and slack-legacy. The threading breaks in odd ways.

Can we revisit why there is a need to break down caching by sub protocol? I'm not understanding the benefit, and the downside is that it will needlessly break persistence (when switching sub-protocols), and making it so that a lookup needs to know the specific type of protocol it might find the parent/origin msg filed under. Again, can we instead cache by something like a base protocol?

PR #557, opened 2018-11-16
Hint at thread replies when messages are unthreaded

patcon:
Lighter touch approach to https://github.com/42wim/matterbridge/issues/343

Unthreaded messages confuse the flow. It's a very different experience on different ends of bridge, depending on if messages are threaded or not.

Let's offer a small hint when things become unthreaded. For example, just prefix unthreaded messages with `thread reply: `. If that formatting is contentious, happy to use another. We could also template this, so that folks could override the default template of `thread reply: {MESSAGE}`.

- preserving threading when possible #529
- quoting parent message #343

**Additional context**
None.

### DIFF
--- a/bridge/slack/slack.go
+++ b/bridge/slack/slack.go
@@ -288,6 +288,12 @@ func (b *Bslack) sendRTM(msg config.Message) (string, error) {
 		return "", err
 	}
 
+	// Handle prefix hint for unthreaded messages.
+	if msg.ParentID == "unthreaded" {
+		msg.ParentID = ""
+		msg.Text = fmt.Sprintf("thread reply: %s", msg.Text)
+	}
+
 	// Handle message deletions.
 	if handled, err = b.deleteMessage(&msg, channelInfo); handled {
 		return msg.ID, err
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -263,7 +263,7 @@ func (gw *Gateway) handleMessage(msg config.Message, dest *bridge.Bridge) []*BrM
 	channels := gw.getDestChannel(&msg, *dest)
 	for _, channel := range channels {
 		// Performs rethreading.
-		msg.ParentID =  gw.getParentID(&msg, dest, channel)
+		msg.ParentID = gw.getParentID(&msg, dest, channel)
 
 		// Only send the avatar download event to ourselves.
 		if msg.Event == config.EventAvatarDownload {
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -56,6 +56,10 @@ func New(cfg config.Gateway, r *Router) *Gateway {
 
 // Find the canonical ID that the message is keyed under in cache
 func (gw *Gateway) FindCanonicalMsgID(protocol string, mID string) string {
+	if mID == "" {
+		return ""
+	}
+
 	ID := protocol + " " + mID
 	if gw.Messages.Contains(ID) {
 		return mID
@@ -250,13 +254,16 @@ func (gw *Gateway) handleMessage(msg config.Message, dest *bridge.Bridge) []*BrM
 		return brMsgIDs
 	}
 
+	// Perform single intensive operation to resolve parent msg ID through
+	// cache, which allows us to rethread messages.
+	gw.resolveCanonicalParentMsgID(&msg)
+
 	originchannel := msg.Channel
 	origmsg := msg
 	channels := gw.getDestChannel(&msg, *dest)
 	for _, channel := range channels {
-		if parentID, isThreaded := gw.handleThreading(&msg, dest, channel); isThreaded {
-			msg.ParentID = parentID
-		}
+		// Performs rethreading.
+		msg.ParentID =  gw.getParentID(&msg, dest, channel)
 
 		// Only send the avatar download event to ourselves.
 		if msg.Event == config.EventAvatarDownload {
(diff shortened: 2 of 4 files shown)


patcon:
Could maybe add something like a `SupportsThreading()` function to Bridger interface or somewhere comparable, so that we're not hardcoding slack-specific stuff, but just asking the bridge itself if it supports threads. Not sure what's best, but the ParentID processing seems like it could be a bit more clear :)

42wim:
Would like to have the logic completely in slack for modifying the message.
Maybe setting ParentID to a specific value (eg `0` ) means that you didn't find the real ParentID and the slack bridge can handle this itself instead of the gateway

patcon:
Sounds good. Since we're not confined to numeric, which about something a little more self-explanatory, like `unavailable` or `unknown`?

patcon:
Wait. How did you mean when you said you wanted this in slack? If we moved this into the bridge, then non-threaded platforms wouldn't get the hint. Don't we want it to happen in the gateway, so that everyone understands that the context is different on the originating end?

Also, for example, if someone responds to an old message in a slack thread, they will be brief and noncontextual, since it is obvious what they mean on their end. But in gitter, it will just be a quick message, that others might assume is about current, main-channel convo. This would be wrong. Prefixing it with "thread reply:" will let people know that the context is missing. (Later, we can even resolve this reply to the parent message, and hyperlink replies to the parent message in gitter.)

tl;dr - this hint is useful for all protocols, and can be made even more useful once persistent cache lands :)

patcon:
Regardless, I moved it into slack bridge before thinking through on the above. Happy to rework if the above appeals to you :)

Otherwise ready for review!

patcon:
To test this branch:
1. enable threading on a gw between two channels on same slack team
2. post a message
3. reply to that message. (threading will work)
4. reply to an old message from before restart. (will see prefix)
5. edit the origin message of that last one. (will see msg update, with prefix maintained)
6. reply to unthreaded msg with prefix. (will see reply appear in original thread)
7. turn off bridge, disable threading, restart.
8. create threads and see all message unthread with prefix.

gateway/gateway.go
```diff
@@ -280,16 +280,14 @@ func (gw *Gateway) handleMessage(msg config.Message, dest *bridge.Bridge) []*BrM
...
-	var canonicalParentMsgID string
-	if msg.ParentID != "" && (gw.BridgeValues().General.PreserveThreading || dest.GetBool("PreserveThreading")) {
-		canonicalParentMsgID = gw.FindCanonicalMsgID(msg.Protocol, msg.ParentID)
-	}
-
 	originchannel := msg.Channel
 	origmsg := msg
 	channels := gw.getDestChannel(&msg, *dest)
 	for _, channel := range channels {
+		if parentID, isThreaded := gw.handleThreading(&msg, dest, channel); isThreaded {
```
42wim: Why do we need a `isThreaded` bool here and in `handleThreading` ?
Is there a reason why we can't always give `msg.ParentID` a value?

patcon: I think I was just copying the `handled bool` style that @Helcaraxon used, which I found very legible compared to setting based on opaque internals of function.

But if on consideration that's not helpful, I think it should be possible to set directly. Let me double-check on a complex gateway though :)

bridge/slack/slack.go
```diff
@@ -288,6 +288,12 @@ func (b *Bslack) sendRTM(msg config.Message) (string, error) {
 		return "", err
 	}
 
+	// Handle prefix hint for unthreaded messages.
+	if msg.ParentID == "unthreaded" {
+		msg.ParentID = ""
+		msg.Text = fmt.Sprintf("thread reply: %s", msg.Text)
```
patcon: @42wim I moved this into the slack bridge as per your request, but in retrospect, why is it that we're not notifying other protocols of unthreading?

42wim: Other protocols also get notified ? As the notification is inside the `msg.ParentID`?

42wim:
Some initial remarks:
* don't like the use of "canonical_parent_mid" in msg.Extra especially as this isn't used to get extra info to the bridges
* would like it to be simpler and more comments, especially documenting the function according to https://golang.org/doc/effective_go.html#commentary

(haven't tested it yet)

patcon:
> don't like the use of "canonical_parent_mid" in msg.Extra especially as this isn't used to get extra info to the bridges

Hm. Ok, I'm struggling with imagining rolling back this bit. It's the product of a nontrivial resolution, and it feels odd to not have a place to hold it in the message object. When I get to working on reactions, this feels like it will also be helpful so that the message can be passed around more easily and processed. What else would you consider for holding the resolved parent ID on the message object, if anything? Thanks!

42wim:
Could you take a look at https://github.com/42wim/matterbridge/commit/3d6c8e448196b7656e3807c37d38f58cb8c476f4 ?
This behaves exactly the same as your patch (tested with the steps above) but is much simpeler

patcon:
Thanks Wim. Yes, it's shorter, which is great! but for a new reader, the parentID logic is really not clear. The goal of `IsThreadable` was to abstract away that very particular logic (which might change). I think I need to try to better understand our shared legibility goals of the codebase. I have been leaning (in what I consider my more thoughtful moments) toward adding small human-readable functions for these, but it seems I keep bumping up against some other set of criteria (not as simple as line count) that I don't quite grasp yet.

But regardless, thank you! I'll test later

42wim:
>  I might wonder, why exactly does having those criteria mean that it's untheaded? 

That's why I left a comment above the line :)

> I think I need to try to better understand our shared legibility goals of the codebase.

My goal is trying to "keep it as simple as possible but not simpler". 
Don't do too many abstractions if they're not really necessary. 
We can always refactor later if it's really needed. 
But this is of course very subjective.

patcon:
Thanks for helping me sync up :)

42wim:
@patcon so should I make a PR of https://github.com/42wim/matterbridge/commit/3d6c8e448196b7656e3807c37d38f58cb8c476f4 did you test it?

42wim:
Going to close this in favor of #684 in a few days if no further feedback.

PR #596, opened 2018-11-16
Add protocol to msg.ID in cache

42wim:
This is a more generic solution for #594 and I fixes slack not always returning `slack id`
We now prepend the protocol of each bridge to each msgID that gets cached (and remove it again when sending it back to the bridge)

### DIFF
--- a/gateway/router.go
+++ b/gateway/router.go
@@ -109,8 +109,8 @@ func (r *Router) handleReceive() {
 					msgIDs = append(msgIDs, gw.handleMessage(msg, br)...)
 				}
 				// only add the message ID if it doesn't already exists
-				if _, ok := gw.Messages.Get(msg.ID); !ok && msg.ID != "" {
-					gw.Messages.Add(msg.ID, msgIDs)
+				if _, ok := gw.Messages.Get(msg.Protocol + " " + msg.ID); !ok && msg.ID != "" {
+					gw.Messages.Add(msg.Protocol+" "+msg.ID, msgIDs)
 				}
 			}
 		}
--- a/bridge/slack/slack.go
+++ b/bridge/slack/slack.go
@@ -321,10 +321,8 @@ func (b *Bslack) deleteMessage(msg *config.Message, channelInfo *slack.Channel) 
 		return true, nil
 	}
 
-	// If we get a "slack <ID>", split it.
-	ts := strings.Fields(msg.ID)
 	for {
-		_, _, err := b.rtm.DeleteMessage(channelInfo.ID, ts[1])
+		_, _, err := b.rtm.DeleteMessage(channelInfo.ID, msg.ID)
 		if err == nil {
 			return true, nil
 		}
@@ -341,9 +339,8 @@ func (b *Bslack) editMessage(msg *config.Message, channelInfo *slack.Channel) (b
 		return false, nil
 	}
 
-	ts := strings.Fields(msg.ID)
 	for {
-		_, _, _, err := b.rtm.UpdateMessage(channelInfo.ID, ts[1], msg.Text)
+		_, _, _, err := b.rtm.UpdateMessage(channelInfo.ID, msg.ID, msg.Text)
 		if err == nil {
 			return true, nil
 		}
@@ -359,7 +356,7 @@ func (b *Bslack) postMessage(msg *config.Message, messageParameters *slack.PostM
 	for {
 		_, id, err := b.rtm.PostMessage(channelInfo.ID, msg.Text, *messageParameters)
 		if err == nil {
-			return "slack " + id, nil
+			return id, nil
 		}
 
 		if err = b.handleRateLimit(err); err != nil {
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -228,13 +228,13 @@ func (gw *Gateway) getDestChannel(msg *config.Message, dest bridge.Bridge) []con
 }
 
 func (gw *Gateway) getDestMsgID(msgID string, dest *bridge.Bridge, channel config.ChannelInfo) string {
-	if res, ok := gw.Messages.Get(msgID); ok {
+	if res, ok := gw.Messages.Get(dest.Protocol + " " + msgID); ok {
 		IDs := res.([]*BrMsgID)
 		for _, id := range IDs {
 			// check protocol, bridge name and channelname
 			// for people that reuse the same bridge multiple times. see #342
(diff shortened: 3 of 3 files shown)


patcon:
Thanks wim! but this feels less ideal, as now, switching between legacy and main slack providers will break a future persistent cache in a way that need not occue. Meaning that past messages won't be editable or threadable after switching provider implementations. This seems an unneccary limitation?

This will work for now, but I feel that we'll have to come back to discussing something like a base protocol in pursuit of landing a persistent cache option :)

patcon:
Also, this still leave "slack-legacy" in the user-facing output of RemoteNickFormat. Which needlessly exposes what is essentially protocol connection logic (since that's the only difference between these protocols) to end-users

patcon:
Actually, since this solves the immediate bug just fine and seems an improvement regardlesa, I'll take the base protocol convo back to the other thread :)

patcon:
Something funny happening here with keys being save a ` <protocol> <id>` (note the prefixing space). Investigating

patcon:
Ok, found some other places "slack" is hardcoded. Grep for `"slack` (with the quote) to find them

gateway/router.go
```diff
@@ -109,8 +109,8 @@ func (r *Router) handleReceive() {
 					msgIDs = append(msgIDs, gw.handleMessage(msg, br)...)
 				}
 				// only add the message ID if it doesn't already exists
-				if _, ok := gw.Messages.Get(msg.ID); !ok && msg.ID != "" {
-					gw.Messages.Add(msg.ID, msgIDs)
+				if _, ok := gw.Messages.Get(msg.Protocol + " " + msg.ID); !ok && msg.ID != "" {
+					gw.Messages.Add(msg.Protocol+" "+msg.ID, msgIDs)
```
patcon: I'm seeing in testing that protocol is already reflected in msg.ID at this point, and so isn't needed. (This is the origin of the extra space as well, as `msg.Protocol` is empty)

Also, msg.Protocol isn't populated anymore, so this is an empty string. I think this got lost in a refactor, and we should bring it back in `populateReceivedMessage` even if it isn't helpful here :)

patcon:
Ok, I now see that I misunderstood you as trying to add Protocol to all instance of MsgID, when in fact you were removing it from everywhere except the cache ID. Still sorting on how to get this working, as I've made the minor fixes I mentioned above, and something is still not working

42wim:
What exactly is not working, can you give an example?

patcon:
Arrrgh. Ok, so so much of this was because buildMessageParameters needed an update. I should have looked there first

Anyhow, got a fix. Will post a diff

42wim:
Did some fixes for the `slack ` I forgot to remove

patcon:
I was a little turned around for a bit, so consider my last attempt misguided.

But it actually came down to `msg.Protocol` no longer being populated added on the msg object:

https://github.com/42wim/matterbridge/blob/ce21ba154585395e792c059ac156299071405247/bridge/slack/helpers.go#L172-L178

Complicated by the fact that prepareMessageParameters still needed an update:

https://github.com/42wim/matterbridge/blob/3c86d7e9d84f7796bacb920e35c6c7a62c28e516/bridge/slack/slack.go#L410-L413

I've made the change, and added test scaffold for `findCanonicalMsgID` here (took some reworking of method):
https://github.com/42wim/matterbridge/pull/597

42wim:
Please don't do PR's on PR's this way, it's gets confusing and time consuming fast.
I've added some changes of your PR, please comment on this PR if it's working or what's not and what you think is missing.

patcon:
Ok, but GitHub has a way to allow this workflow, but you'd need to allow me to use suggestion mode. Otherwise, it's just talk instead of code

You may have guesses that i'm not the best coder, but I 100% do not ever push without consent. And so to be totally hoenst, I'm a little discouraged to still feel so much like an outsider here -- no ability to label, no ability to self-assign tickets, and not ability to suggest

42wim:
I've been trying to reach you on chat for a few days now, but to no avail

42wim:
@patcon if you have time please join the chat, or DM on gitter, discord
Just updated the commit and hopefully it now fixes your issue.

42wim:
> Also, this still leave "slack-legacy" in the user-facing output of RemoteNickFormat. Which needlessly exposes noise of what is essentially protocol connection details (since that's the only difference between these protocols) to end-users

You can use the `Label="slack"` option and also use `{LABEL}` in the `RemoteNickFormat` if you don't want to see slack-legacy

patcon:
Almost :) but we need a message ID back from this function, and right now we're instead getting what is effectively a different cacheID:

https://github.com/42wim/matterbridge/blob/a2d1e6559e9c8aafa4511c24bf58865aebc9068d/gateway/gateway.go#L88-L104

This is the minimum to fix this:

```diff
diff --git a/gateway/gateway.go b/gateway/gateway.go
index 5270bae..935de39 100644
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -87,7 +87,7 @@ func New(cfg config.Gateway, r *Router) *Gateway {
 // Find the canonical ID that the message is keyed under in cache
 func (gw *Gateway) FindCanonicalMsgID(mID string) string {
 	if gw.Messages.Contains(mID) {
-		return mID
+		return strings.Fields(mID)[1]
 	}
 
 	// If not keyed, iterate through cache for downstream, and infer upstream.
@@ -96,7 +96,7 @@ func (gw *Gateway) FindCanonicalMsgID(mID string) string {
 		ids := v.([]*BrMsgID)
 		for _, downstreamMsgObj := range ids {
 			if mID == downstreamMsgObj.ID {
-				return mid.(string)
+				return strings.Fields(mid.(string))[1]
 			}
 		}
 	}
```

You might want to update IDs to reflect that we're dealing with two types now: `cacheID` in, and `msgID` out.

Might also want to check for a space in the input string, or else `strings.Fields` might panic, but maybe that's not necessary? :)

42wim:
I've solved it a bit differently and just setup a 2nd slack bridge to test it, to me it seems to work as it should now. But please test one more time :)
Your feature about the thread preserving is pretty cool btw!

I'm even wondering if there is a reason this option shouldn't be on by default? 
And rename it to `NoPreserveThreading`

patcon:
heh thanks :) I'm ok with either default, but I've hit a couple considerations for you that I'll drop in chat

patcon:
Should we call the first arg in `getDestMsgID()` something besides `msgID`? mID/mid/msgid, etc is really overloaded :) `cacheID` feels unique and truthful to how we use it, but anything else would be an improvement, even `brmsgid`

Issue #574, opened 2018-11-09
[Slack] With file upload, include author in file description, not blank message after file


patcon:
(I realize that posting as the token/bot user is necessary.)

This is what currently happens. File description is just a way to make it a little less noisy

![screen shot 2018-11-09 at 12 36 04 pm](https://user-images.githubusercontent.com/305339/48243615-39847780-e41c-11e8-93be-dff5a14f6e9a.png)

PR #606, opened 2018-11-24
Make sure threaded files stay in thread (slack). Fixes #590

### LINKED ISSUE #590: [Slack] shared files don't stay threaded
**Describe the bug**
When sharing a file in a thread, the file itself goes into the main channel, even when threading should otherwise work. The comment message (often blank, which might be its own bug) ends up in the thread as expected.

**To Reproduce**
Turn on threading. Connect two slack channels on same team across gateway. create thread. send file in thread.

**Expected behavior**
See above.

**Screenshots/debug logs**
No logs at this time. But can get later if needed.
screen shot 2018-11-14 at 7 23 18 pm

 - OS: darwim
 - Matterbridge version: HEAD of master from a few days ago.

**Additional context**
None.

42wim:
@patcon please test :)
It doesn't seem possible to impersonate the user that uploads the file, I've tried it using a legacy api token and a new bot token.

Also no hints on https://api.slack.com/methods/files.upload that this is possible

@Helcaraxan any ideas ?

### DIFF
--- a/bridge/slack/slack.go
+++ b/bridge/slack/slack.go
@@ -356,6 +356,10 @@ func (b *Bslack) editMessage(msg *config.Message, channelInfo *slack.Channel) (b
 }
 
 func (b *Bslack) postMessage(msg *config.Message, messageParameters *slack.PostMessageParameters, channelInfo *slack.Channel) (string, error) {
+	// don't post empty messages
+	if msg.Text == "" {
+		return "", nil
+	}
 	for {
 		_, id, err := b.rtm.PostMessage(channelInfo.ID, msg.Text, *messageParameters)
 		if err == nil {
--- a/bridge/slack/slack.go
+++ b/bridge/slack/slack.go
@@ -389,11 +389,16 @@ func (b *Bslack) uploadFile(msg *config.Message, channelID string) {
 		ts := time.Now()
 		b.Log.Debugf("Adding file %s to cache at %s with timestamp", fi.Name, ts.String())
 		b.cache.Add("filename"+fi.Name, ts)
+		initialComment := fmt.Sprintf("File from %s", msg.Username)
+		if fi.Comment != "" {
+			initialComment += fmt.Sprintf("with comment: %s", fi.Comment)
+		}
 		res, err := b.sc.UploadFile(slack.FileUploadParameters{
-			Reader:         bytes.NewReader(*fi.Data),
-			Filename:       fi.Name,
-			Channels:       []string{channelID},
-			InitialComment: fi.Comment,
+			Reader:          bytes.NewReader(*fi.Data),
+			Filename:        fi.Name,
+			Channels:        []string{channelID},
+			InitialComment:  initialComment,
+			ThreadTimestamp: msg.ParentID,
 		})
 		if err != nil {
 			b.Log.Errorf("uploadfile %#v", err)


42wim:
Added another commit that fixes #574

bridge/slack/slack.go
```diff
@@ -386,10 +390,11 @@ func (b *Bslack) uploadFile(msg *config.Message, channelID string) {
...
 		b.cache.Add("filename"+fi.Name, ts)
 		res, err := b.sc.UploadFile(slack.FileUploadParameters{
-			Reader:         bytes.NewReader(*fi.Data),
-			Filename:       fi.Name,
-			Channels:       []string{channelID},
-			InitialComment: fi.Comment,
+			Reader:          bytes.NewReader(*fi.Data),
+			Filename:        fi.Name,
+			Channels:        []string{channelID},
+			InitialComment:  fi.Comment,
```
patcon: I actually assumed the empty message was to hint at who posted. Could we instead append/prepend the poster's username to the `InitialComment`? Since the file can be either uploaded or "created" (meaning that some content was posted to create the file within slack), then we could use a neutral phrase like:

> File from @username

> File from @username with comment: I am commenting on this file.

My understanding is that these are all the types of files that appear through this method: uploaded files, code snippets, shared docs. (all with and without initial comment to announce.)

screen shot 2018-11-25 at 12 00 51 pm

patcon:
Also, re: not being able to impersonate users on file upload: I had the thought that we could choose to always leave `initial_comment` blank, and instead post a new comment, which can be , so that their face shows in the thread a little more clearly.

But then that throws things off, because the original comment will be a "thread parent" file message, and comments on the other end of the bridge will be child messages of the "thread parent" file message. This seems it would be confusing.

So in the end, despite Slack docs saying that `initial_comment` is a ["short-term accommodation"](https://api.slack.com/changelog/2018-05-file-threads-soon-tread), we should just keep doing what we're doing, with perhaps a modification for including uploader's username :)

bridge/slack/slack.go
```diff
@@ -356,6 +356,10 @@ func (b *Bslack) editMessage(msg *config.Message, channelInfo *slack.Channel) (b
 }
 
 func (b *Bslack) postMessage(msg *config.Message, messageParameters *slack.PostMessageParameters, channelInfo *slack.Channel) (string, error) {
+	// don't post empty messages
+	if msg.Text == "" {
```
Helcaraxan: This doesn't look right to me. What's the rationale behind this? If it's simply a file share than there might not be any text originally associated with it.

Also, if we do expect to always have text on any message then we should ensure that in all the inbound message pipelines on the bridges.

In the case of Slack's inbound pipeline we already check for wellformed messages [with the following](https://github.com/42wim/matterbridge/blob/c6fd65d1d7f5085890c0a00d37be8d6abb458921/bridge/slack/handlers.go#L181-L190):
```go
	if len(ev.Files) == 0 && (rmsg.Text == "" || rmsg.Username == "") {
		if ev.BotID != "" {
			// This is probably a webhook we couldn't resolve.
			return nil, fmt.Errorf("message handling resulted in an empty bot message (probably an incoming webhook we couldn't resolve): %#v", ev)
		}
		return nil, fmt.Errorf("message handling resulted in an empty message: %#v", ev)
	}
```
This would tend to say that we most definitely accept messages with no text as long as they contain files.

bridge/slack/slack.go
Helcaraxan: I would personally prefer the:
> File from @username

or

> File from @username with comment: I am commenting on this file.

over the other suggestion of simply adding the comment after a newline.

patcon: Yeah, I added the newline because the Username by this point is often `patcon <CHANNEL>:` but sometimes <patcon> or whatever, depending on what RemoteNickFormat says. Newline skips having to care if we're doubling up the colon or whatever. I'm fine with whatever, but that's the rationale for the newline :)

patcon:
@42wim ah yes, and tested and for UX alone, seems an improvement as-is 🎉

bridge/slack/slack.go
42wim: The rationale is to fix https://github.com/42wim/matterbridge/issues/574
Yes, we accept messages with no text as they contain files, but those file messages get handled by `uploadFile` (which handles the file and file comment) before calling `postMessage`.  This means that `postMessage` will just post an empty line (if the text is empty, eg by slack regardless of we have a file comment) which we don't want.

42wim:
I've updated the patch with the additional comment changes (without newlines).
I think it's okay now to get merged ?

bridge/slack/slack.go
```diff
@@ -385,11 +389,16 @@ func (b *Bslack) uploadFile(msg *config.Message, channelID string) {
 		ts := time.Now()
 		b.Log.Debugf("Adding file %s to cache at %s with timestamp", fi.Name, ts.String())
 		b.cache.Add("filename"+fi.Name, ts)
+		initialComment := fmt.Sprintf("File from %s", msg.Username)
+		if fi.Comment != "" {
+			initialComment += fmt.Sprintf("with comment: %s", fi.Comment)
```
Helcaraxan: There's not going to be a space between the existing message and the addition. I'd suggest:
```suggestion
			initialComment += fmt.Sprintf("%s with comment: %s", initialComment, fi.Comment)
```

42wim: Your suggestion gives:
image

How it currently works for me:
image

The space should normally already be there because you should've set in the `RemoteNickFormat` :) (which should be visible if you run with `PrefixMessagesWithNick=true`)

I don't agree to adding extra spaces here

Helcaraxan: `RemoteNickFormat` should not end with a space as it is not necessary when you are rewriting the posting-pseudo.

We should in general not be relying on users specifying such essentials things themselves. The input should be sanitised by removing spaces at front and end and then we should add spaces where we need them in the Slack bridge itself.

42wim: I understand your pov, but if we do this we'll probably get requests of users that don't want a default space after the nick.
The reasoning was to explicitly not set spaces so users have full control about how the username is shown by modifying the `RemoteNickFormat`.

PR #609, opened 2018-11-25
Refactor telegram handlers. Fix linting

42wim:
Moves handlers to separate file, fixes linting errors.


42wim:
@TomTheDragon as you are a telegram user, could you try the latest dev builds, to see if I didn't break anything regarding to telegram :) Thanks!

PR #610, opened 2018-11-26
DRY out getChannelsByX functions to make codeclimate happy

patcon:
Reticketed from https://github.com/42wim/matterbridge/pull/585#issuecomment-441519531

### DIFF
--- a/bridge/slack/helpers.go
+++ b/bridge/slack/helpers.go
@@ -44,23 +44,21 @@ func (b *Bslack) getChannel(channel string) (*slack.Channel, error) {
 }
 
 func (b *Bslack) getChannelByName(name string) (*slack.Channel, error) {
-	b.channelsMutex.RLock()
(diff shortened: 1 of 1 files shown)


patcon:
I'm on the fence too, but fwiw, the diff looks messier than reality

Helcaraxan:
From my PoV (and although I ❤️ all refactoring efforts) this looks like we are exchanging 9 lines for 7 (an insignificant change) for an extra layer of indirection that makes understanding this piece of code that more complicated (not much more complex but still an extra layer of complexity).

It's not a big fuss so I think we are good to merge. Just wanted to expand on my line of thought.

patcon:
If we don't find this legible, or having logic in one place instead of two isn't valued here, then let's alternatively adjust the linter globally so this doesn't keep distracting when working in this file. I'm fine with that, but otherwise, I think this is worth merging :)

PR #616, opened 2018-11-28
Tidy up fetching of config values

patcon:
Seems like most of this is holdover from before we were checking both levels in `bridge.GetX('String')`?

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -263,14 +263,14 @@ func (gw *Gateway) handleMessage(msg config.Message, dest *bridge.Bridge) []*BrM
 	}
 
 	// only relay join/part when configured
-	if msg.Event == config.EventJoinLeave && !gw.Bridges[dest.Account].GetBool("ShowJoinPart") {
+	if msg.Event == config.EventJoinLeave && !dest.GetBool("ShowJoinPart") {
 		return brMsgIDs
 	}
 
 	// only relay topic change when used in some way on other side
 	if msg.Event == config.EventTopicChange &&
-		!gw.Bridges[dest.Account].GetBool("ShowTopicChange") &&
-		!gw.Bridges[dest.Account].GetBool("SyncTopic") {
+		dest.GetBool("ShowTopicChange") &&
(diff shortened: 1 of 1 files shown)


gateway/gateway.go
```diff
@@ -439,10 +436,7 @@ func (gw *Gateway) modifyUsername(msg config.Message, dest *bridge.Bridge) strin
 }
 
 func (gw *Gateway) modifyAvatar(msg config.Message, dest *bridge.Bridge) string {
-	iconurl := gw.BridgeValues().General.IconURL
-	if iconurl == "" {
-		iconurl = dest.GetString("IconURL")
```
patcon: Note that this actually changes behaviour, but I don't think it was intentional thought: IT used to be that if icon url was set in general, it wouldn't be checked in this account, but now it only gets general if account value not set

gateway/gateway.go
```diff
@@ -282,7 +282,7 @@ func (gw *Gateway) handleMessage(msg config.Message, dest *bridge.Bridge) []*BrM
 
 	// Get the ID of the parent message in thread
 	var canonicalParentMsgID string
-	if msg.ParentID != "" && (gw.BridgeValues().General.PreserveThreading || dest.GetBool("PreserveThreading")) {
+	if msg.ParentID != "" && dest.GetBool("PreserveThreading") {
```
42wim: This removes  `PreserveThreading` from `general`, why?
(can also be used for mattermost when implemented)

gateway/gateway.go
```diff
@@ -394,14 +394,11 @@ func (gw *Gateway) ignoreMessage(msg *config.Message) bool {
 func (gw *Gateway) modifyUsername(msg config.Message, dest *bridge.Bridge) string {
 	br := gw.Bridges[msg.Account]
 	msg.Protocol = br.Protocol
-	if gw.BridgeValues().General.StripNick || dest.GetBool("StripNick") {
+	if dest.GetBool("StripNick") {
```
42wim: Same question as above, not in general anymore?

Issue #576, opened 2018-11-09
Improved unit-testing


Helcaraxan:
**Is your feature request related to a problem? Please describe.**
My recent work on the Slack bridge has shown that it is hard to make any changes in some of the more complex message pipelines that are contained in each bridge without potentially breaking things in subtle ways.

**Describe the solution you'd like**
We should be attempting to work to improve the way in which the Matterbridge project adheres to engineer best-practices by significantly increasing the amount of unit-testing (for lack of the ability to do actual end-to-end testing). This is obviously not a short or easy task but rather should be a prolonged low-intensity effort.

My default goal would be to aim for at least 60/70% of test coverage in order to close this issue (which we can monitor with the newly added monitoring: https://codeclimate.com/github/42wim/matterbridge - thanks for the idea @patcon !).

**Describe alternatives you've considered**
Not adding testing and risking subtle breakages at every commit that touches core logic. Not ideal :trollface: 

**Additional context**
As usual, when proposing such an FR I would not do it without being ready to invest my own time in this as well. However given the larger scale effort that would be required (and because teamwork is always more pleasant) I'd rather share the idea with others as well. 🙂

patcon:
Yay! I am pretty jazzed to learn more about go testing

Speaking of https://github.com/42wim/matterbridge/issues/534 and static analysis, that thread still feels unresolved. @wim42 I'm super grateful you set up the service, but I'd love to go a step further really see us get the commit status checks showing up in pull requests. Let's get code quality into the code review process, right in with our conversations :) 🎉 

You raised some valid concerns about authorizations, and I think there are two very good approaches to getting both 1) you concerns about minimizing exposure to service, and 2) us getting good feedback in our PRS.

1. Authorize the app, and then revoke all access except "repo:status". Someone can run a test PR to ensure this still works as expected.
2. Alternatively, if the above doesn't work, we can poof a new organization and transfer the repo there.

Thoughts @42wim and @Helcaraxan? Happy to move this convo elsewhere if it's conflating goals :)

42wim:
Mixed feelings: very happy that you want to take time to write more tests (and feeling bad that they don't exist yet because of my lazyness ;-)

@patcon Please don't hijack this issue about codeclimate. I'll add further comments about codeclimate in #534

Helcaraxan:
@42wim, just to be clear, I was not complaining. Building OSS is hard, doing testing in OSS is even two orders of magnitude harder than that. 🙂

[...]

PR #623, opened 2018-11-30
Start documenting test scenarios

patcon:
I'm sure the deciding to use any specific tool should get more discussion, but just wanted to suggest that perhaps we could start keeping tracking of the scenarios that we know can create issues. Even if we never execute them, we can still use them to write future tests using another framework. Also, we can start asking people who encounter bugs to write scenarios for us, so that we can test later. This is a pretty low bar for participation. No everyone writes golang or tests, but everyone can look at the scenarios and make their own us :)

### DIFF
--- a/internal/features/threading/replies.feature
+++ b/internal/features/threading/replies.feature
@@ -55,3 +55,9 @@ Feature: Thread Reply support
     And I delete the parent message in #foo
     And I post another reply in #bar
     Then the final reply should appear in #foo
+
+  Scenario: User replies back across subprotocols
+    Given I have a simple gateway between slack#foo, slack#bar and slack-legacy#baz
+    And I post a message in #foo
+    And I post a reply in #baz
+    Then the reply should appear in #foo and #bar
--- a/internal/features/threading/replies.feature
+++ b/internal/features/threading/replies.feature
@@ -1,0 +1,57 @@
+@replies
+Feature: Thread Reply support
+
+  Scenario: User replies to message with threading disabled
+
+  Scenario: User replies to uncached message with unthreading
+
+  Scenario: User replies to unthreaded message
+
+  Scenario: User replies to own message
+    Given I have a simple gateway between #foo and #bar
+    And I post a message in #foo
+    And I post a reply in #foo
+    Then a reply should appear in #bar
+
+  Scenario: User replies to relayed message
+    Given I have a simple gateway between #foo and #bar
+    And I post a message in #foo
+    And I post a reply in #bar
+    Then a reply should appear in #foo
+
+  Scenario: User edits a reply message
+    Given I have a simple gateway between #foo and #bar
+    And I post a message in #foo
+    And I post a reply in #bar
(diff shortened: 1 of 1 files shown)


patcon:
If we do feel like using BDD, this tool seems nicest: https://github.com/DATA-DOG/godog

Also, these exist:
- https://github.com/cucumber/gherkin-go
- https://github.com/gucumber/gucumber

Helcaraxan:
Instead of writing scenarios that we then, somehow, somewhere will need to translate into tests via a yet unknown and undesigned testing framework I would really suggest we rather focus first on writing unit-test as suggested in #576 as those should cover already most of the scenarios that you would end up writing.

patcon:
I respectfully, and wholeheartedly disagree. Not everyone who comes to these issue queues will be able to write tests. Not because they're lazy, but because they're not that sort of user. More places to hook people in makes for stronger community-building. This is perfect low-hanging fruit to engage drive-by users.

And personally, I've spent a many hours learning to write my few meager tests. Maybe you can do it more quickly, but sometimes I need to just jot down a new edge-case

Commons-based peer production (the academic word for what open source is) flourishes by creating as many niches for participation as possible. Documentation (incl hard-earned test scenarios) are part of that.

Helcaraxan:
I think I might not have expressed my opinion in the right way so apologies for that @patcon. I actually wholeheartedly agree with your point-of-view on encouraging and building contributions. No discussion there.

The point I was trying to convey was that the usefulness of the resulting scenarios would be very meagre without an actual way / enactable way of putting them to good use. This means that my argument is technical in nature and has nothing to do with either contribution building or the varying degree of knowledge of potential contributors.

I have already given quite some thought to the possibilities of enabling easier & meaningful testing so that other people can contribute. Just like you would like to see as well! However this thinking has not yet been fruitful as all my existing ideas up to now have walked into technical difficulties or other similar roadblocks. That's not to say, on the contrary, that there is no good solution, just that I have not encountered it yet and that I am not _currently_ convinced by your approach either.

I would encourage thinking about this more before even trying to build / contribute anything. We can open up a relevant issue for it to contain the discussion. And I'll be more than happy to take part in it. If we can agree with you, @42wim and me on an approach after the necessary discussion and technical evaluation we can decide on how to move forward and what we need to do / build to facilitate those so much sought after contributions.

From my perspective I would not like, in any situation, to encourage people to contribute to something whilst also knowing that those contributions might very well not be of any use. I hope we can agree on that point at least.

42wim:
I agree with @Helcaraxan and especially this point:

> From my perspective I would not like, in any situation, to encourage people to contribute to something whilst also knowing that those contributions might very well not be of any use. I hope we can agree on that point at least.

Regarding to contribution, I'm not sure how we can encourage this more. 
People can already contribute in the wiki, it's as low-entry as can be and it doesn't happen that often.

42wim:
I don't think there is going to be any further work on this PR?
If so, please reopen, but closing for now.

PR #624, opened 2018-11-30
[Slack] Threading breaks on complex gateways with multiple sub-protocols

patcon:
Reticketed from: https://github.com/42wim/matterbridge/pull/594#issuecomment-442945678

This is the minimum fix, though I do stand by the suggestion that caching under a baseProtocol concept is more sensible for what we're actually trying to do. Happy to submit that fix instead :)

steps to repro:
1. Create gateway between 3 channels: 2 slack protocol channels and 1 slack-legacy channel.
2. Post message in `slack` protocol channel. Reply to that message in same channel. See message get threaded on all bridges.
2. Post reply in the other `slack` protocol channel. These should thread in all bridges as well.
3. Post reply message on `slack-legacy` side. The reply will be correctly threaded in the _original parent channel_ (from step 1), but not on the other channel (with the relayed parent from step 2).


42wim:
Sigh, I don't want that extra complexity for something that probably won't get used.

So maybe it's better to remove `slack-legacy` and call it again `slack`, and use the legacy connect when not using a bot-token. (and also remove the 30 second delay).

patcon:
But yes, I'm not sure I understand the goal of making it a separate protocol name. It seems even if we want the delay to encourage migration, we can just add it in response to the detected token type. (Though I am a +1 on removing that delay.) But again, I'll investigate

patcon:
Confirmed that this brings threading back into working order. Ready for review.

Fwiw, the legacy token warning (as it always has), prints for each bridge that uses the legacy token, rather than for each account. Not sure how to get around that, but open to suggestions.

Lastly, I had this, but didn't include it. Thoughts on removing `DisableServer` on matterhook? Seems that if `BindAddress` is set, it should be enabled, and if not, disabled. Feel free to apply this diff if you agree, but whatever works!

```diff
diff --git a/bridge/mattermost/helpers.go b/bridge/mattermost/helpers.go
index c611762..9730811 100644
--- a/bridge/mattermost/helpers.go
+++ b/bridge/mattermost/helpers.go
@@ -41,8 +41,8 @@ func (b *Bmattermost) doConnectWebhookBind() error {
 func (b *Bmattermost) doConnectWebhookURL() error {
 	b.Log.Info("Connecting using webhookurl (sending)")
 	b.mh = matterhook.New(b.GetString("WebhookURL"),
-		matterhook.Config{InsecureSkipVerify: b.GetBool("SkipTLSVerify"),
-			DisableServer: true})
+		matterhook.Config{InsecureSkipVerify: b.GetBool("SkipTLSVerify")},
+	)
 	if b.GetString("Token") != "" {
 		b.Log.Info("Connecting using token (receiving)")
 		err := b.apiLogin()
diff --git a/bridge/rocketchat/rocketchat.go b/bridge/rocketchat/rocketchat.go
index 1dbc7be..1d1ea87 100644
--- a/bridge/rocketchat/rocketchat.go
+++ b/bridge/rocketchat/rocketchat.go
@@ -25,8 +25,8 @@ func (b *Brocketchat) Command(cmd string) string {
 func (b *Brocketchat) Connect() error {
 	b.Log.Info("Connecting webhooks")
 	b.mh = matterhook.New(b.GetString("WebhookURL"),
-		matterhook.Config{InsecureSkipVerify: b.GetBool("SkipTLSVerify"),
-			DisableServer: true})
+		matterhook.Config{InsecureSkipVerify: b.GetBool("SkipTLSVerify")},
+	)
 	b.rh = rockethook.New(b.GetString("WebhookURL"), rockethook.Config{BindAddress: b.GetString("WebhookBindAddress")})
 	go b.handleRocketHook()
 	return nil
diff --git a/bridge/slack/slack.go b/bridge/slack/slack.go
index 61dcf8f..07595ff 100644
--- a/bridge/slack/slack.go
+++ b/bridge/slack/slack.go
@@ -132,8 +132,7 @@ func (b *Bslack) Connect() error {
 	b.mh = matterhook.New(
 		"",
 		matterhook.Config{
-			InsecureSkipVerify: b.GetBool("SkipTLSVerify"),
-			DisableServer:      true,
+			InsecureSkipVerify: b.GetBool(skipTLSConfig),
 		},
 	)
 	if webhookUrl := b.GetString(outgoingWebhookConfig); webhookUrl != "" {
@@ -143,7 +142,6 @@ func (b *Bslack) Connect() error {
 	if bindAddr := b.GetString(incomingWebhookConfig); bindAddr != "" {
 		b.Log.Info("Setting up local webhook for incoming messages.")
 		b.mh.BindAddress = bindAddr
-		b.mh.DisableServer = false
 		go b.handleSlack()
 	}
 	return nil
diff --git a/matterhook/matterhook.go b/matterhook/matterhook.go
index f602ed4..68c23b6 100644
--- a/matterhook/matterhook.go
+++ b/matterhook/matterhook.go
@@ -64,7 +64,10 @@ type Config struct {
 	BindAddress        string // Address to listen on
 	Token              string // Only allow this token from Mattermost. (Allow everything when empty)
 	InsecureSkipVerify bool   // disable certificate checking
-	DisableServer      bool   // Do not start server for outgoing webhooks from Mattermost.
+}
+
+func (c Config) ServerDisabled() bool {
+	return c.BindAddress == ""
 }
 
 // New Mattermost client.
@@ -74,7 +77,7 @@ func New(url string, config Config) *Client {
 		TLSClientConfig: &tls.Config{InsecureSkipVerify: config.InsecureSkipVerify},
 	}
 	c.httpclient = &http.Client{Transport: tr}
-	if !c.DisableServer {
+	if !c.ServerDisabled() {
 		_, _, err := net.SplitHostPort(c.BindAddress)
 		if err != nil {
 			log.Fatalf("incorrect bindaddress %s", c.BindAddress)
```

patcon:
Also, this is based on https://github.com/42wim/matterbridge/pull/626, which should be merged first.

bridge/slack/slack.go
```diff
@@ -116,8 +109,13 @@ func (b *Bslack) Connect() error {
 	b.RLock()
 	defer b.RUnlock()
 
-	if b.GetString(incomingWebhookConfig) == "" && b.GetString(outgoingWebhookConfig) == "" && b.GetString(tokenConfig) == "" {
-		return errors.New("no connection method found: WebhookBindAddress, WebhookURL or Token need to be configured")
+	if b.GetString(incomingWebhookConfig) == "" &&
+		b.GetString(outgoingWebhookConfig) == "" &&
+		b.GetString(tokenConfig) == "" {
+		return fmt.Errorf("no connection method found: %s, %s or %s need to be configured",
```
patcon: No change, just aligning for legibility.

bridge/slack/slack.go
```diff
@@ -134,17 +132,17 @@ func (b *Bslack) Connect() error {
 	b.mh = matterhook.New(
 		"",
 		matterhook.Config{
-			InsecureSkipVerify: b.GetBool("SkipTLSVerify"),
+			InsecureSkipVerify: b.GetBool(skipTLSConfig),
 			DisableServer:      true,
 		},
 	)
-	if b.GetString(outgoingWebhookConfig) != "" {
+	if webhookURL := b.GetString(outgoingWebhookConfig); webhookURL != "" {
```
patcon: Just seemed more legible as someone reading through for first time

42wim:
No, we can't just remove it completely like in your commit https://github.com/42wim/matterbridge/commit/280d322e854c88f927d298c6d916dc25ed9ae1e2 because:
- people could already be using it
- it's fine for normal use, only edge-cases like yours are having issues with this
- it's possible that some features in the future won't be working anymore for legacy tokens, so we'll need separation (https://api.slack.com/custom-integrations/legacy-tokens) 

> You're reading this because you're looking for info on legacy custom integrations - an outdated way for teams to integrate with Slack. **These integrations lack newer features** and they will be deprecated and possibly removed in the future. We do not recommend their use.

So could you let me know if #626 fixes your edge-case ? if so we can possibly close this PR ?
Maybe follow-up this further in #626 (regarding working/not working)?

Helcaraxan:
I would agree with @42wim here. If #626 does the job we should go for that.

PR #626, opened 2018-12-01
Make slack-legacy change less restrictive

42wim:
Be less pushy and let people still use legacy tokens on the `slack` bridge if they want.


bridge/config/config.go
```diff
@@ -161,20 +161,6 @@ type SameChannelGateway struct {
 }
 
 type BridgeValues struct {
```
Helcaraxan: What is this removal doing here in this PR?

PR #627, opened 2018-12-01
Add support for mattermost threading

patcon:
This is untested, as I don't have access to a mattermost instance.

### DIFF
--- a/bridge/slack/helpers.go
+++ b/bridge/slack/helpers.go
@@ -64,7 +64,6 @@ func (b *Bslack) getChannelBy(lookupKey string, lookupMap map[string]*slack.Chan
 const minimumRefreshInterval = 10 * time.Second
 
 func (b *Bslack) populateUsers() {
-	time.Sleep(time.Second)
 	b.refreshMutex.Lock()
 	if time.Now().Before(b.earliestUserRefresh) || b.refreshInProgress {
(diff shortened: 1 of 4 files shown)


bridge/slack/helpers.go
```diff
@@ -64,6 +64,7 @@ func (b *Bslack) getChannelBy(lookupKey string, lookupMap map[string]*slack.Chan
 const minimumRefreshInterval = 10 * time.Second
 
 func (b *Bslack) populateUsers() {
+	time.Sleep(time.Second)
```
Helcaraxan: What is this doing here? Same question for the `populateChannels()` method below.

patcon: Ah shoot. This shouldn't be here. Thanks! (It's my workaround for the users lookup otherwise becoming empty on large slack teams, for which I haven't had time to track down root cause.)

42wim:
@patcon thanks for the initiative :-) I'll test asap on an instance

42wim:
Almost works. You need to actually set the ParentID though :)

Add `ParentID: message.Post.ParentId,` below:
https://github.com/42wim/matterbridge/blob/a2754f15fc2bc2eede94e2e41712f9f0585f2426/bridge/mattermost/handlers.go#L101-L108

42wim:
@patcon are you going to fix this PR with the mattermost issue or should I open a new one for this?

PR #629, opened 2018-12-01
Refactor handleUploadFile (matrix)

42wim:


bridge/matrix/matrix.go
```diff
@@ -99,11 +99,13 @@ func (b *Bmatrix) Send(msg config.Message) (string, error) {
 	// Upload a file if it exists
 	if msg.Extra != nil {
 		for _, rmsg := range helper.HandleExtra(&msg, b.General) {
-			b.mc.SendText(channel, rmsg.Username+rmsg.Text)
+			if _, err := b.mc.SendText(channel, rmsg.Username+rmsg.Text); err != nil {
+				b.Log.Errorf("sendText failed: %s", err)
+			}
 		}
 		// check if we have files to upload (from slack, telegram or mattermost)
 		if len(msg.Extra["file"]) > 0 {
```
Helcaraxan: You can remove this check. Given that the loop in `handleUploadFiles` already iterates over `msg.Extra["file"]` and will thus automatically skip if there are none.

bridge/matrix/matrix.go
```diff
@@ -257,47 +259,53 @@ func (b *Bmatrix) handleDownloadFile(rmsg *config.Message, content map[string]in
 	return nil
 }
 
-// handleUploadFile handles native upload of files
-func (b *Bmatrix) handleUploadFile(msg *config.Message, channel string) (string, error) {
+// handleUploadFiles handles native upload of files
+func (b *Bmatrix) handleUploadFiles(msg *config.Message, channel string) (string, error) {
 	for _, f := range msg.Extra["file"] {
 		fi := f.(config.FileInfo)
```
Helcaraxan: We should check this type cast, skip (and probably log) if it fails.

bridge/matrix/matrix.go
```diff
@@ -257,47 +259,53 @@ func (b *Bmatrix) handleDownloadFile(rmsg *config.Message, content map[string]in
...
-			}
-			b.Log.Debugf("result: %#v", res)
-		}
+		b.handleUploadFile(msg, channel, fi)
 	}
 	return "", nil
 }
 
+// handleUploadFile handles native upload of files
+func (b *Bmatrix) handleUploadFile(msg *config.Message, channel string, fi config.FileInfo) {
```
Helcaraxan: Let's pass the `fi config.FileInfo` as a pointer instead. 🙂

bridge/matrix/matrix.go
```diff
@@ -257,47 +259,53 @@ func (b *Bmatrix) handleDownloadFile(rmsg *config.Message, content map[string]in
 	return nil
 }
 
-// handleUploadFile handles native upload of files
-func (b *Bmatrix) handleUploadFile(msg *config.Message, channel string) (string, error) {
+// handleUploadFiles handles native upload of files
```
Helcaraxan: Nit: we should make all these comments into real sentences with a period at the end.

PR #632, opened 2018-12-02
Split Discord bridge in multiple files

Helcaraxan:
Mainly just a copy-and-paste. Only real code changes:
- split between channels and members RW mutex.
- better clean-up and maintenance of members map on update
- improved username mention replacement logic with accompanying unit-test

### DIFF
--- a/bridge/discord/handlers.go
+++ b/bridge/discord/handlers.go
@@ -1,0 +1,125 @@
+package bdiscord
+
+import (
+	"github.com/42wim/matterbridge/bridge/config"
+	"github.com/bwmarrin/discordgo"
+)
+
+func (b *Bdiscord) messageDelete(s *discordgo.Session, m *discordgo.MessageDelete) {
+	rmsg := config.Message{Account: b.Account, ID: m.ID, Event: config.EventMsgDelete, Text: config.EventMsgDelete}
+	rmsg.Channel = b.getChannelName(m.ChannelID)
+	if b.useChannelID {
+		rmsg.Channel = "ID:" + m.ChannelID
+	}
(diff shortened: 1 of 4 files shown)


Helcaraxan:
The complexity issue from CodeClimate will obviously solved in subsequent PRs to limit the scope of changes in this one.

bridge/discord/handlers.go
```diff
@@ -0,0 +1,125 @@
...
+	b.Log.Debugf("<= Message is %#v", rmsg)
+	b.Remote <- rmsg
+}
+
+func (b *Bdiscord) memberUpdate(s *discordgo.Session, m *discordgo.GuildMemberUpdate) {
+	if m.Member == nil {
+		b.Log.Warnf("Received member update with no member information: %#v", m)
+	}
+
+	b.membersMutex.Lock()
```
42wim: I'm wondering if we don't have a possible locking issue/race here, we're also locking in `Connect()` where we add the `memberUpdate` handle (I know this issue exists in the original code before your PR too :-)

42wim:
Small question, otherwise LGTM

bridge/discord/handlers.go
Helcaraxan: No. That would not be an issue. We would need two locks (which we do have - `channelsMutex`) for which a single codepath tries to acquire one while holding the other. That could then trigger a deadlock.

In this case the code here simply runs and there is nothing inside the locked range on which the execution can block. The handler is only called if the relevant event is received by the endpoint... however even if that is the case then the handler (which will run on its own goroutine anyway triggered by the dispatcher) will simply wait on the main routine to exit the locked range and then acquire the lock for itself.

PR #633, opened 2018-12-02
Clean-up TravisCI config and add test coverage

Helcaraxan:
Doing the necessary to configure test coverage reporting via CodeClimate. The only thing that is missing here is to set the `CC_TEST_REPORTER_ID` env variable in TravisCI as a secret but it can only be retrieved by the repo admin on CodeClimate (@42wim). You can grab the ID from [here](https://codeclimate.com/repos/5bdb7bf561590b026f000d13/settings/test_reporter) (for which I get a permission denied) and add it to the TravisCI setup [here](https://travis-ci.org/42wim/matterbridge/settings).

Have already verified that it works on my fork of the repo and with my own CodeClimate / TravisCI setup.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: go
 go:
     - 1.11.x
+go_import_path: github.com/42wim/matterbridge
 
 # we have everything vendored
 install: true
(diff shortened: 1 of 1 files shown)


42wim:
Thanks!

PR #635, opened 2018-12-03
Move golangci-lint configuration to file

Helcaraxan:
To allow more fine-grained control over what linting is done and how we can use this nice configuration file for our `golangci-lint`linter. It is mostly based on their example file and keeps a lot of the default settings but that's just so that we are aware of what is available by simply reading the file.


.golangci.yaml
```diff
@@ -0,0 +1,216 @@
...
+    simple: true
+    range-loops: true # Report preallocation suggestions on range loops, true by default
+    for-loops: false # Report preallocation suggestions on for loops, false by default
+  gocritic:
+    # which checks should be enabled; can't be combined with 'disabled-checks';
+    # default are: [appendAssign assignOp caseOrder dupArg dupBranchBody dupCase flagDeref
+    # ifElseChain regexpMust singleCaseSwitch sloppyLen switchTrue typeSwitchVar underef
+    # unlambda unslice rangeValCopy defaultCaseOrder];
+    # all checks list: https://github.com/go-critic/checkers
+    enabled-checks:
```
patcon: Should we instead use `disabled-checks`, so when new ones are added, we don't have to track it, we just start seeing failures?

I mean, I understand that that sort of "unpinning" mentality is not a prod mentality, but seems like it would be for us like a newsfeed for new checks :)

patcon:
I learned a ton by skimming through. Thanks! I like the idea of this in repo :)

.golangci.yaml
```diff
@@ -0,0 +1,216 @@
...
+    algo: cha
+
+    # Inspect exported functions, default is false. Set to true if no external program/library imports your code.
+    # XXX: if you enable this setting, unparam will report a lot of false-positives in text editors:
+    # if it's called for subdir of a project it can't find external interfaces. All text editor integrations
+    # with golangci-lint call it on a directory with the changed file.
+    check-exported: false
+  nakedret:
+    # make an issue if func has more lines of code than this setting and it has naked returns; default is 30
+    max-func-lines: 20
```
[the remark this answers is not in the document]
Helcaraxan: Sure. Did not want to enforce it directly but am more than happy to change this to `0`. 😆

.golangci.yaml
```diff
@@ -0,0 +1,216 @@
...
+  unused:
+    # treat code as a program (not a library) and report unused exported identifiers; default is false.
+    # XXX: if you enable this setting, unused will report a lot of false-positives in text editors:
+    # if it's called for subdir of a project it can't find funcs usages. All text editor integrations
+    # with golangci-lint call it on a directory with the changed file.
+    check-exported: false
+  unparam:
+    # call graph construction algorithm (cha, rta). In general, use cha for libraries,
+    # and rta for programs with main packages. Default is cha.
+    algo: cha
```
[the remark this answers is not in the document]
Helcaraxan: True.

.golangci.yaml
```diff
@@ -0,0 +1,216 @@
...
+  depguard:
+    list-type: blacklist
+    include-go-root: false
+    packages:
+    # List of packages that we would want to blacklist for... reasons.
+  misspell:
+    # Correct spellings using locale preferences for US or UK.
+    # Default is to use a neutral variety of English.
+    # Setting locale to US will correct the British spelling of 'colour' to 'color'.
+    locale: US
```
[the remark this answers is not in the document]
Helcaraxan: 🇺🇸

.golangci.yaml
Helcaraxan: Not IMHO. As you mention it yourself that would go against "prod mentality" or even general best-practices. Config files, etc are there to control your build environment, you don't want random things to change and your linting to fail just because your package manager happened to have upgraded `golangci-lint`.

patcon: Not particularly convinced, but def do as you please.

I guess getting the odd new check in our face right away when _we_ update the ci tool seems dandy to me!
https://github.com/42wim/matterbridge/blob/a8dddf88ca27d60ed282d71f04d085d68a8ef674/.travis.yml#L31

One less tedious bit of list cross-checking in the world? Yes, please! Especially when it's a tool we're already expecting to say "Hey, look at this thing you weren't expecting, doofus!" :)

patcon: Oh, hey, speaking of which: Let's pull GOLANGCI_VERSION into global envvar hash

Helcaraxan: Done for the `GOLANGCI_VERSION`.

42wim:
The current config makes it hard to run locally with different settings.

If I want to test with `errcheck` on, it seems I have to edit the file because
```
$ golangci-lint run -E errcheck
ERRO Running error: can't combine options --enable-all and --enable errcheck
```

``` 
$ golangci-lint run --disable-all -E errcheck
ERRO Running error: --enable-all and --disable-all options must not be combined
```

Also the irc `handlers.go` change doesn't need to be in this commit ?

Helcaraxan:
The `handlers.go` was necessary as I got a shadowing linting error when running this initially. Might not be the case anymore but still. Think it is sufficiently minor to group it together.

For the `errcheck`: yes you would need to modify the file. However if you are running `errcheck` then you are very likely working on enabling it anyway so that edit would come in normally anyway as part of your PR (heads-up: I'll probably make that one of my next PRs, i.e. enabling `errcheck`, have already one for `gosec` ready).

Also please bear in mind that because of the `new-since-rev: "~HEAD"` config you are only analysing things that have changed in the last commit. That means that we actually do incremental linting which means that we could potentially enable `errcheck` (for example) without necessarily having to fix all currently offending locations.

Helcaraxan:
Actually... I could theoretically add `errcheck` right now and it would pass CI (just ran did so locally), same for `gosec`.

42wim:
Thanks guys!

PR #637, opened 2018-12-03
Add lightweight Swagger API docs

patcon:
- Uses external render UI at https://petstore.swagger.io
- Added some redirects some unused endpoints help people discover docs.
- Removes token auth on doc file and redirects.

Open questions:
1. should we leave any docstrings in from \#551, to help updating the yaml later? (And to coax future contributors into keeping the docs up-to-date should that make any changes.)


bridge/api/api.go
```diff
@@ -36,9 +38,31 @@ func New(cfg *bridge.Config) bridge.Bridger {
 	if b.GetInt("Buffer") != 0 {
 		b.Messages.SetCapacity(b.GetInt("Buffer"))
 	}
+
+	// Set up Swagger API docs and helpful redirects.
+	e.Pre(middleware.RemoveTrailingSlash())
+	e.Use(middleware.CORSWithConfig(middleware.CORSConfig{
+		AllowOrigins: []string{"https://petstore.swagger.io", "http://petstore.swagger.io"},
+	}))
+	e.File("/swagger/docs.yaml", "docs/swagger/swagger.yaml")
```
patcon: Move to `bridge/api/swagger.yaml`?

42wim:
Thanks for the work, having this documentation will be useful for newcomers!
But I would make it even more lightweight as in no changes in `api.go`.

People can read the swagger spec via the direct link in the README, or they can download the swagger app locally.

patcon:
I will admit to feeling this not very considerate of users, but I'm happy to have some variation of better docs. (ie. making people retrace steps and copy-paste from a doc, instead of taking them to the answer that we _know_ they're trying to find)

42wim:
> not as considerate of user experience 

I'm all for "the path of least resistance".
Isn't clicking a link in the README to see the spec as least resistance as possible ? :-) 
Otherwise they need to compile/download matterbridge, configure the API and surf to the webpage, which are a lot more steps.

42wim:
@patcon https://swagger.io/tools/swaggerhub/ is maybe also interesting

42wim:
I think we can close this with commits https://github.com/42wim/matterbridge/commit/85680935d49ba2c96508f22d1b66075c5c35f64e and https://github.com/42wim/matterbridge/commit/0d94746f4afe2b2e6ca7628c222b6696fc6b4cde

patcon:
Works for me. Note for posterity that this won't be able to be explorable via the UI as the more complex alternative was, and it won't let devs explore a deployed instance, as it doesn't actually talk to the API (?), but def feels like a big improvement and I'm grateful!

PR #644, opened 2018-12-04
[slack] populate users on member channel join

dajohi:


patcon:
Thanks @dajohi! Great find.

I don't know this section of codebase so well, but wondering if maybe we could just insert the user in question from the payload so that it's less intensive, like here:

https://github.com/42wim/matterbridge/blob/f5659d455d2c28a6f2fe4c40f4dc344419ff523a/bridge/slack/handlers.go#L80-L83

dajohi:
@patcon request for your second review :)

bridge/slack/helpers.go
```diff
@@ -63,6 +63,18 @@ func (b *Bslack) getChannelBy(lookupKey string, lookupMap map[string]*slack.Chan
 
 const minimumRefreshInterval = 10 * time.Second
 
+func (b *Bslack) populateUser(userID string) {
+	user, err := b.sc.GetUserInfo(userID)
+	if err != nil {
```
Helcaraxan: Only thing that I would add here is handling rate-limiting correctly.

You can see the typical way of doing this in the code of the `populateUsers()` method below.

aslafy-z:
Any chance you finish this (quite critical) PR @dajohi? Thanks!

bridge/slack/helpers.go
dajohi: I decided to check the cache instead of implementing rate-limiting code.  This is much simpler.

42wim:
LGTM, thanks for your work @dajohi !

PR #646, opened 2018-12-05
Enable errcheck linter

Helcaraxan:
Some fixes that allows to enable the `errcheck` linter.

Also drive-by fixes for two issues that were flagged by the `unparam` linter.


bridge/slack/slack.go
```diff
@@ -195,16 +195,16 @@ func (b *Bslack) Send(msg config.Message) (string, error) {
...
 	// Use webhook to send the message
 	if b.GetString(outgoingWebhookConfig) != "" {
-		return b.sendWebhook(msg)
+		return "", b.sendWebhook(msg)
 	}
 	return b.sendRTM(msg)
 }
 
 // sendWebhook uses the configured WebhookURL to send the message
-func (b *Bslack) sendWebhook(msg config.Message) (string, error) {
```
Helcaraxan: Removing the returned `string` as suggested by the `unparam` linter. It's value is always `""`.

bridge/slack/slack.go
```diff
@@ -317,7 +317,7 @@ func (b *Bslack) sendRTM(msg config.Message) (string, error) {
 	return b.postMessage(&msg, channelInfo)
 }
 
-func (b *Bslack) updateTopicOrPurpose(msg *config.Message, channelInfo *slack.Channel) (bool, error) {
```
Helcaraxan: Removing the returned `bool` as suggested by the `unparam` linter. It's value is always `true`.

bridge/sshchat/sshchat.go
```diff
@@ -31,8 +31,7 @@ func (b *Bsshchat) Connect() error {
 			b.w = w
 			b.r.Scan()
 			w.Write([]byte("/theme mono\r\n"))
-			b.handleSSHChat()
-			return nil
+			return b.handleSSHChat()
```
Helcaraxan: I will revisit this function in a wider manner in a subsequent PR as the synchronisation logic in it is broken.

matterclient/matterclient.go
```diff
@@ -208,7 +208,10 @@ func (m *MMClient) StatusLoop() {
 			return
 		}
 		if m.WsConnected {
-			m.checkAlive()
+			if err := m.checkAlive(); err != nil {
+				log.Errorf("Connection is not alive: %#v", err)
+				break
```
42wim: The break should be removed here, we don't want to escape the for loop here.

PR #648, opened 2018-12-08
Refactor gateway

42wim:
* Decrease complexity of handleMessage, handleReceive, handleFiles
* Move handlers to handlers.go
* Split ignoreMessage up in ignoreTextEmpty, ignoreNicks and IgnoreTexts
* Add ignoreEvent
* Add testcase for ignoreTextEmpty, ignoreNicks, ignoreTexts and ignoreEvent

### DIFF
--- a/gateway/router.go
+++ b/gateway/router.go
@@ -108,41 +108,23 @@ func (r *Router) getBridge(account string) *bridge.Bridge {
 func (r *Router) handleReceive() {
 	for msg := range r.Message {
 		msg := msg // scopelint
-		if msg.Event == config.EventFailure {
-		Loop:
-			for _, gw := range r.Gateways {
-				for _, br := range gw.Bridges {
-					if msg.Account == br.Account {
-						go gw.reconnectBridge(br)
-						break Loop
-					}
-				}
-			}
-		}
-		if msg.Event == config.EventRejoinChannels {
-			for _, gw := range r.Gateways {
-				for _, br := range gw.Bridges {
-					if msg.Account == br.Account {
-						br.Joined = make(map[string]bool)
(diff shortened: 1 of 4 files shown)


gateway/router.go
```diff
@@ -108,41 +108,23 @@ func (r *Router) getBridge(account string) *bridge.Bridge {
...
-				gw.modifyMessage(&msg)
-				gw.handleFiles(&msg)
-				for _, br := range gw.Bridges {
-					msgIDs = append(msgIDs, gw.handleMessage(msg, br)...)
-				}
-				// only add the message ID if it doesn't already exists
-				if _, ok := gw.Messages.Get(msg.Protocol + " " + msg.ID); !ok && msg.ID != "" {
-					gw.Messages.Add(msg.Protocol+" "+msg.ID, msgIDs)
-				}
+			if gw.ignoreMessage(&msg) {
```
patcon: If we're thinking about a pattern that we can lean on more often, and in more places (which admittedly we might not be doing), I personally prefer the condition checks exiting near the top, and not nesting. It also offers closure while reading code, because we can discard that context, as we're not keep context in our head in anticipation of an "else". For example, this would seem to support more logical breakdown of the exit rationale:

```go
if gw.conditionOne($msg) {
  continue
}

if gw.conditionTwoA(&msg) || gw.conditionTwoB(&msg) {
  continue
}
```

Any thoughts on this as a general strategy in the codebase? (This is more than just this nit, I think, and prob something we could doc if we had a convention :) )

42wim: But in this particular piece of code it is already near the top ? I don't understand the issue here ?

PR #653, opened 2018-12-11
Add wait option for populateUsers/Channels. Fixes #579

### LINKED ISSUE #579: [Slack] users list not populated under high API usage
**Describe the bug**
When deploying the amazing new refactor on Heroku, no messages were sending. I've debugged it to find that the `b.users` slice is empty, so I get this in the logs with my debug output (see full log):

    &errors.errorString{s:"could not find information for user with id U4MCXJKNC"}

Having said that, about half the time running locally (where my connection is slower), the b.users populated with 4000+ users (I'm also on a high-usercount slack)

**To Reproduce**
Add about 11 bridges over 2 gateways on a no too slow connection, with a large slack team of 4200 users.

**Expected behavior**
Would expect b.users to be populated and messages to send.

**Screenshots/debug logs**
Logs: https://gist.github.com/patcon/b45c1e80cfd249eb54ce6f877a62e9ea

 - OS: osx
- working on the head of master

**Additional context**
Added some debug code to make use of the RateLimitEvent from slack's rtm api, and not seeing it raise a flag when the b.users var fails to populate, so I'm thinking it must be something else

I'm wondering whether it might be related to the giant team and not being able to use the exclude_members flag which slack's api docs warn are heavy on big teams: https://github.com/nlopes/slack/pull/169

Might be worth reviving that PR in the scope of this, if we run out of options

cc @Helcaraxan if you have any thoughts :)

42wim:
When setting wait to true, we wait until the populating isn't in progress anymore.
This is used on startup connections where we really need the initial information
which could take a long time on big servers.

### DIFF
--- a/bridge/slack/handlers.go
+++ b/bridge/slack/handlers.go
@@ -75,7 +75,7 @@ func (b *Bslack) handleSlackClient(messages chan *config.Message) {
 			// When we join a channel we update the full list of users as
 			// well as the information for the channel that we joined as this
 			// should now tell that we are a member of it.
-			b.populateUsers()
+			b.populateUsers(false)
 
 			b.channelsMutex.Lock()
 			b.channelsByID[ev.Channel.ID] = &ev.Channel
@@ -83,8 +83,8 @@ func (b *Bslack) handleSlackClient(messages chan *config.Message) {
 			b.channelsMutex.Unlock()
 		case *slack.ConnectedEvent:
 			b.si = ev.Info
-			b.populateChannels()
-			b.populateUsers()
+			b.populateChannels(true)
+			b.populateUsers(true)
 		case *slack.InvalidAuthEvent:
 			b.Log.Fatalf("Invalid Token %#v", ev)
 		case *slack.ConnectionErrorEvent:
@@ -200,7 +200,7 @@ func (b *Bslack) handleMessageEvent(ev *slack.MessageEvent) (*config.Message, er
 func (b *Bslack) handleStatusEvent(ev *slack.MessageEvent, rmsg *config.Message) bool {
 	switch ev.SubType {
 	case sChannelJoined, sMemberJoined:
-		b.populateUsers()
+		b.populateUsers(false)
 		// There's no further processing needed on channel events
 		// so we return 'true'.
 		return true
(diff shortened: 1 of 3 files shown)


42wim:
@patcon could you try this PR for your issue in #579 ?

bridge/slack/helpers.go
```diff
@@ -63,15 +63,18 @@ func (b *Bslack) getChannelBy(lookupKey string, lookupMap map[string]*slack.Chan
...
-	if time.Now().Before(b.earliestUserRefresh) || b.refreshInProgress {
+	if !wait && (time.Now().Before(b.earliestUserRefresh) || b.refreshInProgress) {
 		b.Log.Debugf("Not refreshing user list as it was done less than %v ago.",
 			minimumRefreshInterval)
 		b.refreshMutex.Unlock()
 
 		return
 	}
+	for b.refreshInProgress {
+		time.Sleep(time.Second)
```
Helcaraxan: You can actually determine the right duration based on current time and `b.earliestUserRefresh`.

Also you will likely want to rewrite the logic somewhat here as you'll need to release the lock while you are sleeping. And once you are finished you'll need to recheck on whether you have passed the `earliestUserRefresh` time (as there's always the possibility of a race-condition.

bridge/slack/handlers.go
```diff
@@ -197,15 +197,15 @@ func (b *Bslack) handleMessageEvent(ev *slack.MessageEvent) (*config.Message, er
...
+		b.populateUsers(false)
 		// There's no further processing needed on channel events
 		// so we return 'true'.
 		return true
 	case sChannelJoin, sChannelLeave:
 		rmsg.Username = sSystemUser
 		rmsg.Event = config.EventJoinLeave
 	case sChannelTopic, sChannelPurpose:
-		b.populateChannels()
+		b.populateChannels(false)
```
Helcaraxan: Maybe not as part of this PR but I can't seem to remember why we are calling populateChannels here and don't just register the new info from the event in the corresponding channel.

bridge/slack/helpers.go
42wim: Yes, but I want to do it earlier then the `b.earliestUserRefresh` that's why I don't use it here :-)
Why should I release the lock while sleeping? It may block here, as it'll only happen on startup where we want the blocking.

bridge/slack/handlers.go
42wim: it's added in f5659d45
let's keep this for another PR yes

patcon:
Yeah! Sorry, finishing my fellowship, and so a little distracted ;)

And yeah, this resolves the issue! This fix consistently resolve the intermittent issue, and reverting it bring back the observed error and failures. Thanks @42wim 🎉 🎉

bridge/slack/helpers.go
42wim: I'm going to merge this now so I can release a v1.12.3 which has a couple of slack bugfixes. We can rewrite this later in a safer way.

PR #656, opened 2018-12-17
Slack download issue

### LINKED ISSUE #655: Slack platform sends message with file before file is ready for download
**Description**
When matterbridge is configured with a MediaServer and it receives an event from the Slack platform that indicates a user uploaded a file (File object embedded in a MessageEvent), then the download can fail if matterbridge requests the file too quickly. This results in an HTML page getting downloaded instead (a redirect to the main login page on slack.com).

**To Reproduce**
This is incredibly hard to reproduce. It requires a request be made in well under a second.  

**Expected behavior**
The file should be downloaded properly so that it can be re-hosted/uploaded to the destination platform.

**Screenshots/debug logs**
None

 - OS: linux amd64
 - Matterbridge version: 1.12.3-dev

**Additional context**
I have completed a support request with Slack and they say that this is expected behavior. They have no plans to adjust the behavior of the platform. I have filed a formal request to have a new event type emitted (i.e. file_ready), but they aren't committing to adding it. Their only suggestion is to add a delay to the download routine.

I have written an update for the handleDownloadFile that attempts a single retry if the downloaded file size doesn't match what was originally declared in the File object on the MessageEvent and have been using it in my own environment for the last week. I will be submitting a PR to go along with this issue.

jheiselman:
Fixes #655


42wim:
@jheiselman thanks! can you fix the travis failing checks ?

42wim:
it's basically removing the two int() cast that are not necessary because it's already an int

bridge/slack/handlers.go
```diff
@@ -292,11 +292,18 @@ func (b *Bslack) handleDownloadFile(rmsg *config.Message, file *slack.File) erro
 		return fmt.Errorf("download %s failed %#v", file.URLPrivateDownload, err)
 	}
 
+	if len(*data) != int(file.Size) && retry != true {
```
patcon: Just pulling from travis :)

```
bridge/slack/handlers.go:295:22: unnecessary conversion (unconvert)
	if len(*data) != int(file.Size) && retry != true {

```

bridge/slack/handlers.go
```diff
@@ -292,11 +292,18 @@ func (b *Bslack) handleDownloadFile(rmsg *config.Message, file *slack.File) erro
...
+		b.Log.Debugf("Data size (%i) is not equal to size declared (%i)\n", len(*data), int(file.Size))
```
patcon:
```
bridge/slack/handlers.go:296:86: unnecessary conversion (unconvert)
		b.Log.Debugf("Data size (%i) is not equal to size declared (%i)\n", len(*data), int(file.Size))
```

42wim:
Thanks! 👍

PR #661, opened 2018-12-26
SSHChat - fix connection logic and remove send bug

Helcaraxan:
The existing logic for establishing connections in the SSHChat bridge contained potential race-conditions as it did not wait for the connection to be successful before returning.

Additionally we reintroduce a missing `if` condition on handling file attachments that was previously wrongly removed.


42wim:
Thanks! Look into this later.
I'm curious why codeclimate doesn't trigger anymore, didn't change anything. Maybe also on holiday :)

42wim:
Thanks!

PR #662, opened 2018-12-26
Logging rewiring - Import linting

Helcaraxan:
First of two PRs:

1. (This one) - Linting of imports and removal of import-renaming for `github.com/sirupsen/logrus` to remove potential confusion with native `log` package when reading code.

2. Rewiring of `logrus.Logger` instances throughout the code-base to perform a single instantiation within at the root. This will make it easier to forward logs to non-default destinations (ELK, etc).

### DIFF
--- a/bridge/bridge.go
+++ b/bridge/bridge.go
@@ -1,10 +1,10 @@
 package bridge
 
 import (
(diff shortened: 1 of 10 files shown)


42wim:
👍

PR #670, opened 2019-01-02
Add support for markdown to HTML conversion (matrix). Closes #663

### LINKED ISSUE #663: [Riot/matrix] Support for markdown by supporting Riot's formatted_body tag
When a message containing markdown is sent to Riot from anothe client (tested with rocket chat) - the markdown is not parsed correctly, as it renders the literal input like in the following example:

riot-rocket

I posted this as a question in #matterbridge:matrix.org - I am out of my depth here, but wim answered the following:
<w​im> 
i think riot replaces clientside markdown input to commonmark, haven't looked into this.
```
{"body":"*def*", "format":"org.matrix.custom.html", "formatted_body":"<em>def</em>", "msgtype":"m.text"}
```

I considered other bridging solutions, but there does not seem to be anything ready for deployment to bridge rocket chat and Riot.

**Additional context**
I love matterbridge and the versatility it brings to bridging chat protocols. I am heavily advocating it and use it to bridge some rooms of our communities +giveth:matrix.org and +aragon:matrix.org between Telegram/Riot/rocket chat and in the past Slack.

42wim:
This uses our own gomatrix lib with the SendHTML function which adds HTML to formatted_body in matrix.
golang-connmark is used to convert markdown into valid HTML.


patcon:
Cool! golang-commonmark also seems to be even more easily extensible than blackfriday, and so able to be overridden/customized for slackdown (slack's custom markdown variant):

https://github.com/golang-commonmark/markdown/blob/a8f139058164d82e67bb370994c655d8adc8f9e7/plugins.go#L94-L107

PR #681, opened 2019-01-07
Discord: Only require one webhook

### LINKED ISSUE #674: Discord can use just one webhook
There is a limit to the number of webhooks you can create on Discord. And creating webhooks is painful manual work.

1. The bridge should not need a webhook per channel, it should just edit webhooks
2. The bridge should create/delete webhooks automatically when it starts/stops

Only enabling the edit feature (you still have to manually create _one_ webhook), and this still means that only one webhook is needed per guild, which is awesome.

**Additional context**
I currently do this in my bot. When the bot starts, it deletes all webhooks that were previously created by the bot (in case the bot crashed). It uses a "prefix" to determine which webhook(s) are ours. This is not necessary if auto creation/deletion is not implemented (as per "alternatives" above).

discordgo supports editing of webhooks as of 0.19.0 (this is the version that matterbridge currently uses)

The reason auto-creation/deletion is to maintain "conversations" across channels. At the time of implementation (at least 1 year ago), editing a webhook prevented "stacking" of messages by the same user; i.e, in the following scenario, Bob's avatar would appear twice, instead of once.

1. Bob speaks in #general (webhook edited to be Bob, if not already Bob)
2. Alex speaks in #random (webhook edited to be Alex)
1. Bob speaks in #general (webhook edited to be Bob, and Bob's avatar appears twice, instead of once)

This will be re-tested. If this bug no longer occurs, sticking to one webhook will work. Otherwise, multiple webhooks are required to enable stacking of avatars.

**Example implementation**

See example source: https://github.com/qaisjp/go-discord-irc/tree/master/transmitter.

This code has a bug where if something happens Discord-side, webhook state will become out of sync, and the bridge will send messages to the wrong channel. This is because the bridge thinks the webhook is on another channel; a simple fix for this is to check the channel every message send, but this introduces some latency.

Any PR submitted that fixes this issue should not introduce this bug.

I'm happy to work on this!

nylen:
Fixes #674.

Quoting from https://github.com/42wim/matterbridge/issues/674#issuecomment-451840010:

> On the matterbridge side maybe add a `AutoWebhook` option (not sure yet if this should be enabled by default)

I don't think this is necessary.  Instead, the PR preserves existing [now-documented](https://github.com/42wim/matterbridge/wiki/Section-Discord-(basic)#usernameavatar-spoofing) functionality, and takes advantage of the account-level `WebhookURL` setting that previously was not working as intended:

- If a `WebhookURL` is specified for the account connection (in the `[discord.servername]` section along with the API token), then it is treated as a **global** webhook and the channel is modified each time it is used.  (Currently, a webhook URL specified in this section will always send to the single channel it is configured for, regardless of gateway settings.  This doesn't seem like desirable behavior.)

- If a `WebhookURL` is specified in the gateway settings, then it is treated as a single-channel webhook and the channel is not edited.  This is the same behavior as before this PR.

> At the time of implementation (at least 1 year ago), editing a webhook prevented "stacking" of messages by the same user; i.e, in the following scenario, Bob's avatar would appear twice, instead of once.
>
>1. Bob speaks in #general (webhook edited to be Bob, if not already Bob)
>2. Alex speaks in #random (webhook edited to be Alex)
>3. Bob speaks in #general (webhook edited to be Bob, and Bob's avatar appears twice, instead of once)

This still needs to be tested, but I don't think it's a big issue.

If something like this PR is merged, and a new release is made, I'll update the [Discord documentation](https://github.com/42wim/matterbridge/wiki/Section-Discord-(basic)#usernameavatar-spoofing) accordingly.

### DIFF
--- a/bridge/discord/discord.go
+++ b/bridge/discord/discord.go
@@ -152,7 +152,8 @@ func (b *Bdiscord) Send(msg config.Message) (string, error) {
 		msg.Text = "_" + msg.Text + "_"
 	}
 
-	// use initial webhook
+	// use initial webhook configured for the entire Discord account
+	isGlobalWebhook := true
 	wID := b.webhookID
 	wToken := b.webhookToken
 
@@ -161,6 +162,7 @@ func (b *Bdiscord) Send(msg config.Message) (string, error) {
 	if ci, ok := b.channelInfoMap[msg.Channel+b.Account]; ok {
 		if ci.Options.WebhookURL != "" {
 			wID, wToken = b.splitURL(ci.Options.WebhookURL)
+			isGlobalWebhook = false
 		}
 	}
 	b.channelsMutex.RUnlock()
@@ -188,6 +190,18 @@ func (b *Bdiscord) Send(msg config.Message) (string, error) {
 		// discord username must be [0..32] max
 		if len(msg.Username) > 32 {
 			msg.Username = msg.Username[0:32]
+		}
+		// if we have a global webhook for this Discord account, use it, but
+		// first set its channel to the message channel
+		// TODO: this isn't necessary if the last message from this webhook was
+		// sent to the current channel
+		if isGlobalWebhook {
+			b.Log.Debugf("Setting webhook channel to \"%s\"", msg.Channel)
+			_, err := b.c.WebhookEdit(wID, "", "", channelID)
+			if err != nil {
+				b.Log.Errorf("Could not set webhook channel: %v", err)
+				return "", err
+			}
 		}
 		err := b.c.WebhookExecute(
 			wID,


bridge/discord/discord.go
```diff
@@ -189,6 +191,18 @@ func (b *Bdiscord) Send(msg config.Message) (string, error) {
 		if len(msg.Username) > 32 {
 			msg.Username = msg.Username[0:32]
 		}
+		// if we have a global webhook for this Discord account, use it, but
+		// first set its channel to the message channel
+		// TODO: this isn't necessary if the last message from this webhook was
```
patcon: Any thoughts on how you'd implement this check? I feel like it can't be as simple as checking `existingWebhook.ChannelID` via `existingWebhook, err := b.c.Webhook(wID)`, what with multiple gateways possibly fighting over the one webhook...

I'm open to this being out-of-scope for this PR, but curious if a need for locks might change this code :)

(I am still a beginner with golang)

42wim:
Tested it and it works, the only issue is that it breaks current setups which have only one webhook and do not have the "manage webhooks" permission.

I suggest to:
* add a `useGlobalWebhook` bool to the [bridge struct](https://github.com/42wim/matterbridge/blob/master/bridge/discord/discord.go#L18)
* check for the required permission with [State.UserChannelPermissions](https://godoc.org/github.com/bwmarrin/discordgo#State.UserChannelPermissions) in the `Connect` function (https://github.com/42wim/matterbridge/blob/master/bridge/discord/discord.go#L50) 
* set `useGlobalWebhook` to `true` if the permissions aren't there.

This way legacy setups keep working.

qaisjp:
I don't think this PR fixes #674 entirely as I would still like to implement auto-creation, for convenience sake.

nylen:
> the only issue is that it breaks current setups which have only one webhook and do not have the "manage webhooks" permission

Happy to make these changes, but aren't these setups already kind of broken?  They'll always post to the same channel (the one associated with the webhook) regardless of how the channels are configured in the gateway.

> I don't think this PR fixes #674 entirely as I would still like to implement auto-creation, for convenience sake.

:+1: I've removed the "fixes" text from the PR description so that the issue will stay open.

bridge/discord/discord.go
nylen: I'm open to suggestions here too.

> checking `existingWebhook.ChannelID` via `existingWebhook, err := b.c.Webhook(wID)`

This + appropriate locking (if needed?) would work.  Possibly even better (but slightly more risky) would be a mechanism to remember the last channel that the webhook used, which would avoid the extra API call to check the webhook.

Since this isn't trivial (to me at least), and it might end up causing _more_ total API calls rather than _less_, I'm inclined to leave it for a separate PR.

qaisjp:
> Happy to make these changes, but aren't these setups already kind of broken? They'll always post to the same channel (the one associated with the webhook) regardless of how the channels are configured in the gateway.

I agree! At the moment this currently actually prevents me from (wanting to go ahead with) setting up up a discord-slack gateway for two channels on the same guild.

Maybe would be good to have a "transition" period though? To warn users that this will change in x version(s)? What is the transition period for legacy-slack (or is it just on Slack's timescale of whenever they phase out the legacy system?)

42wim:
> Happy to make these changes, but aren't these setups already kind of broken? They'll always post to the same channel (the one associated with the webhook) regardless of how the channels are configured in the gateway.

Well, if people only have one discord channel (probably the usecase for a lot af people) it works fine, so I'd like to keep this working :-)

nylen:
@42wim I've updated the PR according to your feedback in https://github.com/42wim/matterbridge/pull/681#issuecomment-452095615.

> add a `useGlobalWebhook` bool to the bridge struct

I decided to call this `canEditWebhooks` instead, which I think is more accurate.  Regardless of whether the permission is present, the global webhook will still be used.

> check for the required permission with `State.UserChannelPermissions` in the `Connect` function

Since permissions are channel-specific, I think it makes sense to check the permission for every channel that we find upon joining the guild, but let me know if you had a different approach in mind.

Issue #614, opened 2018-11-28
Allow some form of robust templating


patcon:
Handlebar is a templating library that allows simple logic. It might be nice to use it to allow for more robust templating for certain types of messages/strings. 

One benefit of this is that comprehensive format rules could be set at `General` level, rather than setting them at each gateway.

Places to implement:

- `RemoteNickFormat`. Could allow for setting certain format based on protocol or. 
- messages that are unthreaded, re: https://github.com/42wim/matterbridge/pull/557
- message edits that can't be in-place, re: https://github.com/42wim/matterbridge/issues/588#issuecomment-441417020
- translated messages, re: https://github.com/42wim/matterbridge/pull/512
- `InitialComment` template, re: https://github.com/42wim/matterbridge/pull/606#discussion_r236061032

Seems the best library in go: https://github.com/aymerick/raymond

patcon:
We could perhaps also use custom helpers to do away with config like "StripNick", as people could do that via templates: https://github.com/aymerick/raymond#helpers

patcon:
Potential use-cases:
- Allow special avatar for system user.
- Allow messages to have quick-access permalink to original message if they're on the same team, for example, during translation.
- Append prefix if parent message is unthreaded.

patcon:
Defaults here would be helpful, so might be nice to have a more robust default process, rather than simply empty strings or falsey values

patcon:
Not sure whether mustache is the most appropriate engine, so generalizing this issue

qaisjp:
Go has a text templating system in its standard library: [text/template](https://golang.org/pkg/text/template/).

patcon:
Thanks @qaisjp! `text/template`'s looping didn't feel very friendly to the vast number of people more familiar with non-golang template engines -- and feels like templates would be very user-facing :)

42wim:
I think using tengo is the solution for this instead of templating, much more powerful

patcon:
Good call. Closing out :) Thanks!

PR #682, opened 2019-01-09
Discord: Mark edited webhook messages with a pencil icon

nylen:
When using a webhook to send messages to Discord with [avatar spoofing](https://github.com/42wim/matterbridge/wiki/Section-Discord-(basic)#usernameavatar-spoofing), edited messages are posted twice.  This is because Discord doesn't allow editing a message sent by a webhook.

We can't fix that, but we can make a cosmetic improvement to make it clearer that the second copy of the message is actually an edit of the previous message:

| In (Slack) | Out (Discord) |
|--|--|
| 2019-01-08t23 02 26-05 00 | 2019-01-08t23 02 44-05 00 |

https://github.com/bwmarrin/discordgo/issues/622 will allow further improvements, for example:

- Remove the hack of using `-webhook-` as a placeholder for the ID of the original message on Discord
- Add a :memo: reaction to the original message to make it clearer that it has been edited

### DIFF
--- a/bridge/discord/discord.go
+++ b/bridge/discord/discord.go
@@ -13,7 +13,10 @@ import (
 	"github.com/bwmarrin/discordgo"
 )
 
-const MessageLength = 1950
+const (
+	MessageLength               = 1950
+	webhookPlaceholderMessageID = "-webhook-message-"
+)
 
 type Bdiscord struct {
 	*bridge.Config
@@ -203,7 +206,7 @@ func (b *Bdiscord) Send(msg config.Message) (string, error) {
 			return "", nil
 		}
 
-		if msg.ID == "-webhook-" {
+		if msg.ID == webhookPlaceholderMessageID {
 			// Received an edit for a message that was previously sent via a
 			// webhook.  We can't edit the original message, but we can mark
 			// the new version with a pencil icon.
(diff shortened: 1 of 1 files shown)


patcon:
Related (as the edit format might not suit all bridge platforms or use-case): https://github.com/42wim/matterbridge/issues/614

42wim:
Good idea!
Seems like I haven't implemented the `EditSuffix` for the discord webhook.
The rest of my comment is the same as in #683 
We could make a `EditPrefix` instead of hardcoding this.

nylen:
> Seems like I haven't implemented the `EditSuffix` for the discord webhook.

Cool, I'll take a look at this.  I didn't know there was an `EditSuffix` setting.

> We could make a `EditPrefix` instead of hardcoding this.

I'd also be fine with a suffix here actually.  Maybe we should start with that, and see if we still want a prefix in another PR?

42wim:
@nylen would be great if you would make a PR to have `EditSuffix` fixed for webhooks too!
If the suffix works and is enough, we don't need the prefix one.

patcon:
just to clarify, the prefix-suffix decision is a moot point if we instead lean on templating for message types :)

`EditSuffix` can be deprecated in favour of smthng like `EditFormat`, with:
- default: `EditFormat="{{ content }}"`
- suffix via `EditFormat="{{content}} (edited)"`
- prefix via `EditFormat="[:pencil:] {{content}}"`
- etc etc

42wim:
@nylen did you test it with the `EditSuffix` ? if so can we close this PR? and the other about slack ?

nylen:
Closing.  I haven't had time to revisit this but it is probably covered by existing functionality.

PR #683, opened 2019-01-09
Slack: Mark edited messages with a pencil icon

nylen:
When sending messages to Slack, edited messages are updated silently without the `(edited)` text that appears when a regular user edits a message.

We can't fix that, but we can make a cosmetic improvement to make it clearer that the message was edited:

| In (Discord) | Out (Slack) |
|--|--|
| 2019-01-08t23 08 21-05 00 | 2019-01-08t23 08 06-05 00 |

### DIFF
--- a/bridge/slack/slack.go
+++ b/bridge/slack/slack.go
@@ -390,6 +390,7 @@ func (b *Bslack) editMessage(msg *config.Message, channelInfo *slack.Channel) (b
 		return false, nil
 	}
 	messageOptions := b.prepareMessageOptions(msg)
+	msg.Text = "[:pencil:] " + msg.Text
 	for {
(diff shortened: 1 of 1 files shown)


patcon:
Related: https://github.com/42wim/matterbridge/issues/614

nylen:
Same as in #682...

> There already is an `EditSuffix` option which allows you to set a "(edited)" message of your choice. (you could set it there)

I didn't see this, maybe it's in a different part of the code.  I think this will work fine...

> I think it's better to add a `EditPrefix` option if you/other people really need it prefixed.

No strong feelings about this.  Probably not worth adding another setting, especially if the suffix already works.

I'll test the suffix and either fix it in this PR, or close this PR.

nylen:
Closing, I haven't had time to revisit this, and as discussed above it may already be covered by existing functionality.

PR #685, opened 2019-01-09
matrix: fix displaying usernames for plain text clients.

dajohi:


42wim:
Oops, good catch! thanks!

PR #689, opened 2019-01-13
Allow clients to bridge non-bot Discord users

zomboy-alfrir:
Discord's API allows us to connect using user tokens instead of bot ones. To do this, we simply don't prefix "Bot " to the token when creating a socket.

This change creates a new config value that prevents this from happening automatically. It will not change any existing behavior for current users, as `b.GetBool("NotBot")` will return false if the config entry is not there.

### DIFF
--- a/bridge/discord/discord.go
+++ b/bridge/discord/discord.go
@@ -58,8 +58,10 @@ func (b *Bdiscord) Connect() error {
 	} else {
 		b.Log.Info("Connecting using webhookurl (for posting) and token")
 	}
-	if !strings.HasPrefix(b.GetString("Token"), "Bot ") {
-		token = "Bot " + b.GetString("Token")
+	if !b.GetBool("NotBot") {
+		if !strings.HasPrefix(b.GetString("Token"), "Bot ") {
+			token = "Bot " + b.GetString("Token")
+		}
 	}
 	b.c, err = discordgo.New(token)
 	if err != nil {


zomboy-alfrir:
> Isn't using your own account as a bot against the discord ToS?

I've never read about that, but with this we don't have to be a bot: We can use it only to read, and then use a bot account to write to a server.

I've updated my fork with your propsed changes

zomboy-alfrir:
I seem to have gotten a linter error yet I can't find any documentation for the suggested function:

```wrapperFunc: use strings.ReplaceAll method in `strings.Replace(token, "Bot ", "", -1)````

patcon:
Thanks @zomboy-alfrir! would you mind documenting this in the wiki in a way that makes sense to you? :)
https://github.com/42wim/matterbridge/wiki/Section-Discord-(basic)

AJolly:
That patch doesnt seem to work. I fixed it with the following:
	// if we have a User token, remove the `Bot` prefix
	if strings.HasPrefix(b.GetString("Token"), "User ") {
		token = strings.Replace(token, "Bot User ", "", -1)
	}

(now if we could just get telegram working without a bot user token! Maybe use PWRTelegram or use Telethon's code as a guide)

42wim:
@AJolly thanks for the comment! I've fixed it in e3cb665d9213a5cdc970eb8ad4db54ccb8c2cd4a

Issue #475, opened 2018-08-04
Whatsapp Support?


Tjird:
https://github.com/Rhymen/go-whatsapp
Maybe useful if you want to implement whatsapp support?

42wim:
Ok, interesting, api based on reverse engineering. Will not be done by me probably, but if someone implements it, I'll merge it.

moffz0r:
> Ok, interesting, api based on reverse engineering. Will not be done by me probably, but if someone implements it, I'll merge it.

I'd be interested in seeing this developed if anyone felt inclined to do so; I can offer limited help in testing too. It would be pretty cool to get WhatsApp bridged! :)

patcon:
Thanks for the feedback @jamoffat. We're tracking this is the linked mega-issue, so I think we'll prob close this out until someone is working on it :) cc @42wim

42wim:
Also interesting https://github.com/tulir/mautrix-whatsapp for anyone that wants to make a PR.

KrzysztofMadejski:
@42wim I'd give it a try. I'd welcome some guidance on how to start a new bridge, best if on wiki. :)

KrzysztofMadejski:
As I can mimick other implementations it would be even better to have some guidance on how to test it so I could proceed with TDD

KrzysztofMadejski:
I managed to receive the message from WhatsApp and send it out (I copied code from Gitter message handling). I have a connected channel on Slack, but the message didn't arrive there. 

Am I missing any fields?
```
time="2019-02-11T17:48:05+01:00" level=debug msg="<= Sending message from 48remotenumber@s.whatsapp.net on whatsapp.cfall to gateway" prefix=whatsapp
time="2019-02-11T17:48:27+01:00" level=debug msg="<= Message is config.Message{Text:"Bla", Channel:"mygroup@g.us", Username:"48remotenumber@s.whatsapp.net", UserID:"48remotenumber@s.whatsapp.net", Avatar:"", Account:"whatsapp.cfall", Event:"", Protocol:"", Gateway:"", ParentID:"", Timestamp:time.Time{wall:0x0, ext:63685500474, loc:(*time.Location)(0x2319220)}, ID:"7D5824A0570F2CBE19C1B775A5616A84", Extra:map[string][]interface {}(nil)}" prefix=whatsapp
```

Logs related to Slack connection looks ok:
```
[...]

PR #711, opened 2019-02-12
WhatsApp support

KrzysztofMadejski:
Alpha version. Discussion is on #475


42wim:
First of all, great job! Even more if this is your first go program 👍 
Also extra points for editing the README and the matterbridge.toml.sample 🥇 

Regarding the PR, some general comments before I look at the code:
* You can ignore all the fixmie messages for now, exported functions should have comments, see https://golang.org/doc/effective_go.html#commentary (so it would be nice if those could be done, but no blocker)
* The travis failure can not be ignored, the reason it fails now is because you didn't update the vendor directory. You can do this using `go mod vendor`
* About the code, can you restructure it like the slack code, put the handlers in `handlers.go`, helper functions in `helpers.go` and the main bridge functions in ``whatsapp.go`

KrzysztofMadejski:
On it, thank you!

My main concern right now is a UI one: a channel on Slack is shown as a unique ID string of the WhatsApp channel, which doesn't look great. I have to use unique IDs for groups, because WhatsApp allows for non-unique group names. I'm wondering if there is any possibility that in the config file there will be group ID, but Slack (and other bridges) would show some kind of group label? Did you have such problem for any of the other bridges/chats?

chat-slack

KrzysztofMadejski:
>  The travis failure can not be ignored, the reason it fails now is because you didn't update the vendor directory. You can do this using `go mod vendor`

Shouldn't downloading vendors be part of the build process? Keeping vendor files that can be accessed remotely in this repo sounds like a bad practice to me.

KrzysztofMadejski:
Question: When `func (b *Bwhatsapp) Disconnect()` is being called?

In WhatsApp I can [`logout`](https://github.com/Rhymen/go-whatsapp/blob/c31092027237441cffba1b9cb148eadf7c83c3d2/session.go#L377-L381), but I'm not sure if this is what we want.

KrzysztofMadejski:
> About the code, can you restructure it like the slack code, put the handlers in `handlers.go`, helper functions in `helpers.go` and the main bridge functions in ``whatsapp.go`

I separated also a `senders.go` file to put there all the handling coming from the relay into WhatsApp. It will grow substantially. Let me know if you would prefer it to be named differently.

42wim:
Some quick comments, later some more :-)
* vendor is added because
  * reproducible builds and backup (upstream can go *poof* anytime)
  * we need to build against older versions of go which don't  support the go modules but do support vendor

* Disconnect() isn't used for now, it's a placeholder, so no worries about that.

* senders.go (or whatever you choose) is fine, but please keep Send() in the main whatsapp.go

KrzysztofMadejski:
`All checks have passed`

KrzysztofMadejski:
But please hold some time with merging that in. I see that it is quite easy for telephone numbers to be exposed via relay. There might be no way around it, but at least I'd like to research that a little bit more and document for users to be aware of. #unintentedconsequences

42wim:
> My main concern right now is a UI one: a channel on Slack is shown as a unique ID string of the WhatsApp channel, which doesn't look great. I have to use unique IDs for groups, because WhatsApp allows for non-unique group names.

I don't understand this: "a channel on Slack is shown". 
You mean "a channel on slack is shown on whatsapp" ? because the attached screenshot is from slack? Need some more examples/screenshot to help :-)

> I see that it is quite easy for telephone numbers to be exposed via relay
You mean by running with `-debug` ? or because numbers can't get mapped to users?

> But please hold some time with merging that in.
Ok, please also remove the large commented part in the Send() in whatsapp.go and the slack references in New(). You also probably won't need the lru cache.
There's also a race condition in usersAvatars and users map, but we can fix this after the initial merge.

KrzysztofMadejski:
Sorry, I should have had elaborated more. One photo is worth a thousand words:

whatsapp

## 1. Channel Label

> My main concern right now is a UI one: a channel on Slack is shown as a unique ID string of the WhatsApp channel, which doesn't look great. I have to use unique IDs for groups, because WhatsApp allows for non-unique group names.

I'm using a `RemoteNickFormat="[{LABEL} #{CHANNEL}] @{NICK}"` which results in `[CfP #48111222333-1549986983@g.us] @Virus APP [10:45 AM]` being shown as sender. 

I don't know if WhatsApp is the first case of long unique channel id or did you have this issue before? What I'd love to have is an ability to set a channel label that would be used here instead of channel id when set.

```
    [[gateway.inout]]
    account="whatsapp.cfall"
    channel="48111222333-1549986983@g.us"
    channelLabel = "Code for Pakistan"
```
That would result in above RemoteNickFormat to be formatted as `[CfP #Code for Pakistan] @Virus APP [10:45 AM]`. But maybe you know a smart hack to handle it without introducing a new setting.

## 2. Unintended consequences
> I see that it is quite easy for telephone numbers to be exposed via relay

WhatsApp primarily identify users by telephone numbers. Altough internally there are independent unique ids being used, still I saw that telephone numbers pop out here and then. 

To bring the picture again:

whatsapp

Exposing numbers happens in two places (that I found so far):
1. group/channel ID contains the telephone number of the owner
2. mentioning other people on WhatsApp puts their telephone numbers as a mention "code".

While for WhatsApp using numbers is natural, relaying messages to many other bridges is an issue in my opinion. Ppl might not want their phone numbers to be exposed to "strangers" on public chats. How was it solved in Telegram?

Issue 1 could be solved by above `channelLabel` mechanism and issue 2 I can handle easily in the code. Regardless I think that somewhere in the Wiki we should put a notice that it might happen that numbers might be shared involuntarily to other bridges.

That made me also think about the EU's GDPR regulations and user consent before you share any of their personal data, but that's a bigger topic.

---

Thinking further into unintended consequences.. WhatsApp imposed a maximum limit of members in the group to stop disinformation from spreading. Disinfo bubbles spreading via WhatsApp groups and forwarding resulted in killings in India and Philippines. This bridge overcomes some of those limits. How do we address it? I might start another issue for #unintentedconsequences. Maybe do a review along https://ethicalos.org/ guidelines.

## 3. Code

> Ok, please also remove the large commented part in the Send() in whatsapp.go and the slack references in New(). You also probably won't need the lru cache.

Cleaned the code. Still I have to work on the commented part in the Send() to handle edits and deletes.

> There's also a race condition in usersAvatars and users map, but we can fix this after the initial merge.

Where is the race? I write to it in one place and time, and then only read from them. 

But for the future I will need to update them as ppl join the channels, so suggestion how to guard races would be welcomed.

patcon:
Perhaps related to "1. channel label" aspect: https://github.com/42wim/matterbridge/issues/614

A robust templating engine would allow people to set a high-level default RemoteNickFormat that took special-cases into account, without solving it in core:

```mustache
[{% label %}{% #if protocol != "whatsapp" %} #{% channel %}{% /if %}] @{% nick %}
```

42wim:
> ## 1. Channel Label
> > My main concern right now is a UI one: a channel on Slack is shown as a unique ID string of the WhatsApp channel, which doesn't look great. I have to use unique IDs for groups, because WhatsApp allows for non-unique group names.
> 
> I'm using a `RemoteNickFormat="[{LABEL} #{CHANNEL}] @{NICK}"` which results in `[CfP #48111222333-1549986983@g.us] @Virus APP [10:45 AM]` being shown as sender.
> 
> I don't know if WhatsApp is the first case of long unique channel id or did you have this issue before? What I'd love to have is an ability to set a channel label that would be used here instead of channel id when set.

Well, I think using {CHANNEL} in `RemoteNickFormat` is an edge case. I don't see it used in issues raised here (nor do I use it myself). Channels are mostly bridged 1:1 so you know what channel you've bridged and it can put it in the topic for example, no need to put it in every message ?
(Telegram also has those weird numeric ID channel names)

But I agree if you do use it, a channel-label feature would be nice (but it doesn't exist).

> But maybe you know a smart hack to handle it without introducing a new setting.

A possible hack to achieve this is to create a new whatsapp bridge per channel, this way you can use the existing `label` option.

> 
> ## 2. Unintended consequences
> > I see that it is quite easy for telephone numbers to be exposed via relay
> 
> WhatsApp primarily identify users by telephone numbers. Altough internally there are independent unique ids being used, still I saw that telephone numbers pop out here and then.
> 
> To bring the picture again:
> 
> whatsapp
> 
> Exposing numbers happens in two places (that I found so far):
> 
> * [ ]  group/channel ID contains the telephone number of the owner
> * [x]  mentioning other people on WhatsApp puts their telephone numbers as a mention "code".
> 
> While for WhatsApp using numbers is natural, relaying messages to many other bridges is an issue in my opinion. Ppl might not want their phone numbers to be exposed to "strangers" on public chats. How was it solved in Telegram?


Afaik telegram always returns a name and if the name is empty (when a user hasn't set a public name) we return "unknownUser" 

https://github.com/42wim/matterbridge/blob/5af1d8005509e42102b2e2c2319b1fd754bb46f3/bridge/telegram/handlers.go#L102-L105

I don't use telegram myself, so I haven't seen this in action.

> 
> Issue 1 could be solved by above `channelLabel` mechanism and issue 2 I can handle easily in the code [update: fixed already]. Regardless I think that somewhere in the Wiki we should put a notice that it might happen that numbers might be shared involuntarily to other bridges.

Sounds good.
 
> That made me also think about the EU's GDPR regulations and user consent before you share any of their personal data, but that's a bigger topic.
> 
> Thinking further into unintended consequences.. WhatsApp imposed a maximum limit of members in the group to stop disinformation from spreading. Disinfo bubbles spreading via WhatsApp groups and forwarding resulted in killings in India and Philippines. This bridge overcomes some of those limits. How do we address it? I might start another issue for #unintentedconsequences. Maybe do a review along https://ethicalos.org/ guidelines.

You raise some good points, but I have no idea how we could control this. I think this point should also be raised at the library, maybe messages have a "forwarding" count somewhere. Another option here is to detect if the message is a forwarded message and drop it (but anyway who wants to overrule this can just modify the code).

> 
> ## 3. Code
> > Ok, please also remove the large commented part in the Send() in whatsapp.go and the slack references in New(). You also probably won't need the lru cache.
> 
> Cleaned the code.
> 
> > There's also a race condition in usersAvatars and users map, but we can fix this after the initial merge.
> 
> Where is the race? I write to it in one place and time, and then only read from them.

The one in users map that I thought isn't there, you're indeed not writing again to the users map.
The usersAvatars is because you're filling the map in a go function that could still be running when receive a message. (in L83 of the diff)

You can run `go build -race` which will show you the detected races as you run matterbridge with the whatsapp bridge.

> 
> But for the future I will need to update them as ppl join the channels, so suggestion how to guard races would be welcomed.

Look at the slack bridge, it adds a mutex 
https://github.com/42wim/matterbridge/blob/5af1d8005509e42102b2e2c2319b1fd754bb46f3/bridge/slack/slack.go#L34

Which then get a R(ead)Lock when data is read, and a Lock when you also write data to your map
https://github.com/42wim/matterbridge/blob/5af1d8005509e42102b2e2c2319b1fd754bb46f3/bridge/slack/helpers.go#L14-L26

So you best use those functions to modify/read the map instead of directly accessing the map.
More info on: https://gobyexample.com/mutexes and https://medium.com/golangspec/sync-rwmutex-ca6c6c3208a0

Regarding the PR itself some small nitpicks left before it can be merged:
* please remove the .gitignore file
* also the GetBoolOrDefault() function is not necessary, especially because you're using `false` as default (and the default of bool is always `false` if not initialized).

KrzysztofMadejski:
## Channel Label

> > I'm using a `RemoteNickFormat="[{LABEL} #{CHANNEL}] @{NICK}"` which results in `[CfP #48111222333-1549986983@g.us] @Virus APP [10:45 AM]` being shown as sender.

> Well, I think using {CHANNEL} in `RemoteNickFormat` is an edge case. I don't see it used in issues raised here (nor do I use it myself). Channels are mostly bridged 1:1 so you know what channel you've bridged and it can put it in the topic for example, no need to put it in every message ?

I use Matterbridge to link different communities together. I'd like them to know where does message originated from so they can head to another chat to get into 1-on-1 direct messages or other channels that are not relayed. It's true though that in most cases the channel names are very similar. I might drop it for simplicity and to solve above problem temporarily.

> But I agree if you do use it, a channel-label feature would be nice (but it doesn't exist).

I've separated it into an issue, will work on it in the free time -> https://github.com/42wim/matterbridge/issues/725

## Code

> The usersAvatars is because you're filling the map in a go function that could still be running when receive a message. (in L83 of the diff)

What you said is true, but I don't see there any race conditions. The write function fills in the map at its own pace. It is the only thread that does the writing. The message handler reads from the usersAvatars - it will either:
1. return url for avatar -> cool we have it!
2. return nil if this wasn't set (yet) -> cool, we don't have it, we will get it next time

The setting of the url is done at once in the map, atomically, so no need to guard writing there.

I understand that Slack might need mutexes because you have multiple writers. Each message being handled might result in modifying the users map. Here there is no such issue

----

> * please remove the .gitignore file

I added it to make it harder to accidentally add developer's `matterbridge.toml` file with tokens. That's a standard procedure to ignore "secrets". But I removed it according to your suggestion.

-----

> * also the GetBoolOrDefault() function is not necessary, especially because you're using `false` as default (and the default of bool is always `false` if not initialized).

It is not neccessary, but it was making the code cleaner for me. Removed.

42wim:
Thanks for the changes.

For me the current code can be merged, let me know if it's ok for you too.
I'm going to release a 1.14.0 soon, would be nice if this could be included in that release.

> What you said is true, but I don't see there any race conditions. The write function fills in the map at its own pace. It is the only thread that does the writing.

Made a simple example of the issue on https://play.golang.org/p/ow2ik2qoTNA
Build that code with go build -race and run it, it'll show a datarace as below in the comment. 
(in your case the go thread will probably be finished before a textmessage is received)

KrzysztofMadejski:
It's good to go from my side. Thank you a lot for taking the time to peer-review it!

Tjird:
Maybe handy to make a example how to save the session file with a volume (For Docker).

KrzysztofMadejski:
@Tjird I'm not an expert of Docker, could you propose a `docker-compose.yml` that could help?

I added a mention of the need to store the session file to Azure deploy's guide https://github.com/42wim/matterbridge/wiki/Deploy:-Azure

Tjird:
@KrzysztofMadejski I'm making a PR myself for that small thing if this PR is merged.

42wim:
Let's party! 👍 
@KrzysztofMadejski thanks again!

PR #723, opened 2019-02-17
Allow zulip bridge to specify topic per channel. Closes #701

### LINKED ISSUE #701: Zulip per gateway Topic configuration
I'd like to be able to set Zulips topic per gateway, rather than per bridge. 
Currently, the topic is a bridge setting, and channel (stream in Zulips language) is configured per gateway.   

Add an ability to override the bridge setting and set Zulips topic per gateway:  
For example:

```
[zulip.streamchat]
Token="Token"
Login="botlogin"
Server="ZulipServerAddress/"

#Topic of the messages matterbridge will use
Topic="defaulttopic"

[[gateway]]
name="chat"
enable=true
	
	[[gateway.inout]]
	account="zulip.streamchat"
	channel="streamname"
	Topic="overridetopic"

```

42wim:
Allows you to specific a topic per channel

```
    [[gateway.inout]]
    account="zulip.streamchat"
    channel="general"
        [gateway.inout.options]
        topic="topic1"
```

### DIFF
--- a/bridge/config/config.go
+++ b/bridge/config/config.go
@@ -146,6 +146,7 @@ type Protocol struct {
 type ChannelOptions struct {
 	Key        string // irc, xmpp
 	WebhookURL string // discord
+	Topic      string // zulip
 }
 
 type Bridge struct {
--- a/bridge/zulip/zulip.go
+++ b/bridge/zulip/zulip.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"io/ioutil"
 	"strconv"
+	"sync"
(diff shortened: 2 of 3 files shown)


AJolly:
This doesn't seem to work for me.  It will always use the same topic: (From the gateway that is last in the config file).

42wim:
Could you paste your config? Are you using the same channel on different gateways?

42wim:
@AJolly PR has been reverted, didn't work for multiple gateways, see #751 for now way of configuring topics.

PR #726, opened 2019-02-19
allow sending discriminator with Discord username

DeclanHoare:
Add a new flag to the Discord account `UseDiscriminator` which will append the user's "#xxxx" discriminator to the username while `UseUserName` is also enabled.


42wim:
Thanks for the PR!

Did you test this? As it's not specified in config (https://github.com/42wim/matterbridge/blob/master/bridge/config/config.go#L75) it'll always return false.

Could you also add documentation about this setting to the `matterbridge.toml.sample` https://github.com/42wim/matterbridge/blob/master/matterbridge.toml.sample#L708

DeclanHoare:
Sorry, I've tested it both ways now and it's working

42wim:
Looks good, thanks!

PR #731, opened 2019-02-23
Add scripting (tengo) support for every incoming message

### LINKED ISSUE #558: Conditions and formatting scripts
Matterbridge seems to bridge things just fine, for example from Slack to Mattermost. However, we can't filter things and most of the formatting is lost. For example, if I have an app that exclusively integrates with slack, it would be great if I could forward the relevant messages with a specific formatting. Alternatively, one could filter certain messages, for example to bridge all messages that contain a specific keyword.

Basically something like Zapier for people that know how to write scripts / use the console. So basically two functions:
 - filter stage: decide under which conditions the message is forwarded
 - format stage: decide how / what parts of the message posted on the output protocol

Zapier

42wim:
TengoModifyMessage allows you to specify the location of a tengo (https://github.com/d5/tengo/) script.

This script will receive every incoming message and can be used to modify the Username and the Text of that message.
The script will have the following global variables:
to modify: msgUsername and msgText
to read: msgChannel and msgAccount

The script is reloaded on every message, so you can modify the script on the fly.

Example script can be found in https://github.com/42wim/matterbridge/tree/master/gateway/bench.tengo and https://github.com/42wim/matterbridge/tree/master/contrib/example.tengo

The example below will check if the text contains blah and if so, it'll replace the text and the username of that message.
```
    text := import("text")
    if text.re_match("blah",msgText) {
        msgText="replaced by this"
        msgUsername="fakeuser"
    }
```

More information about tengo on: https://github.com/d5/tengo/blob/master/docs/tutorial.md and
https://github.com/d5/tengo/blob/master/docs/stdlib.md

This also fixes #558


patcon:
This is pretty amazing, wim! Really excited to see how people use it :)

Is the idea to encourage existing transformations to be migrated to tengo? And will contrib be a place to collect general ones that others create? If so, I'd be happy to submit a PR to doc this sort of thing in readme!

42wim:
@patcon yes, my idea is to move as much as possible to tengo
Haven't thought about contrib yet, but it's a good idea.
Maybe a `contrib/tengo/` or if you have another idea to store them ?

Also just added another PR which allows tengo for the `RemoteNickFormat` #793

PR #756, opened 2019-03-04
Implement Rocket.Chat attachments handling

drndos:
I have implemented working version of rocket.chat attachments and tested it with Rocket.chat -> Slack.
TODO: 
- [ ] Generify DownloadFileAuth
- [ ] Remove unnecessary dependencies from handlers
- [x] do we actually need "rocket_attachments"?
- [x] More caching options ?
- [x] More checks if file is ok ?
- [ ] Even more checks if file is ok and better handling if not
- [x] Actually use cache ? is it even  necessary ?
- [ ] Think more about texts of attachment

If anyone has any worthy opinion please share, thanks.

### DIFF
--- a/bridge/helper/helper.go
+++ b/bridge/helper/helper.go
@@ -46,6 +46,29 @@ func DownloadFileAuth(url string, auth string) (*[]byte, error) {
 	return &data, nil
 }
 
+func DownloadFileAuthRocket(url string, token string, userId string) (*[]byte, error) {
+	var buf bytes.Buffer
+	client := &http.Client{
+		Timeout: time.Second * 5,
+	}
+	req, err := http.NewRequest("GET", url, nil)
+
+	req.Header.Add("X-Auth-Token", token)
+	req.Header.Add("X-User-Id", userId)
+
+	if err != nil {
+		return nil, err
+	}
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+	io.Copy(&buf, resp.Body)
+	data := buf.Bytes()
+	return &data, nil
+}
+
 // GetSubLines splits messages in newline-delimited lines. If maxLineLength is
 // specified as non-zero GetSubLines will also clip long lines to the maximum
 // length and insert a warning marker that the line was clipped.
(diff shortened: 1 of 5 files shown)


42wim:
Thanks for contributing 👍 
If you need any help feel free to ask here or on the [chat](https://github.com/42wim/matterbridge/#chat-with-us)

### Some answers on your questions:
> do we actually need "rocket_attachments"?

Too bad it's incompatible with slack attachments.. mattermost/slack use the same structure.
So I'm afraid we'll need to have a separate until a better idea pops up. Maybe call it "rocketchat_attachments" (it's longer, but more consistent with the naming)

> More checks if file is ok ?

I think you're referring to that check in the slack code ? That's because the slack servers are too slow (or matterbridge too fast) in downloading the file, the URL is already there, but the file wasn't on the server yet. I'm not sure if that's the case with rocketchat, so maybe you can drop that check.

> Actually use cache ? is it even necessary ?

With slack the cache is there because of the way slack announces file events, iirc it does it multiple times, so too avoid downloading the same file also multiple times the cache is there. If rocketchat does it only once, you could drop the cache.

### Travis/build errors
Before we can merge this, the errors on [travis](https://travis-ci.org/42wim/matterbridge/builds/501765450?utm_source=github_status&utm_medium=notification) should be fixed. 
Don't hesitate to ask questions if it isn't clear on how to fix this.

drndos:
more TODO:
* What to do when rocket.chat restarts, it seems that the matterbridge needs restart after that
* When sending ZIP files over the bridge it seems that either some HTML page is sent or the bridge hangs and fails
* Also it would be great if we could resend all the messages that didn't get relayed after hang

drndos:
Testing it on production like a boss :). I found some bugs and will need to implement some kind of monitoring because the process doesn't get killed but the messages are not getting relayed in some conditions.

42wim:
I've noticed the "What to do when rocket.chat restarts, it seems that the matterbridge needs restart after that" issue myself, and am running a patch on my own bot. https://github.com/42wim/matterbridge/commit/1539185bd47dba573c7bf7c6de2470336940bdf4

Not sure if you're seeing the same messages about invalid and expired login or if you have another issue.

42wim:
Let me know if you need any help. :)

42wim:
implemented in another PR

PR #762, opened 2019-03-10
Split-out Slack user and channel management

### LINKED ISSUE #759: Races in `slack/helpers.go`
**Describe the bug**
I believe there are races around the `b.users` usage in `slack/helpers.go`. Let me explain:

Let's look at `getUser` function
https://github.com/42wim/matterbridge/blob/57f156be8328529666aadaee780711f12b6e162b/bridge/slack/helpers.go#L14-L26

- Line 15 : the lock is set , that's good, other threads wait ..
- Line 16 : we are checking if the user is set, so far so good
- Line 17 : the lock is released (I believe it shouldn't)
- Line 18: let's assume the user is not set `ok == false`
- Line 21: we are populating user

Now another thread enters, the lock is not set (we unlocked in line 17), gets a lock, checks if the user is set on line 16 (it is not) and jump also into populating user in Line 21. 

The locks should be used to guard bigger chunks of logic, many lines, that needs to be invoked separately from other threads. Here, I believe that the whole function should be guarded as such:

```go
 func (b *Bslack) getUser(id string) *slack.User { 
 	b.usersMutex.RLock() 
        defer  b.usersMutex.RUnlock() 

 	user, ok := b.users[id] 
 	if ok { 
 		return user 
 	} 
 	b.populateUser(id) // BTW populate could be returning the user
        return b.users[id] 
}
```
----

Same goes for populateUser:
https://github.com/42wim/matterbridge/blob/57f156be8328529666aadaee780711f12b6e162b/bridge/slack/helpers.go#L73-L77

Helcaraxan:
First PR preparing the grounds for #759.

No functional changes. Only copying over methods to a seperate file and turning `handleRateLimit` into a stand-alone function.

Next step will attach channel and user management methods to dedicated, separate `structs` (to be defined).

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -158,7 +158,6 @@ linters-settings:
       - regexpMust
       - singleCaseSwitch
       - sloppyLen
-      - sloppyReassign
       - switchTrue
       - typeSwitchVar
       - typeUnparen
--- a/bridge/slack/helpers.go
+++ b/bridge/slack/helpers.go
@@ -1,7 +1,6 @@
 package bslack
 
 import (
-	"context"
 	"fmt"
 	"regexp"
 	"strings"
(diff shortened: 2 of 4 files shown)


Helcaraxan:
Also deactivating the `sloppyReassign` linter in `go-critic` as it was raising things that would result in unwanted variable shadowing.

PR #766, opened 2019-03-12
Refactor channel and user management

Helcaraxan:
Follow up on #762. After this one we can actually do the work for #759.

This PR moves all methods for handling user and channel information to separate `struct` types that can be used by the Slack bridge main `struct` to abstract away the required logic.


42wim:
:+1:

PR #767, opened 2019-03-18
Fix race-condition in populateUser()

### LINKED ISSUE #759: Races in `slack/helpers.go`

Helcaraxan:
Fix #759's root-cause of by introducing synchronisation points for individual user fetches.

### DIFF
--- a/bridge/slack/users_channels.go
+++ b/bridge/slack/users_channels.go
@@ -18,8 +18,9 @@ type users struct {
 	log *logrus.Entry
 	sc  *slack.Client
 
-	users      map[string]*slack.User
-	usersMutex sync.RWMutex
+	users           map[string]*slack.User
+	usersMutex      sync.RWMutex
+	usersSyncPoints map[string]chan struct{}
 
 	refreshInProgress bool
 	earliestRefresh   time.Time
@@ -31,6 +32,7 @@ func newUserManager(log *logrus.Entry, sc *slack.Client) *users {
 		log:             log,
 		sc:              sc,
 		users:           make(map[string]*slack.User),
+		usersSyncPoints: make(map[string]chan struct{}),
 		earliestRefresh: time.Now(),
 	}
 }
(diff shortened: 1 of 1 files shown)


42wim:
It's getting complex with all those mutexes.
I'm getting a panic, introduced in your previous PR (fb2667631da8c0a2849fbfa595a99090b0acc9ed) when testing with multiple slacks.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xc38d37]

goroutine 1 [running]:
github.com/42wim/matterbridge/bridge/slack.(*channels).populateChannels(0x0, 0xea2500)
        /home/wim/git/go/src/github.com/42wim/matterbridge/bridge/slack/users_channels.go:265 +0x37
github.com/42wim/matterbridge/bridge/slack.(*Bslack).JoinChannel(0xc00037a2a0, 0xc0000f3ae6, 0x7, 0xc0000f3aa0, 0xa, 0x108b319, 0x5, 0xc000464540, 0x11, 0xc000506810, ...)
        /home/wim/git/go/src/github.com/42wim/matterbridge/bridge/slack/slack.go:154 +0x43
github.com/42wim/matterbridge/bridge.(*Bridge).joinChannels(0xc000610e80, 0xc000506a80, 0xc000506ab0, 0x1, 0x1)
        /home/wim/git/go/src/github.com/42wim/matterbridge/bridge/bridge.go:72 +0x2f0
github.com/42wim/matterbridge/bridge.(*Bridge).JoinChannels(0xc000610e80, 0x0, 0x0)
        /home/wim/git/go/src/github.com/42wim/matterbridge/bridge/bridge.go:58 +0x3d
github.com/42wim/matterbridge/gateway.(*Router).Start(0xc0003dac80, 0x1302100, 0xc000506540)
        /home/wim/git/go/src/github.com/42wim/matterbridge/gateway/router.go:78 +0x4ed
main.main()
        /home/wim/git/go/src/github.com/42wim/matterbridge/matterbridge.go:57 +0x41a
```

42wim:
Found the issue in legacy connect #768

42wim:
Thanks btw, this isn't the most fun part to work on I can imagine :)

PR #780, opened 2019-03-31
CI fixes and improvements

Helcaraxan:
Multiple small pieces in this PR:
- Fixing the CI builds as it was broken due to the recent update of TravisCI config to Go 1.12.x
- Moving around CI steps to push them into CI scripts (that can be run locally) and use a build matrix to test in different environments.
- The `algo` parameter for the `unparam` linter has been removed and we
should thus no longer specify it in our `.golangci.yaml`. Also, bumping the GolangCI-lint
version to the latest available minor release.
   See:
   - mvdan/unparam@e6a6d1c51b6f03ac1a9d120f1ea07bc3a3f0e0b9
   - golangci/golangci-lint#367

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -105,10 +105,6 @@ linters-settings:
     # with golangci-lint call it on a directory with the changed file.
     check-exported: false
   unparam:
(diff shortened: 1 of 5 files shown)


42wim:
looking good, thanks!

Issue #175, opened 2017-05-23
insert zero width space in nicks on irc


NikkyAI:
Please answer the following questions. 

### Which version of matterbridge are you using?
run ```matterbridge -version```
version: 0.12.0 8c7549a

### Please describe the expected behavior.

prefixed nicknames on irc should have a no ping character or zero width space ` U+200B`

in the following line a zero width space is inserted at position 1 of the nickname
`<N​ickname> some message`

### Please describe the actual behavior. 

there is no inserted zwsp and there seem to be no options to enable such behaviour

#### (be sure to exclude or anonymize private data (tokens/passwords))

```toml
[irc]
    [irc.esper]
    Server="irc.esper.net:6697"
    Nick="ruby" #effectively "ruby_" while testing
    
    UseTLS=true
    UseSASL=true

    NickServNick="username"
    NickServPassword="hunter2"

    IgnoreNicks="cord spammer"

[discord]
    [discord.chat]
    Token="token"
    Server="servername"
    EditDisable=false
    EditSuffix=" (edited)"
    IgnoreNicks="[cord] spammer"
    RemoteNickFormat="<{NICK}> "
    ShowJoinPart=false

[[gateway]]
#REQUIRED and UNIQUE
name="cord"
enable=true
    [[gateway.inout]]
    account="irc.esper"
    channel="#cord"

    [[gateway.inout]]
    account="discord.chat"
    channel="cord"
```

42wim:
Sorry, I don't really understand. 
You want to insert a ZWSP after the nick ? and it doesn't work by using the RemoteNickFormat, is that the issue ?

NikkyAI:
well i am not sure what the remote nick format is ? i have not see nthat mentioned in the sample anywhere but i am probably just blind

and not after the nickname.. inside it

42wim:
You want to insert a ZWSP inside a nick ? why?
``` RemoteNickFormat="<{NICK}> "```  is what I am talking about

NikkyAI:
because people who are on both platforms ping themself when chatting .. and yes their irc clients are stupid but that seems to be the easiest way to fix this

NikkyAI:
and ignoring the bot would not really help because then other users that send through the bot cannot ping you

42wim:
Ok, I understand now.
Let me think about it.

NikkyAI:
and unrelated to that.. coloring irc names basd on role colors from irc? not sure if anybody would want it but it would be cool to have it available..
[...]

PR #799, opened 2019-04-14
Add remotenickformat-zerowidth.tengo to contrib

qaisjp:
Example situation:

- User `qaisjp` is on both Discord and IRC with the same username
- Whenever `@qaisjp` (Discord) speaks, a message is sent to IRC like `<qaisjp> message`.
- Because the IRC username is in that string `qaisjp` receives a notification on IRC which can be very annoying.

This pull request includes a zero-width space between the first and second character of `qaisjp` so that IRC clients are not nickalerted by usernames.

This is currently in use and currently works. This is also a common solution for many bridges.

### DIFF
--- a/contrib/remotenickformat-zerowidth.tengo
+++ b/contrib/remotenickformat-zerowidth.tengo
@@ -1,0 +1,16 @@
+/*
+This script will return the nick except with multi-character usernames
+containing a zero-width space between the first and second character letter.
+
+Single character usernames will be left untouched.
+
+This is useful to prevent remote users from nickalerting
+IRC users of the same name when the remote user speaks.
+
+This result can be used in {TENGO} in RemoteNickFormat.
+*/
+
+result = nick
+if len(nick) > 1 {
+    result = string(nick[0]) + "​" + nick[1:]
+}


42wim:
Unless i'm mistaken, that option already exists? https://github.com/42wim/matterbridge/blob/master/matterbridge.toml.sample#L150 

But it's a good example of a tengo script!
Thanks!

qaisjp:
> #The string "{NOPINGNICK}" (case sensitive) will be replaced by the actual nick / username, but with a ZWSP inside the nick, so the irc user with the same nick won't get pinged. See https://github.com/42wim/matterbridge/issues/175 for more information


Whoops, I had no idea that existed! Cheers 🙂

qaisjp:
Not sure the script is worth being in contrib now, maybe it would be better to add this as an example in the wiki and say that you can actually use `{NOPINGNICK}`.

42wim:
I disagree, it's a good example of the tengo "power" :)
I also recently updated the documentation with all the settings on the wiki: 
https://github.com/42wim/matterbridge/wiki/Settings#RemoteNickFormat

Next up is refactoring the matterbridge.toml.sample :-)

qaisjp:
Very nice page!

Issue #798, opened 2019-04-12
IRC colors should not get stripped when bridged to another IRC channel


chotaire:
### Introduction

In an earlier commit IRC colors were stripped so that they are not malformatted on services the messages are being bridged to. 

### Problem

However, if they are bridged to another IRC channel, they should definitely not get stripped as e.g. nagios, check_mk, jenkins and so many more applications use IRC colors for important tags such as **OK** and **FAIL**. 

### Proposal

I am proposing either one of the two: 

a) IRC colors get stripped when not sent to another IRC channel. If sent to an IRC channel however, they are retained.
b) IRC colors do not get stripped if sent to another service that supports color. If this is not another IRC channel, the IRC colors need to be reformatted so they are displayed correctly.

This could be made a configuration parameter. Option b) is a lot of work and testing, but color supporting services that would receive reformatted colors instead of stripped IRC colors could be added over time. 

PR: https://github.com/42wim/matterbridge/pull/801

42wim:
Good point, I'm going for option a) for now though as this seems the most useful.

chotaire:
I'm on standby once you need testing, my matterbridge is connected to multiple IRC networks.

chotaire:
Here's a quick hack: https://github.com/chotaire/matterbridge/releases/tag/v1.14.3-pre-no-irc-colors

qaisjp:
> I'm going for option a) for now though as this seems the most useful.

@42wim is there a standard way for "message translation"? `b)` here seems like a nicer solution imo as it stays true to original content wherever possible. But it doesn't seem easy to implement this whilst retaining the current decoupled nature of things.

(The reason I ask is so that later I can use the same method to explore "@mention" translation from one platform to another.)

42wim:
@qaisjp no there isn't a standard way.
Maybe with the tengo scripting support for in and out messages option b becomes easier?

42wim:
Fixed in https://github.com/42wim/matterbridge/pull/806 with tengo

PR #801, opened 2019-04-17
Show IRC colors when bridging to IRC

chotaire:
### Show IRC colors when bridging to other IRC channels

IRC colors were stripped when sending to another IRC channel.

- Removed generic IRC color stripping from IRC handlers
- Added IRC color stripping to all protocol bridges other than IRC

This should fix https://github.com/42wim/matterbridge/issues/798. I have only tested with IRC and Rocketchat, please review and optimize.

Thanks for your time.


chotaire:
All checks now passed. Push rights to https://github.com/chotaire/matterbridge/tree/irc-enhancements granted to 42wim. Feel free to optimize my mess and make IRC users happy :)

chotaire:
QA: I built a test binary which is available here: https://github.com/chotaire/matterbridge/releases/tag/v1.14.3-pre-irc-colors

qaisjp:
This regex is being redefined and recompiled every message send — are you sure you don't want to just define it once in another package and then use it?

chotaire:
Thank you for taking your time to review @quaisjp and you're right, that makes more sense. I'm new to Go. I'll work on a child package later this week. Have a good one! If you would like to commit something in the meantime, do not hesitate.

chotaire:
Updated binary at https://github.com/chotaire/matterbridge/releases

42wim:
Hi @chotaire thanks for your work, much appreciated!

### About the IRC colors PR
I'm adding some code to refactor this issue using tengo, by removing the code from matterbridge and moving them into scripts. That way everyone can easily add/remove features they don't like.
So I'm putting this PR on hold for now, if my redesign doesn't work out, we can look at merging your PR.

### About the `Verbose IRC joins/parts`
Good idea to make this more verbose, but could you open a separate PR for this?
You also don't need to make an option for this, it's a debug message, it benefits from giving more information, so you can make a PR to make your changes the default.

chotaire:
Hi wim, please look at the code for verbose IRC joins/parts. It's not only a debug message, it will be bridged over the relay.  Example:

VerboseJoinPart = true
`<system> Chotaire (chotaire@example.com) parts`

VerboseJoinPart = false
`<system> Chotaire parts`

I thereby believe it is good to make it an option. While IRC users would expect to see ident@host as a measure of identifying if someone is real, this information may not be wanted when bridging to other protocols, whether that be for privacy reasons or because they liked the previous behaviour. 

Thanks for taking your time to review.

42wim:
@chotaire oops, sorry.

Good arguments :)
Could you open a separate PR with this feature though? (looks good as-is)

chotaire:
Moving" Verbose joins/parts" to separate PR.

This is the binary for testing "Show IRC colors when bridging to IRC": 
https://github.com/chotaire/matterbridge/releases/tag/v1.14.3-pre2-irc-colors

chotaire:
Here is a merged binary with both https://github.com/42wim/matterbridge/pull/805 and https://github.com/42wim/matterbridge/pull/801:

♥ https://github.com/chotaire/matterbridge/releases/tag/v1.14.3-pre4

poVoq:
Any news when this will be merged?

chotaire:
> Any news when this will be merged?

It is no longer necessary to be merged. You can now use tengo to strip IRC colors only if you're not relaying to IRC. See https://github.com/42wim/matterbridge/pull/806

PR #806, opened 2019-04-18
Add scripting (tengo) support for every outgoing message

42wim:
Adds a new key `OutMessage` under `[tengo]` table, which specifies the location of the script that
will be invoked on each message being sent to a bridge and can be used to modify the Username
and the Text of that message.

The script will have the following global variables:
read-only:
```
inAccount, inProtocol, inChannel, inGateway
outAccount, outProtocol, outChannel, outGateway
```
read-write:
```
msgText, msgUsername
```
The script is reloaded on every message, so you can modify the script on the fly.

The default script in https://github.com/42wim/matterbridge/tree/master/internal/tengo/outmessage.tengo
is compiled in and will be executed if no script is specified.


chotaire:
This is a better approach than mine because it is more versatile. Dev build confirmed working, including live updates to the tengo script. However, documentation has issues, a lof of them. That needs to be fixed. 

For now the configuration is user-unfriendly so I would really like to emphasize that bridging IRC colors to IRC channels should be enabled by default (as this is very likely the desired behavior for most users). 

Color-stripping is normally done by IRC clients or by setting a chanmode. Enabling IRC colors (something that works on IRC by default) should not require messing around with tengo scripts, disabling them however is another story. I'll be happy to help with the docs.

contrib/outmessage-irccolors.tengo
```diff
@@ -0,0 +1,7 @@
+// See https://github.com/42wim/matterbridge/issues/798
+
+// if we're not sending to an irc bridge we strip the IRC colors
```
qaisjp: Should we be using all the `contrib/*.tengo` files by default? imo this would still be a bug with default configuration

42wim: The irccolors is already enabled by default, see https://github.com/42wim/matterbridge/blob/master/internal/tengo/outmessage.tengo which is also in this PR

The other contrib tengo files are just examples.

Or do I misunderstand you and is there another bug you're talking about?

qaisjp: Oh right! My bad

PR #831, opened 2019-05-27
Clean-up XMPP handling code

Helcaraxan:
Some clean-up to the XMPP, adding extra logging in the process. This should improve debuggability of XMPP related issues (as one was recently raised on the chat).


42wim:
LGTM, thanks for picking up the slack (or in this case the xmpp ;)

Issue #814, opened 2019-04-24
Ignore user_typing events received from Slack by IRC bridge (and others?)


patcon:
**Describe the bug**
Slack's user_typing messages show as blank messages with no `NICK` in IRC.

**To Reproduce**
Connect a Slack and IRC bridge to the same gateway. Enable `ShowUserTyping`. Type some text in slack, without hitting send. blank messages should appear.

**Expected behavior**
I'd expect thee user_typing messages to be passed through the gateway, but stopped at the IRC bridge code, and perhaps in all other bridges.

**Screenshots/debug logs**
![Screen Shot 2019-04-24 at 2 02 28 PM](https://user-images.githubusercontent.com/305339/56682563-d02ab500-6699-11e9-9d06-9248eb5799e3.png)

Logs: https://gist.github.com/patcon/cf4c2fc29dfc188fd320bc43c905d5c2

**Environment (please complete the following information):**
 - OS: linux
 - Matterbridge version: admittedly old fork with changes merged from 53c8670, at `v1.13.0-pre` (due to being on my translation fork). But a quick check seems to indicate it's still an issue.

**Additional context**

`user_typing` messages are exempt from the gateway-level filter for blank messages here:
https://github.com/42wim/matterbridge/blob/f04d360ee2664595c391358506eba060b4df2199/gateway/gateway.go#L265-L267

Wondering thoughts on a best path. For now, I could just add a special-case skip to the IRC bridge, but perhaps we could centrally manage the list of "compatible" protocols that handle user_typing events. For now, it would just be slack, but the bridge-level code could call back to a function that checks if it's compat, and skips these messages if so. Curious for any feedback on this approach :)

patcon:
Alternatively, what might the consequences of adding another skip-check for blank messages in IRC bridge code? Presumably that wasn't done in the past because it was caught in gateway, but now we pass through some action events with no text, so it's not totally consistent...

Helcaraxan:
This should have been resolved now with #832. 🙂

PR #832, opened 2019-05-27
Only forward user-typing messages if supported by protocol

Helcaraxan:
Fixes issue #814. 

This is a somewhat hacky way of achieving the required goal but it seems
like this is the least problematic way of getting there.

We might want to redesign some bridge information later such that we
have a standardised way of specifying what is and what isn't supported
by each chat protocol / bridge.

### DIFF
--- a/gateway/bridgemap/bridgemap.go
+++ b/gateway/bridgemap/bridgemap.go
@@ -3,35 +3,41 @@ package bridgemap
 import (
 	"github.com/42wim/matterbridge/bridge"
 	"github.com/42wim/matterbridge/bridge/api"
-	"github.com/42wim/matterbridge/bridge/discord"
-	"github.com/42wim/matterbridge/bridge/gitter"
-	"github.com/42wim/matterbridge/bridge/irc"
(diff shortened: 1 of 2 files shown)


42wim:
Thanks, the hack doesn't add too much, so good enough for now :)

gateway/bridgemap/bridgemap.go
```diff
@@ -3,35 +3,41 @@ package bridgemap
...
+		"sshchat":      bsshchat.New,
+		"steam":        bsteam.New,
+		"telegram":     btelegram.New,
+		"whatsapp":     bwhatsapp.New,
+		"xmpp":         bxmpp.New,
+		"zulip":        bzulip.New,
+	}
+
+	UserTypingSupport = map[string]struct{}{
+		"slack": {},
```
qaisjp: Discord does support sending user typing signals, but only as the general bot user. Would you accept a PR to add this?

patcon: Slack is the same way -- docs recommend renaming bot handle to "someone (on the other side of the bridge)" :)

You should def submit a PR for it!

Issue #835, opened 2019-05-29
XMPP skiping timestampless / all messages


Humorhenker:
**Describe the bug**
With some ejabberd XMPP servers all messages from XMPP get skipped by the bridge no matter to which service they are supossed to be relayed.

**To Reproduce**
Matterbridge config:
```
[general]
MediaServerDownload="https://URL/download"
MediaDownloadPath="PATH/download"
MediaDownloadSize=1000000
 
[xmpp.myxmpp1]
Server="SERVER:5222"
#Jid your userid
Jid="JID"
Password="PASS"
Muc="conference.SERVER"
Nick="NICK"
RemoteNickFormat="[{PROTOCOL}] <{NICK}> "
SkipTLSVerify=true
 
[telegram.mytelegram1]
#See https://core.telegram.org/bots#6-botfather
#and https://www.linkedin.com/pulse/telegram-bots-beginners-marco-frau
Token="TOKEN"
RemoteNickFormat="[{PROTOCOL}] <{NICK}> "
 
[irc.freenode1]
Server="irc.freenode.net:6697"
Nick="NICK"
NickServNick="NickServ"
NickServPassword="PASS"
UseTLS=true
UseSASL=false
SkipTLSVerify=false
RemoteNickFormat="[{PROTOCOL}] <{NICK}> "
 
[matrix.mymatrix1]
##Server is your homeserver (eg https://matrix.org)
Server="https://matrix.org"
 
#login/pass of your bot.
#Use a dedicated user for this and not your own!
#Messages sent from this user will not be relayed to avoid loops.
Login="USERLOGIN"
Password="PASS"
 
RemoteNickFormat="[{PROTOCOL}] <{NICK}> "
 
#Whether to send the homeserver suffix. eg ":matrix.org" in @username:matrix.org
#to other bridges, or only send "username".(true only sends username)
#OPTIONAL (default false)
#NoHomeServerSuffix=true
 
[[gateway]]
name="gateway1"
enable=true
 
[[gateway.inout]]
account="xmpp.myxmpp1"
channel="CHANNEL"
 
[[gateway.inout]]
account="telegram.mytelegram1"
channel="-CHANNELNUMBER"
 
#[[gateway.inout]]
#account="irc.freenode1"
#channel="#CHANNEL"
 
[[gateway.inout]]
account="matrix.mymatrix1"
channel="#CHANNEL:matrix.org"
```
Try the jabber.ccc.de server for example.

**Expected behavior**
[...]

PR #836, opened 2019-05-29
Xmpp stamp.is zero fixing issue #835

Humorhenker:
Giving the config option to fix issue [#835](https://github.com/42wim/matterbridge/issues/835)
and taking @Helcaraxan Clean-up XMPP handling code with it.

### DIFF
--- a/bridge/xmpp/xmpp.go
+++ b/bridge/xmpp/xmpp.go
@@ -331,10 +331,10 @@ func (b *Bxmpp) skipMessage(message xmpp.Chat) bool {
 		return true
 	}
 
-	if b.GetBool("skipdelayedmsg") {
+    if b.GetBool("skipdelayedmsgfix") {
+        return false
+    } else {
 		// skip delayed messages
-		return message.Stamp.IsZero()
-	} else {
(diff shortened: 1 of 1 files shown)


42wim:
Hi @Humorhenker thanks a lot for your contribution and debugging :-) 👍 

I propose to change the name of your bool to `DelayedEnable` or `BacklogEnable`. This way it possibly can be reused for other protocols and we can even add a `DelayedAmount` or `BacklogAmount` that in the future to limit the number of messages.

Your PR needs to make a few changes though before I can merge this:
* rebase on current master, @Helcaraxan PR got merged.
* Add the `DelayedEnable`  or `BacklogEnable` var also to https://github.com/42wim/matterbridge/blob/master/bridge/config/config.go#L81-L82 this way we know which variables are used.
* Add the variable with a little explanation why people would activate it to the doc in matterbridge.toml.sample https://github.com/42wim/matterbridge/blob/master/matterbridge.toml.sample#L214-L216

Thanks!

Humorhenker:
Thank you. I will open a new pull request.

PR #837, opened 2019-05-31
Fix broken go.mod

Helcaraxan:
Apparently our recent move back to the upstream WhatsApp library has resulted in our dependency graph breaking down. I tracked this down to the `maunium.net/go/mautrix-whatsapp` dependency having itself a broken `go.mod` which references an inexistent version for `github.com/Rhymen/go-whatsapp`.

We could not however downgrade as even those earlier versions would break down due to their internal reliance on a fork of `github.com/Rhymen/go-whatsapp`. I've put up a PR to fix their broken `go.mod` (tulir/mautrix-whatsapp#66) but in the mean time this did not solve the issue.

The current solution is to also override our dependency on `Rhymen/go-whatsapp` with the `tulir/go-whatsapp` fork as this is required by `tulir/mautrix-whatsapp`. I've already raised an issue (tulir/mautrix-whatsapp#67) about upstreaming changes in that fork so that we might in the future rely again on the upstream.


42wim:
The strange thing is that this PR build was OK, https://travis-ci.org/42wim/matterbridge/jobs/522271692
Seems like someone did a git push -force on the mautrix-whatsapp repo

Helcaraxan:
Closing in favour of #838.

PR #838, opened 2019-05-31
Fix go mod issue by removing whatsapp-ext

42wim:


42wim:
@Helcaraxan I suggest using this PR instead of yours #837 
Removes the dependency on the project as the usage is very minimal, that way we also don't need to do those hacky `go.mod` changes.

WDYT ?

PR #849, opened 2019-06-14
Add an option to skip the Mattermost server version check

joohoi:
This pull request adds a new bool configuration option: `SkipVersionCheck` that can be used to skip the Mattermost server version checks that are normally done when connecting. 

The usage scenario for this feature would be when the Mattermost instance is hosted behind a reverse proxy that suppresses "non-standard" response headers in flight. If the option is not defined in such scenario, `matterbridge` will fail to connect with an error message: 
```
[0000] FATAL main:         Starting gateway failed: Bridge mattermost.whatever failed to start: unsupported mattermost version:
```


42wim:
Looks good, thanks for the PR 👍

PR #850, opened 2019-06-15
Add .gitignore

qaisjp:


42wim:
Ok, seems like the gitignore request keeps coming back, so I'll add it :-)
Thanks 👍

qaisjp:
Does this need to be added to the 1.15.1 milestone (or is it not because it's not user-facing?)

PR #853, opened 2019-06-17
Support webhook message deletions (discord)

qaisjp:
Messages sent via webhook can now be deleted. It seems it can do this
without any special permissions.

This copies discordgo.WebhookExecute and makes it support the returning
of discordgo.Message.

A pull request has been sent upstream, so we should use that if
@bwmarrin accepts the pull request:

https://github.com/bwmarrin/discordgo/pull/663

Changes in behaviour (webhook mode only):
- Previously messages *edited* on other platforms would just be
retransmitted as a brand new message to Discord.
- Message *edits* will now be ignored.
- Debug: message edits will now print out a "permission error".

In the future it may be good to send an "message edited" react to those
webhook messages, so at least people know that the message was edited on
other platforms. (Even though it can't actually show the new message.)

Alternatively, message edits could just send a brand new message with a
link back to the old one. This is a little ugly but it would ensure that
Discord users are able to see the edited message. These "message edit
notifications" would be sent from the bot user (not from a webhook), so
we could edit the "edit notification" if subsequent edits to the
original message are made.

### DIFF
--- a/bridge/discord/discord.go
+++ b/bridge/discord/discord.go
@@ -208,11 +208,21 @@ func (b *Bdiscord) Send(msg config.Message) (string, error) {
 	b.channelsMutex.RUnlock()
 
 	// Use webhook to send the message
-	if wID != "" && msg.ID == "" {
+	if wID != "" && msg.Event != config.EventMsgDelete {
 		// skip events
 		if msg.Event != "" && msg.Event != config.EventJoinLeave && msg.Event != config.EventTopicChange {
 			return "", nil
 		}
+
+		// If we are editing a message, delete the old message
+		if msg.ID != "" {
+			b.Log.Debugf("Deleting edited webhook message")
+			err := b.c.ChannelMessageDelete(channelID, msg.ID)
+			if err != nil {
+				b.Log.Errorf("Could not delete edited webhook message: %s", err)
+			}
+		}
+
 		b.Log.Debugf("Broadcasting using Webhook")
 		for _, f := range msg.Extra["file"] {
 			fi := f.(config.FileInfo)
--- a/bridge/discord/discord.go
+++ b/bridge/discord/discord.go
@@ -208,7 +208,7 @@ func (b *Bdiscord) Send(msg config.Message) (string, error) {
 	b.channelsMutex.RUnlock()
 
 	// Use webhook to send the message
-	if wID != "" {
+	if wID != "" && msg.ID == "" {
 		// skip events
 		if msg.Event != "" && msg.Event != config.EventJoinLeave && msg.Event != config.EventTopicChange {
 			return "", nil
@@ -250,7 +250,7 @@ func (b *Bdiscord) Send(msg config.Message) (string, error) {
 				return "", err
 			}
 		}
-		err := b.c.WebhookExecute(
+		msg, err := b.webhookExecute(
 			wID,
 			wToken,
 			true,
@@ -259,7 +259,7 @@ func (b *Bdiscord) Send(msg config.Message) (string, error) {
 				Username:  msg.Username,
 				AvatarURL: msg.Avatar,
 			})
-		return "", err
+		return msg.ID, err
 	}
 
 	b.Log.Debugf("Broadcasting using token (API)")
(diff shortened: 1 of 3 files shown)


bridge/discord/discord.go
```diff
@@ -208,7 +208,7 @@ func (b *Bdiscord) Send(msg config.Message) (string, error) {
 	b.channelsMutex.RUnlock()
 
 	// Use webhook to send the message
-	if wID != "" {
+	if wID != "" && msg.ID == "" {
```
qaisjp: The permission error debug message (when editing) can be suppressed if I add a similar guard to the `// skip events` check below.

42wim:
Nice work! :+1:

But I think the changes in behaviour is an issue, I think people rely on having the edited message being relayed.

This could be fixed though I think if we treat an edit (for the webhook) as a delete and a new message.
So check `if msg.Event != delete && msg.ID != ""` => we have an edited message, delete the message with this ID and post the complete message.

Could be that I'm forgetting something though :-) what are your thoughts?

qaisjp:
Cheers.

> This could be fixed though I think if we treat an edit (for the webhook) as a delete and a new message.

(delete older messages and let the new ones be out-of-order)

We _usually_ do this manually, so I agree that it's a better option than just dropping edit events. (Sometimes we don't if edits are minor, we do whatever is cleanest.)

I've pushed a non-working fix up. It works for the first message, but does not work for subsequent edits. I think this is because existing IDs take precedence over newly returned IDs. I am not sure how to fix this!

qaisjp:
Could it be this? https://github.com/42wim/matterbridge/blob/b79bf7d4143f7a962faa00e10c2d6705524607ee/gateway/handlers.go#L234

and https://github.com/42wim/matterbridge/blob/120bf39f55c8df3d70a6263ccd2bff180d5e24e2/gateway/router.go#L141 (lines 141 and 145)?

I'm not sure if `lru.Cache.Add` updates the value if the key already exists.

qaisjp:
Oh! I think the fix is here:

https://github.com/42wim/matterbridge/blob/120bf39f55c8df3d70a6263ccd2bff180d5e24e2/gateway/router.go#L144

It should be

```diff
- // only add the message ID if it doesn't already exists
+ // only add the message ID if it has changed
```

qaisjp:
It works now. 

delete

There's a few different changes in there, so let me know if you don't want to squash, and I can clean up the commit history. Happy to get rid of 84c6e96 as well if you want.

gateway/router.go
```diff
@@ -134,17 +135,17 @@ func (r *Router) handleReceive() {
...
-			// only add the message ID if it doesn't already exists
-			if _, ok := gw.Messages.Get(msg.Protocol + " " + msg.ID); !ok && msg.ID != "" {
-				gw.Messages.Add(msg.Protocol+" "+msg.ID, msgIDs)
-			}
-			idx++
+
+			// Always add/update the message ID. This is necessary as msgIDs
+			// will change if a bridge returns a different ID in response to edits.
+			gw.Messages.Add(msg.Protocol+" "+msg.ID, msgIDs)
 		}
```
42wim: The `msg.ID  != ""` check is gone though, i'm not sure if this makes a difference.
I'm also not sure if your change is going to break other protocols, so just to be sure could you make this an exception for discord only?
My suggestion is to keep the original lines 143-146 and add a new check below just for discord (or if you're want to test this for the other protocols and it doesn't break anything thats fine too :-)
```
// Always add/update the message ID. This is necessary as msgIDs
// will change if a bridge returns a different ID in response to edits.
if msg.Protocol == "discord"  {
   gw.Messages.Add(msg.Protocol+" "+msg.ID, msgIDs)
}

42wim:
Great! nice animation btw :-)
Looking good except my one comment. 
I'll squash it when merging.

gateway/router.go
qaisjp: Yes, good catch, the removal of `msg.ID != ""` was an accident.

42wim:
Thanks!

bwmarrin:
Just an FYI, the submitted PR to the discordgo repo was merged.  So you can switch back to using it if you would prefer :)

PR #861, opened 2019-07-12
Fix #860: discord channel & category name clash

### LINKED ISSUE #860: Discord: Messages not relayed when channel and category names are identical
**Describe the bug**
In Discord if the category name (A) is identical to the channel name (B) and you try to set a gateway (C) as the channel name in the Matterbridge config file it will produce an error when sending messages to channel B.

**To Reproduce**
Steps to reproduce the behavior: Create a Discord server with the Category name "GENERAL" and a channel under that category called "general". Create an entry in the matterbridge.toml file that includes a gateway connection to the general channel:

```
##############################################

[[gateway]]
name="test-general"
enable=true

[[gateway.inout]]
account="discord.mydiscord"
channel="general"

[[gateway.inout]]
account="gitter.mygitter"
channel="testing/notgeneral"

##############################################
```

Messages will be relayed from the Discord channel to Gitter, but not the other direction.

**Expected behavior**
Messages would be transmitted between the Gitter channel and the Discord channel.

**Screenshots/debug logs**
Whenever the message send failure to the general channel happens the following error log is displayed:
`ERROR gateway:      SendMessage failed: HTTP 400 Bad Request, {"code": 50008, "message": "Cannot send messages in a non-text channel"}`

 - OS: Ubuntu 64bit
 - Matterbridge version: 1.15.0

qaisjp:
Untested, sorry.

Fixes #860

### DIFF
--- a/bridge/discord/discord.go
+++ b/bridge/discord/discord.go
@@ -95,7 +95,8 @@ func (b *Bdiscord) Connect() error {
 	b.channelsMutex.Lock()
 	for _, guild := range guilds {
 		if guild.Name == serverName || guild.ID == serverName {
-			chans, err := b.c.GuildChannels(guild.ID)
+			var chans []*discordgo.Channel
+			chans, err = b.c.GuildChannels(guild.ID)
 			if err != nil {
 				break
 			}
--- a/bridge/discord/discord.go
+++ b/bridge/discord/discord.go
@@ -95,12 +95,13 @@ func (b *Bdiscord) Connect() error {
 	b.channelsMutex.Lock()
 	for _, guild := range guilds {
 		if guild.Name == serverName || guild.ID == serverName {
-			b.channels, err = b.c.GuildChannels(guild.ID)
-			b.guildID = guild.ID
-			guildFound = true
+			chans, err := b.c.GuildChannels(guild.ID)
 			if err != nil {
 				break
 			}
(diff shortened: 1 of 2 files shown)


qaisjp:
```
bridge/discord/discord.go:98:11: shadow: declaration of "err" shadows declaration at line 52 (govet)
			chans, err := b.c.GuildChannels(guild.ID)
			       ^
```

Is this a new go vet default? I shadow variables all the time, especially with `err`. 😠

Helcaraxan:
Hello @qaisjp,

Sorry for jumping out of the woodwork (should really get back to doing some work on Matterbridge myself). The shadowing is not a new error, it's one of the default checks of `govet`. In a general manner shadowing can lead to unintentional bad code and other unnecessary dangerous circumstances.

In this case you can easily change the code by either renaming the indented `err` to `chanErr` (or similar) or you can predeclare `chans` with a `var chans ...` just above the currently failing line.

42wim:
TIL discord categories :)
@qaisjp thanks a lot for the quick patch 👍 
I've tested it patch and it works fine but it's still an issue for multiple channels with the same name, with this categories thing you can have x channels called `#general`.
So I'm thinking about expanding support for this by using `channel="category/channel"` 

Going to merge this for now though, thanks again.

PR #863, opened 2019-07-14
Add support for discord category channels (discord)

42wim:
This adds support for the discord category option that can be used
to group channels in. This means we can have multiple channels with
the same name.

We add the option to specify a category in the channel option of a
discord account under `[[gateway]]`

Besides `channel="channel"` or `channel="ID:channelID"`, now also
`channel="category/channel"` can be specified.

This change remains backwards compatible with people that haven't
specified the category and incorporates the fix in #861

@qaisjp sorry but this roll backs a bit of your code in #861 

Any comments?

### DIFF
--- a/bridge/discord/discord.go
+++ b/bridge/discord/discord.go
@@ -95,12 +95,10 @@ func (b *Bdiscord) Connect() error {
 	b.channelsMutex.Lock()
 	for _, guild := range guilds {
 		if guild.Name == serverName || guild.ID == serverName {
-			var chans []*discordgo.Channel
-			chans, err = b.c.GuildChannels(guild.ID)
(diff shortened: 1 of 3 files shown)


qaisjp:
Is [this] the line that incorporates the fix in #861?

[this]: https://github.com/42wim/matterbridge/pull/863/files#diff-60dd29b722f4e85b2f4716fda5db43eeR66

Much better than my solution!

42wim:
> Is [this](https://github.com/42wim/matterbridge/pull/863/files#diff-60dd29b722f4e85b2f4716fda5db43eeR66) the line that incorporates the fix in #861?

Yup

> Slash is an invalid character for a channel name on Discord, so this seems fine.

Ah good point about the slash, didn't think about that :)

PR #872, opened 2019-08-13
Support for webhook files in discord

MOZGIII:
This PR adds support for sending files via webhooks to discord. It's a work in progress, as it relies on a not-yet-merged change to `discordgo`. It is also not tested, because I worked on it while on the plane.
It is already captures the core idea of the change, and I'd appreciate a review.

### DIFF
--- a/bridge/discord/discord.go
+++ b/bridge/discord/discord.go
@@ -382,7 +382,12 @@ func (b *Bdiscord) handleUploadFile(msg *config.Message, channelID string) (stri
 // webhookSend send one or more message via webhook, taking care of file
 // uploads (from slack, telegram or mattermost).
 // Returns messageID and error.
-func (b *Bdiscord) webhookSend(msg *config.Message, webhookID, token string) (res *discordgo.Message, err error) {
+func (b *Bdiscord) webhookSend(msg *config.Message, webhookID, token string) (*discordgo.Message, error) {
+	var (
+		res *discordgo.Message
+		err error
+	)
+
 	// WebhookParams can have either `Content` or `File`.
 
 	// We can't send empty messages.
--- a/bridge/discord/discord.go
+++ b/bridge/discord/discord.go
@@ -261,16 +261,11 @@ func (b *Bdiscord) Send(msg config.Message) (string, error) {
 				return "", err
 			}
 		}
-		msg, err := b.webhookExecute(
-			wID,
-			wToken,
-			true,
-			&discordgo.WebhookParams{
-				Content:   msg.Text,
-				Username:  msg.Username,
-				AvatarURL: msg.Avatar,
-			})
-		return msg.ID, err
+		msg, err := b.webhookSend(&msg, wID, wToken)
+		if err != nil {
+			return "", err
+		}
+		return msg.ID, nil
 	}
 
 	b.Log.Debugf("Broadcasting using token (API)")
@@ -375,4 +370,52 @@ func (b *Bdiscord) handleUploadFile(msg *config.Message, channelID string) (stri
 		}
 	}
 	return "", nil
+}
+
+// webhookSend send one or more message via webhook, taking care of file
+// uploads (from slack, telegram or mattermost).
+// Returns messageID and error.
+func (b *Bdiscord) webhookSend(msg *config.Message, webhookID, token string) (*discordgo.Message, error) {
+	var err error
+
+	// WebhookParams can have either `Content` or `File`.
+	res, err := b.c.WebhookExecute(
+		webhookID,
+		token,
+		true,
+		&discordgo.WebhookParams{
+			Content:   msg.Text,
+			Username:  msg.Username,
+			AvatarURL: msg.Avatar,
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	if msg.Extra != nil {
+		for _, f := range msg.Extra["file"] {
+			fi := f.(config.FileInfo)
+			file := discordgo.File{
+				Name:        fi.Name,
+				ContentType: "",
+				Reader:      bytes.NewReader(*fi.Data),
+			}
+			_, err := b.c.WebhookExecute(
+				webhookID,
+				token,
+				false,
+				&discordgo.WebhookParams{
+					Username:  msg.Username,
+					AvatarURL: msg.Avatar,
+					File:      &file,
+				},
+			)
+			if err != nil {
+				return nil, fmt.Errorf("file upload failed: %s", err)
+			}
+		}
+	}
+
+	return res, nil
 }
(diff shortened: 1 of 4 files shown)


qaisjp:
Thank you for getting rid of the WebhookExecute fork introduced in #853.

I guess this depends on https://github.com/bwmarrin/discordgo/pull/684 (I doubt 831b5b3 will be accepted?)

MOZGIII:
Right, it is blocked, and it’s also wip. I wasn’t able to confirm the fix yet, or whether the code os working. I guess we if I test it and prove it working, @42wim himself can fork `discordgo` to keep it under control, and accommodate my branch with fixed `WebhookExecute`.

MOZGIII:
Added more fixed and tested it. Works for me. I can't provide any other assurance except that one.

bridge/discord/discord.go
```diff
@@ -376,3 +378,52 @@ func (b *Bdiscord) handleUploadFile(msg *config.Message, channelID string) (stri
 	}
 	return "", nil
 }
+
+// webhookSend send one or more message via webhook, taking care of file
+// uploads (from slack, telegram or mattermost).
+// Returns messageID and error.
+func (b *Bdiscord) webhookSend(msg *config.Message, webhookID, token string) (res *discordgo.Message, err error) {
```
42wim:
```suggestion
func (b *Bdiscord) webhookSend(msg *config.Message, webhookID, token string) (*discordgo.Message, error) {
```

We don't use named returns in matterbridge, just to be consistent remove this here also

bridge/discord/discord.go
```diff
@@ -376,3 +378,52 @@ func (b *Bdiscord) handleUploadFile(msg *config.Message, channelID string) (stri
...
+
+	if msg.Extra != nil {
+		for _, f := range msg.Extra["file"] {
+			fi := f.(config.FileInfo)
+			file := discordgo.File{
+				Name:        fi.Name,
+				ContentType: "",
+				Reader:      bytes.NewReader(*fi.Data),
+			}
+			_, e2 := b.c.WebhookExecute(
```
42wim:
```suggestion
			_, err = b.c.WebhookExecute(
```

bridge/discord/discord.go
```diff
@@ -376,3 +378,52 @@ func (b *Bdiscord) handleUploadFile(msg *config.Message, channelID string) (stri
...
+				webhookID,
+				token,
+				false,
+				&discordgo.WebhookParams{
+					Username:  msg.Username,
+					AvatarURL: msg.Avatar,
+					File:      &file,
+				},
+			)
+			if e2 != nil {
```
42wim:
```suggestion
			if err != nil {
```

bridge/discord/discord.go
```diff
@@ -376,3 +378,52 @@ func (b *Bdiscord) handleUploadFile(msg *config.Message, channelID string) (stri
...
+			token,
+			true,
+			&discordgo.WebhookParams{
+				Content:   msg.Text,
+				Username:  msg.Username,
+				AvatarURL: msg.Avatar,
+			},
+		)
+		if err != nil {
+			b.Log.Errorf("Could not send text (%s) for message %#v: %s", msg.Text, msg, err)
```
42wim: I think we better also do a `return nil,err` after the log line

bridge/discord/discord.go
```diff
@@ -376,3 +378,52 @@ func (b *Bdiscord) handleUploadFile(msg *config.Message, channelID string) (stri
 	}
 	return "", nil
 }
+
+// webhookSend send one or more message via webhook, taking care of file
+// uploads (from slack, telegram or mattermost).
+// Returns messageID and error.
+func (b *Bdiscord) webhookSend(msg *config.Message, webhookID, token string) (res *discordgo.Message, err error) {
+	// WebhookParams can have either `Content` or `File`.
+
```
42wim:
```
var (
     res *discordgo.Message
     err error
)
```

bridge/discord/discord.go
```diff
@@ -376,3 +378,52 @@ func (b *Bdiscord) handleUploadFile(msg *config.Message, channelID string) (stri
...
+				token,
+				false,
+				&discordgo.WebhookParams{
+					Username:  msg.Username,
+					AvatarURL: msg.Avatar,
+					File:      &file,
+				},
+			)
+			if e2 != nil {
+				b.Log.Errorf("Could not send file %#v for message %#v: %s", file, msg, e2)
```
42wim: I think we better also do a return nil,err after the log line

42wim:
```suggestion
				b.Log.Errorf("Could not send file %#v for message %#v: %s", file, msg, err)
```

42wim:
Nice work! 👍 
Just a lot of small nitpicks, added a suggestion where possible.

I agree with the idea of using our own fork, I've updated matterbridge/discordgo and created a `matterbridge` branch. 

Could you make a PR with your discordgo changes against https://github.com/matterbridge/discordgo/tree/matterbridge ?

bridge/discord/discord.go
MOZGIII: I tried it with this - turns out we then loose parts if message that could be otherwise delivered. I didn’t dig in deep enough into the downstream stack to figure out what possible errors we could get there, and went for a safe bet of trying to deliver with best effort, and not aborting on errors gave me best results. Unless there’s a more intelligent  ways of handling an error than just returning it - I don’t think this need to be changed.

bridge/discord/discord.go
MOZGIII: See my response below. This way of handling errors is well thought on my end. We can covert it to non-named return values, but I’d leave the logic of only returning the text message sending error after handling files as is.

bridge/discord/discord.go
MOZGIII: See below.

bridge/discord/discord.go
MOZGIII: See below.

bridge/discord/discord.go
42wim: Ok, that's fine for me if you're willing to that, refactor with non-named, but keep your logic intact.

42wim:
See my last comment regarding refactoring.
I'll also update matterbridge with the new discordgo fork later today.

MOZGIII:
Hey, I'll be out today, if you'll get to it, could you please take care of switching to named parameters to speed up the process? You should be able to push directly to the branch I'm using in this PR. Alternatively, I'll do it in a couple of days.

42wim:
Sorry for the delay, real life happened :(
I'll handle the conflicts and do the merge.

PR #873, opened 2019-08-13
make getChannelIdTeam behave like GetChannelId for groups

bucko909:
`GetChannelId` will support names generated from query groups when a team is not set, but *not* when a team is set since it falls through to `getChannelIdTeam` which has a different inner loop. This pull makes the two implementations do the same thing.

### DIFF
--- a/matterclient/channels.go
+++ b/matterclient/channels.go
@@ -38,7 +38,7 @@ func (m *MMClient) GetChannelHeader(channelId string) string { //nolint:golint
 
 func getNormalisedName(channel *model.Channel) string {
 	if channel.Type == model.CHANNEL_GROUP {
-		// (deprecated in favour of ReplaceAll in go 1.12)
+		// (deprecated in favor of ReplaceAll in go 1.12)
 		res := strings.Replace(channel.DisplayName, ", ", "-", -1) //nolint: gocritic
 		res = strings.Replace(res, " ", "_", -1)                   //nolint: gocritic
 		return res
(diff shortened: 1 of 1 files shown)


bucko909:
I believe this would be better factored out into some helper function, but I'm not really sure where to put it. I came across the discrepancy while testing https://github.com/42wim/matterircd/pull/255.

bucko909:
OK, so it looks like I've a choice: fail lint ("should use `ReplaceAll`") or fail tests (simply won't build on go before `ReplaceAll` was added). Your choice. :)

matterclient/channels.go
```diff
@@ -46,8 +46,8 @@ func (m *MMClient) GetChannelId(name string, teamId string) string { //nolint:go
 	for _, t := range m.OtherTeams {
 		for _, channel := range append(t.Channels, t.MoreChannels...) {
 			if channel.Type == model.CHANNEL_GROUP {
-				res := strings.Replace(channel.DisplayName, ", ", "-", -1)
```
42wim: Could you add `//nolint: gocritic` after the line

bucko909:
OK, that looks like it passes CI for you. I even made you a nice refactor.

bucko909:
Github believes there are changes requested, but I changed the lines in question and resolved the rquest. Am I missing something, or is Github?

42wim:
Sorry for the delay!
Thanks for your work 👍

PR #878, opened 2019-08-26
Replace bwmarrin/discordgo with matterbridge/discordgo

42wim:
Needed for #872


qaisjp:
Btw will this fork be kept up to date? Personally I don't think it makes sense for such a small addition which could have been copied into the main source

42wim:
@qaisjp well you're right of course, but this is the 2nd time we needed a fork of discordgo because of a patch that wasn't upstream yet.

With the new go modules `replace` option it seems to be very easy to switch between different forks without having to change the import paths in the code. 

Also, upstream takes most of the time a while to release a new stable version, the current one is from november 2018. I normally try to keep the forked dependencies up to date if applicable.

PR #923, opened 2019-10-20
Add support for receiving attachments (keybase)

goncalor:
These changes make the Keybase bridge able to receive attachments.
It seems the Keybase API does not currently support uploading files directly from RAM, so the file has to be temporarily written to the filesystem before uploading and is removed afterwards.

I tested these changes by sending attachments from Telegram to Keybase.
This is my first time both contributing to this project and writing Go. Improvements/suggestions are appreciated.

### DIFF
--- a/bridge/keybase/keybase.go
+++ b/bridge/keybase/keybase.go
@@ -79,16 +79,21 @@ func (b *Bkeybase) Send(msg config.Message) (string, error) {
 			return "", err
 		}
 		defer os.RemoveAll(dir)
-		fname := msg.Extra["file"][0].(config.FileInfo).Name
-		fdata := *msg.Extra["file"][0].(config.FileInfo).Data
-		fcaption := msg.Extra["file"][0].(config.FileInfo).Comment
-		fpath := filepath.Join(dir, fname)
-		if err = ioutil.WriteFile(fpath, fdata, 0600); err != nil {
-			return "", err
+
+		for _, f := range msg.Extra["file"] {
+			fname := f.(config.FileInfo).Name
+			fdata := *f.(config.FileInfo).Data
+			fcaption := f.(config.FileInfo).Comment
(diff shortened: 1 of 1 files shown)


qaisjp:
Have you considered sending a pull request to the Keybase bridge to add support for loading from memory?

goncalor:
@qaisjp if you mean the library `github.com/keybase/go-keybase-chat-bot/kbchat`, yes I thought about it and looked at the code a bit. But I don't think the keybase client itself supports this at the moment, so the library above couldn't provide it. I tried to mess around with `keybase chap api` and for uploads the filename seems to be mandatory (but I haven't found any great documentation for the API).

qaisjp:
Fair enough. Btw, your code only looks like it supports one attachment? 

[This search](https://github.com/42wim/matterbridge/search?q=Extra%5B%22file%22%5D&unscoped_q=Extra%5B%22file%22%5D) suggests you should be doing a for loop through `msg.Extra["file"]`

42wim:
Hi @goncalor thanks for your PR!
I agree with @qaisjp regarding the looping over the uploads, could you add this change? 
Then I can merge your PR

goncalor:
Hi @42wim. You're welcome. I agree with both you and @qaisjp, it should definitely loop through attachments.

I just have one question related to how `matterbridge` works and that was in part with I didn't add the looping from the start: it is ok to loop through the files and call `SendAttachmentByTeam()` for each of them (which should create more than one message on Keybase) but then return a single `resp.Result.MsgID`?

42wim:
For uploads of files we ignore the msgID because we can only return 1 msgID at the moment.
So it's fine to just return `"",nil` when finished

goncalor:
Should be OK to merge now. Please check if any further changes are needed. Thank you.

42wim:
looks good, thank you 👍

PR #970, opened 2020-01-07
Add websocket to API

haykam821:
This pull request implements the `/api/socket` endpoint, which is a websocket.

- [x] Server → client mirrors `/api/stream`
- [x] Client → server mirrors `/api/message`


bridge/api/api.go
```diff
@@ -44,6 +45,7 @@ func New(cfg *bridge.Config) bridge.Bridger {
 	e.GET("/api/health", b.handleHealthcheck)
 	e.GET("/api/messages", b.handleMessages)
 	e.GET("/api/stream", b.handleStream)
+	e.GET("/api/socket", b.handleWebsocket)
```
42wim:
```suggestion
	e.GET("/api/websocket", b.handleWebsocket)
```
I think this is a better name

42wim:
Thanks for the PR, looks good.
Take a look at the linting errors in the Travis CI build though.

bridge/api/api.go
haykam821: Renamed in df6875d.

haykam821:
I'll fix the lint errors.

This is still a draft for now since I haven't implemented the server-bound functionality to send messages to Matterbridge.

NikkyAI:
so did the travis-ci crash or what happened ?

qaisjp:
I don't think we use Travis anymore.

NikkyAI:
then maybe it should be changed to be not a `required` check anymore
or is this because this PR is a old draft?

42wim:
@NikkyAI If you can tell me how to do that let me know .. I removed everything from travis and it's still there.

42wim:
Still two linting issues, you can fix those by doing `_ = conn.WriteJSON(msg)`
As you implemented it now, it's blocking on the reader, so if you as client don't send any messages, you can also read no new messages from matterbridge.

Maybe https://github.com/gorilla/websocket/blob/master/examples/chat/client.go can be an inspiration?

42wim:
LGTM, thanks

PR #990, opened 2020-01-31
Add Docker Compose configuration

pbek:
I like the project and use it with Docker Compose, this would be the configuration for all who use Docker Compose too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,25 +262,7 @@ Usage of ./matterbridge:
 
 ### Docker
 
-Create your matterbridge.toml file locally eg in `/tmp/matterbridge.toml`
-
-```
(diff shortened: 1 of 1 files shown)


qaisjp:
Is it normal to use Docker Compose instead of plain Docker for single-service applications?

pbek:
If you are running your other containers also via docker compose it makes sens.

42wim:
@pbek thanks!

I think this would fit better in the wiki.

Would you like to make a page like this https://github.com/42wim/matterbridge/wiki/Deploy:-Azure but for docker (eg https://github.com/42wim/matterbridge/wiki/Deploy:-docker) and put the docker and your docker-compose example in it?

And then just add a link from the README to the wiki.

pbek:
Done.

42wim:
Thank you! 👍

Issue #997, opened 2020-02-03
[Discord --> Matrix] uploading image doesn't work


t-my:
**Describe the bug**
Uploading image Discord will just create a a link to Matrix which will result into 401 Unauthorized 

**To Reproduce**
1. Upload image to Discord
2. A Link will appear in Matrix 
3. Click the link

**Expected behavior**
Upload the image to Matrix instead of passing a URL

MayeulC:
The resulting link works fine in most cases to me, [here](https://matrix.to/#/!LZirCxnkkeBudrQzPj:matrix.org/$1580803385730431iuwMc:matrix.org?via=matrix.org&via=librem.one&via=privacytools.io) is an example in the Pinephone channel , [here an other one](https://matrix.to/#/!sbmInimykuNPjjIQMN:matrix.org/$1580575656211221ePwHc:matrix.org?via=matrix.org&via=t2bot.io&via=privacytools.io)

However, sometimes links are mangled as underscores are interpreted as _markdown_, like for [This IRC link](https://matrix.to/#/!sbmInimykuNPjjIQMN:matrix.org/$1580401999498387Vghxb:matrix.org?via=matrix.org&via=t2bot.io&via=privacytools.io) (IRC should never be parsed as markdown), or [that discord image](https://matrix.to/#/!sbmInimykuNPjjIQMN:matrix.org/$1580254981135025FkurN:matrix.org?via=matrix.org&via=t2bot.io&via=privacytools.io).

Here is the source for that last message:
```json
  "content": {
    "body": "[D] <The Computer Guy> https://forum.pine64.org/showthread.php?tid=8956 this page is broken\nhttps://cdn.discordapp.com/attachments/463285150898192384/671862828130500608/Screenshot_20200128-154230_DuckDuckGo.jpg",
    "formatted_body": "[D] &lt;The Computer Guy&gt; https://forum.pine64.org/showthread.php?tid=8956 this page is broken<br>\nhttps://cdn.discordapp.com/attachments/463285150898192384/671862828130500608/Screenshot<em>20200128-154230</em>DuckDuckGo.jpg",
    "msgtype": "m.text",
    "format": "org.matrix.custom.html"
  },
```
See the `<em>` tags? Those cut the lin, so it has to be copied, underscores inserted, and pasted in a browser.

This looks related to #989, so not sure if this has been fixed.

In any case, uploading the image to Matrix would be preferred.

PR #999, opened 2020-02-04
Enable intra-word emphasis supression in markdown

MartijnBraam:
This fixes plain links sent to Matrix being broken if they contain
underscores. Fixes issue #997


42wim:
Thanks! 👍

PR #1000, opened 2020-02-05
Add ability to procure avatars from the destination bridge

qaisjp:
This adds a new `UseRemoteAvatar bool` setting that makes it possible for gateways to procure avatars from destination bridges.

**Use case**

Messages from IRC use a static avatar on Discord. Sometimes a user will be on both IRC and Discord, so we could try to use the matching Discord user's avatar instead.

**How this fits in**

If a `msg.Avatar` is equal to `"UseRemoteAvatar"`, the gateway will ask the destination bridge to find an avatar based on the source message.

If the destination bridge doesn't implement `FindAvatar`, or `FindAvatar` was unsuccessful, it will drop back onto `IconURL`.

**Other notes**

I've left notes in the code about whether or not we want debug output. I suspect it will be very spammy for existing bridges that don't support `UseRemoteAvatar`. If we don't want debug messages, I can simplify this code to be less nested.

An alternate way to do this, without using an interface, is to move this logic to `Bdiscord`. Then in `gateway.go` we can just set avatar to `""` if the destination bridge is not discord. Would that be better?

### DIFF
--- a/bridge/discord/discord.go
+++ b/bridge/discord/discord.go
@@ -384,7 +384,7 @@ func (b *Bdiscord) webhookSend(msg *config.Message, webhookID, token string) (*d
 	// If avatar is unset, maybe we need to set the local avatar
 	if msg.Avatar == "" {
 		for _, val := range b.GetStringSlice("UseLocalAvatar") {
-			if msg.Protocol != val {
+			if msg.Protocol != val && msg.Account != val {
 				continue
 			}
 			if avatar, ok := b.findAvatar(msg); ok {
--- a/bridge/discord/discord.go
+++ b/bridge/discord/discord.go
@@ -381,6 +381,19 @@ func (b *Bdiscord) webhookSend(msg *config.Message, webhookID, token string) (*d
 		err error
 	)
 
+	// If avatar is unset, maybe we need to set the local avatar
+	if msg.Avatar == "" {
+		for _, val := range b.GetStringSlice("UseLocalAvatar") {
+			if msg.Protocol != val {
+				continue
+			}
+			if avatar, ok := b.findAvatar(msg); ok {
+				msg.Avatar = avatar
+			}
+			break
+		}
+	}
+
 	// WebhookParams can have either `Content` or `File`.
 
 	// We can't send empty messages.
@@ -429,4 +442,12 @@ func (b *Bdiscord) webhookSend(msg *config.Message, webhookID, token string) (*d
 		}
 	}
 	return res, err
+}
+
+func (b *Bdiscord) findAvatar(m *config.Message) (string, bool) {
+	member, err := b.getGuildMemberByNick(m.Username)
+	if err != nil {
+		return "", false
+	}
+	return member.User.AvatarURL(""), true
 }
--- a/bridge/discord/discord.go
+++ b/bridge/discord/discord.go
@@ -381,16 +381,16 @@ func (b *Bdiscord) webhookSend(msg *config.Message, webhookID, token string) (*d
 		err error
 	)
 
-	// If avatar is unset, maybe we need to set the local avatar
+	// If avatar is unset, check if UseLocalAvatar contains the message's
+	// account or protocol, and if so, try to find a local avatar
 	if msg.Avatar == "" {
 		for _, val := range b.GetStringSlice("UseLocalAvatar") {
-			if msg.Protocol != val && msg.Account != val {
-				continue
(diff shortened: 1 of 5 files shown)


qaisjp:
573816a77a2a59381b859e6cd3663366e2bf84a5 is not a necessary commit. I thought I needed it, but it turns out I didn't. I'm not sure why the receiving message is being modified here, though. Is that intentional?

I've also just decided to go for the simpler (non-interface) way of checking if a bridge supports UserRemoteAvatar. Much more consistent with the existing system.

qaisjp:
missing mutex lock/unlock

42wim:
Good idea!

I think we can keep the code cleaner (as in only touching discord code) if we rename `UseRemoteAvatar` to an `UseLocalAvatar` `[]string` option.

So you can set `UseLocalAvatar=["irc","mattermost"]` in your discord config.

That way we can do the checks in discord on `msg.Protocol` or if you want more specific also on `msg.Account`

Even better would be tengo, but because of the extra discord lib functions we can't do that.

Wdyt?

qaisjp:
Hmm, that would be a lot more flexible, _especially_ being able to test against `msg.Protocol` and `msg.Account`.

I'm assuming that a user can also just stick their `UseLocalAvatar` into the `[general]` section if they don't want to copy and paste the same line into every gateway.

I'll give it a go!

qaisjp:
Okay, I've done that. 

Let me know if you want to keep 669b5d5 or want me send it in another PR, it is not a necessary commit.

bridge/discord/discord.go
```diff
@@ -381,6 +381,19 @@ func (b *Bdiscord) webhookSend(msg *config.Message, webhookID, token string) (*d
...
+	// If avatar is unset, maybe we need to set the local avatar
+	if msg.Avatar == "" {
+		for _, val := range b.GetStringSlice("UseLocalAvatar") {
+			if msg.Protocol != val {
+				continue
+			}
+			if avatar, ok := b.findAvatar(msg); ok {
+				msg.Avatar = avatar
+			}
+			break
```
42wim: Isn't this break redundant ?

42wim:
Thanks!
Yes, please a different PR for https://github.com/42wim/matterbridge/commit/71b82612792f6958c9d123865c3a88f79450e0fa

Is 1c631a7341478a6d4bae7a56e292c01b0d25f3ab also really necessary? Did it fail otherwise? (if not really necessary, I'd rather not add it now)

Could you also update the `matterbridge.toml.sample` with the new feature?

bridge/discord/discord.go
qaisjp: I think it needs to be one line up

qaisjp:
> Yes, please a different PR for [71b8261](https://github.com/42wim/matterbridge/commit/71b82612792f6958c9d123865c3a88f79450e0fa)

Ok!

> Is [1c631a7](https://github.com/42wim/matterbridge/commit/1c631a7341478a6d4bae7a56e292c01b0d25f3ab) also really necessary? Did it fail otherwise? (if not really necessary, I'd rather not add it now)

Yes, necessary, otherwise `msg.Protocol` is `""`. I did a "Find all references" for `config.Message.Protocol` and it is only being set in a few places.

If I revert that commit and include this patch:

```patch
diff --git a/bridge/discord/discord.go b/bridge/discord/discord.go
index 00aa3ed1..247dd966 100644
--- a/bridge/discord/discord.go
+++ b/bridge/discord/discord.go
@@ -383,6 +383,7 @@ func (b *Bdiscord) webhookSend(msg *config.Message, webhookID, token string) (*d

        // If avatar is unset, maybe we need to set the local avatar
        if msg.Avatar == "" {
+               fmt.Printf("THE msg.Protocol ISSSSSSSS %#v\n", msg.Protocol)
                for _, val := range b.GetStringSlice("UseLocalAvatar") {
                        if msg.Protocol != val {
                                continue
```

I get the following output:

```
THE msg.Protocol ISSSSSSSS ""
```

> Could you also update the matterbridge.toml.sample with the new feature?

yes

bridge/discord/discord.go
qaisjp: No it's fine as it is, but it's confusing, so I will refactor this chunk

qaisjp:
OK, so:
- new PR is in #1004,
- tweaked the code surrounding that `break`
- updated `matterbridge.toml.sample`

Re-requesting review.

42wim:
LGTM, thanks for the work!

PR #1035, opened 2020-03-15
Fix duplicate separator on empty description/url (discord). Fixes #1008

### LINKED ISSUE #1008: No description in Discord embed causes duplicate separator
If an embed with no description is sent, the separator in the embed will duplicate.

## Reproduction Steps

* Receive a Discord embed without a description on another location.

## Expected Behavior

There should only be one separator between the embed’s title and URL when it has no description. Instead of `Title - - https://example.org`, it should be just `Title - https://example.org`.

## Screenshots / Logs

Embed with no description

## Environment

* **OS:** linux
* **Matterbridge version:** 1.16.5 4be00bbe

## Additional Context

The bug occurs here:

https://github.com/42wim/matterbridge/blob/4c44515f9df8eb0eb67116f05cb50820d800f961/bridge/discord/handlers.go#L106

```toml
```

42wim:


bridge/discord/handlers.go
```diff
@@ -103,7 +103,16 @@ func (b *Bdiscord) messageCreate(s *discordgo.Session, m *discordgo.MessageCreat
 	// if we have embedded content add it to text
 	if b.GetBool("ShowEmbeds") && m.Message.Embeds != nil {
 		for _, embed := range m.Message.Embeds {
-			rmsg.Text = rmsg.Text + "embed: " + embed.Title + " - " + embed.Description + " - " + embed.URL + "\n"
+			if embed.Title != "" {
+				rmsg.Text += "embed: " + embed.Title
```
qaisjp: Is it possible for `embed.Title` to be empty but for the other fields to be set? If so, should `embed: ` be missing whilst the others are appended?

42wim: Yeah probably :-)
I hate these kind of checks, do you know a more elegant way to handle this?

qaisjp: I hate them too. You could probably range over a string array, keeping track of whether you've appended something yet, and append the prefix / str / separator when needed.

But it's probably a bit unnecessary for appending three strings.

In the future with Go contracts we could probably do something like

```go
"embed: " + strings.Join(slices.RemoveZero([]string{embed.Title, embed.Description, embed.URL}), " - ")`
```

42wim: Went all the way now ;-)

Issue #1032, opened 2020-03-12
Telegram: Markdown V2 doc


burner1024:
Please add [Markdown V2](https://core.telegram.org/bots/api#markdownv2-style) to doc/examples.
The original markdown [isn't parsing underscores in links correctly](https://github.com/42wim/matterbridge/issues/444#issuecomment-397072900).
I tried setting `MessageFormat="MarkdownV2"` in `matterbridge.toml`, and surprisingly, I think it just works? At least, underscores aren't swallowed by the parser. Not yet sure if there are any side effects.

42wim:
Thanks! Seems to be a recent feature, december 2019
Could you make a PR with an updated `matterbridge.toml.sample`  ?
Also feel free to edit the wiki on https://github.com/42wim/matterbridge/wiki/Settings#telegram

burner1024:
Made, edited.

42wim:
Thanks for your help! 👍

PR #1037, opened 2020-03-16
added markdownv2 mode for telegram, see #1032

burner1024:
#1032 V2 works better for discord embeds, other than that I don't know if there are any differences.


42wim:
Thank you very much!

PR #1045, opened 2020-03-18
support JSON and YAML config formats

jakubgs:
The [`viper`](https://github.com/spf13/viper) library already supports JSON and YAML, we just need to pass the config type based on file extension, and default to TOML if it can't be detected.

This is useful because tools like [Ansible](https://www.ansible.com/) have filters like [`to_nice_yaml`](https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#filters-for-formatting-data) or [`to_nice_json`](https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#filters-for-formatting-data) but no `to_nice_toml`. Which makes generation of the config on the fly a pain.

### DIFF
--- a/bridge/config/config.go
+++ b/bridge/config/config.go
@@ -3,6 +3,7 @@ package config
 import (
 	"bytes"
 	"io/ioutil"
+	"path/filepath"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


42wim:
Thanks for the PR!
Not a big fan of yaml/json but can understand the use-case.
I'm not going to support yaml/json config files in issue bug reports though :-)

jakubgs:
Well, any bug in JSON/YAML parsing would be a `viper` bug, so you probably won't have to worry about that. But makes sense. I personally find TOML to be not readable for humans.

Thanks for merging, appreciate it!

42wim:
I'm not worried about the bugs in parsing.
I'm worried about people attaching .json / .yaml files to bug reports that I have to parse :)
Don't wonna start a yaml/toml war, but json definitely isn't not fun to read.

PR #1056, opened 2020-03-22
Don't transmit typing events from ourselves

qaisjp:
Typing events from the bridge results in an "infinite loop" of typing events


42wim:
LGTM, thx

Issue #969, opened 2020-01-06
Audio message link to mediaserver (origin: telegram) does not appear in matrix


darkmattercoder:
**Describe the bug**
I bridge matrix, telegram, irc, mattermost and xmpp and have a media server for media content. Normally only images are shared which is working in all combinations. Today we had an user sending an audio message from telegram, which was relayed as a link to an `ogg` in xmpp, mattermost and irc. But in the matrix room, the link did not appear.

**To Reproduce**
* Bridge telegram, xmpp, irc, mattermost and matrix
* Send an audio message from telegram

**Expected behavior**
* a link to the ogg file should appear in the matrix room

**Screenshots/debug logs**

Matterbridge log entry:

```
[...]

PR #1057, opened 2020-03-22
Remove build dependencies from final image (multistage build)

awigen:
This multistage build takes the resulting image size down from 346MB to
90MB.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,13 @@
-FROM alpine:edge
-ENTRYPOINT ["/bin/matterbridge"]
+FROM alpine:edge AS builder
 
 COPY . /go/src/github.com/42wim/matterbridge
-RUN apk update && apk add go git gcc musl-dev ca-certificates mailcap \
+RUN apk update && apk add go git gcc musl-dev \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -9,3 +8,8 @@ RUN apk update && apk add go git gcc musl-dev ca-certificates mailcap \
         && go build -x -ldflags "-X main.githash=$(git log --pretty=format:'%h' -n 1)" -o /bin/matterbridge \
         && rm -rf /go \
         && apk del --purge git go gcc musl-dev
+
+FROM alpine:edge
+RUN apk --no-cache add ca-certificates
```
42wim:
```suggestion
RUN apk --no-cache add ca-certificates mailcap
```
We need mailcap for mimetypes: see  #969

42wim:
Thank you, good idea!

Dockerfile
awigen: I was unsure if that was a build dependency or not. I'll update the branch.

42wim:
Thanks!

PR #1086, opened 2020-04-11
fix the behavior of ShowTopicChange and SyncTopic

ldruschk:
Currently, the "topic_change" events are ignored if both, `ShowTopicChange` and `SyncTopic` are set, and forwarded otherwise. This pull requests changes the behavior such that the events are only forwarded if one of those two config options is set to true and ignored otherwise.


42wim:
Thanks!

Issue #1024, opened 2020-03-05
Avatars of XMPP Users


Polynomdivision:
Since matterbridge allows avatar spoofing with bridges that support it, I decided to implement this functionality - as the current XMPP bridge does not.

In my current implementation, it requests the avatar of XMPP users whenever they send a message and stores it in the media server download. I haven't tested it in production, but the avatar gets correctly downloaded and saved.

This, however, required adding support for XEP-0084 (User Avatar) via a bit of XEP-0060 (Publish-Subscribe) to mattn/go-xmpp, which appears to be matterbridge's XMPP library's upstream.

For me, before I can contribute this code here, this requires that I either open a PR in go-xmpp's upstream and upgrade matterbridge in this regard or port it to matterbridge's fork of go-xmpp.

TL;DR: I have an (untested) implementation of spoofing of XMPP users' avatars, which requires
me to bring some changes into either 42wim/go-xmpp or mattn/go-xmpp

qaisjp:
Personally I recommend you attempt to get the changes merged upstream. The additions in the fork can always be rebased on top of the latest upstream.

I'm looking into getting our forked changes merged upstream.

Polynomdivision:
I have now created an [issue upstream](https://github.com/mattn/go-xmpp/issues/118) and will attempt to get my patches merged. Probably would've done it anyway.

42wim:
I agree with @qaisjp about trying upstream, I'll rebase the fork until it isn't necessary anymore.
You're welcome to open a PR at my fork too, (do it against the work branch).

FYI, I'll be doing some matterbridge work this weekend, so if you also have time it can get merged quickly in my branch, while waiting on upstream.

42wim:
@qaisjp thank you for merging the other stuff upstream btw! 💯

Polynomdivision:
Now that my User Avatar patch has been merged into matterbridge/go-xmpp, I will try to clean up my hacky branch of matterbridge and create a PR the coming days.

42wim:
@Polynomdivision thanks, I've just also updated matterbridge with your changes.

Polynomdivision:
I have now a prototype for which I rewrote my initial code to better
mimic what the mattermost bridge does. This led me to a curious
discovery: While the User Avatar specification for XMPP says that the
avatar data MUST be `image/png`, my server returned `image/jpg`.
Presumably because my web server has disabled MIME type sniffing, Discord - which is the bridge I am testing against - refuses to display my
[...]

PR #1092, opened 2020-04-17
Check if XMPP users support User Avatars

Polynomdivision:
As requested per [#1024](https://github.com/42wim/matterbridge/issues/1024#issuecomment-614878273) This PR checks whether a given XMPP user supports avatars before trying to request the data.

This PR is not yet tested or finished as it requires another patch to go-xmpp. Once this PR is not a WIP anymore, I will squash the commits.

**EDIT**: PR has been created [upstream](https://github.com/mattn/go-xmpp/pull/124). I also created one for the [go-xmpp fork](https://github.com/matterbridge/go-xmpp/pull/6)

### DIFF
--- a/bridge/xmpp/handler.go
+++ b/bridge/xmpp/handler.go
@@ -32,3 +32,13 @@ func (b *Bxmpp) handleDownloadAvatar(avatar xmpp.AvatarData) {
 		b.Remote <- rmsg
 	}
 }
+
+func (b *Bxmpp) handleDisco(items xmpp.DiscoItems) {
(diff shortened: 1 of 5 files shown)


Polynomdivision:
Okay, so I just tested my implementation and for some reason, instead of getting a list of supported items, the disco request returns a *feature-not-implemented*. It works when querying my full JID, but the JID provided by the MUC does not work. It could be that this is a prosody-specific problem, but I am not sure.

Another solution, though maybe not as nice as a Disco request, would be to just always request the avatar and set the availability to *unavailable* until we get the data.

PR #1117, opened 2020-04-23
Prevent Re-requesting XMPP Avatar Data

Polynomdivision:
This PR is another version of #1092 . This version does not discover what a user in a MUC supports, but just requests avatar data anyway. If we get a response: use it; if we don't: don't ask again.

The reason for this "alternate" version is that I cannot use #1092 on my prosody server (See the PR for more details). While #1092 would be the much more elegant way, this one is much simpler and is tested.

**EDIT**: Typo


42wim:
Thanks! sorry for the delay.

PR #1148, opened 2020-05-24
Update Dockerfile so inotify works

xnaas:
This change would be required for the Docker image to actually read `RELOADABLE` config options from the `matterbridge.toml`.

This edit would require https://github.com/42wim/matterbridge/wiki/Deploy:-Docker to be updated as well to mention that mounting would have to change to mounting a ***directory*** not a file. inotify inside Docker cannot read directly mounted files, only directories, for whatever reason.

-----
-----

Suggested wiki change:

## Docker

Create your matterbridge.toml file locally eg in `./conf/matterbridge.toml`.

```bash
docker run -ti -v ./conf:/conf
```

## Docker Compose

Create your matterbridge.toml file locally for example in `./conf/matterbridge.toml` and use this yaml as `docker-compose.yml`.

```yml
version: '3.7'
services:
  matterbridge:
    image: 42wim/matterbridge
    restart: unless-stopped
    volumes:
    - ./conf:/conf
```

Afterwards start the container with `docker-compose up -d`.

----

The above wiki changes could be altered as desired, but I wanted to illustrate what these Dockerfile changes would entail.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,4 +10,7 @@ RUN apk update && apk add go git gcc musl-dev \
 FROM alpine:edge
 RUN apk --no-cache add ca-certificates mailcap
 COPY --from=builder /bin/matterbridge /bin/matterbridge
+RUN mkdir /conf
+RUN touch /conf/matterbridge.toml
+RUN ln -sf /matterbridge.toml /conf/matterbridge.toml
 ENTRYPOINT ["/bin/matterbridge", "-conf", "/conf/matterbridge.toml"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,4 +10,4 @@ RUN apk update && apk add go git gcc musl-dev \
 FROM alpine:edge
 RUN apk --no-cache add ca-certificates mailcap
 COPY --from=builder /bin/matterbridge /bin/matterbridge
-ENTRYPOINT ["/bin/matterbridge"]
+ENTRYPOINT ["/bin/matterbridge", "-conf", "/conf/matterbridge.toml"]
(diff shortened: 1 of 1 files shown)


xnaas:
With my latest change:

> This will preserve setups that were configured to run the old way without breaking them and new configs can be setup "correctly" without issue.

I've tested it locally a few different ways and it seems to work every time.

qaisjp:
I'd suggest putting it in `/usr/share/matterbridge/matterbridge.toml` or `/etc/matterbridge/matterbridge.toml` or `/usr/local/etc/matterbridge/matterbridge.toml`, since those are more standard places to put these sort of config files.

xnaas:
In most Docker containers, I find the files are as near `/` (root) as possible. Most usually do a `/conf` or `/config` folder. I'm fine with whatever wim decides. 😛

42wim:
@xnaas thanks for your work, looks good, love it that you didn't break the "old" configs with it :+1:
I'd like to move it to `/etc/matterbridge/matterbridge.toml` like @qaisjp suggests, if you could change that, I can merge it.

xnaas:
Should be good now @qaisjp and @42wim. (As soon as it finishes building/testing, anyway.)

42wim:
Thanks yes update the wiki please if you want! 🙏 💯

SuperSandro2000:
There is no mention of this breaking change in the release notes.

xnaas:
It should be breaking anymore @SuperSandro2000. In my testing, the created symlink allows old configurations to continue working. They just don't get to take advantage of actually reading RELOADABLE config options (but they never could anyway).

SuperSandro2000:
My config wasn't at the old default place so the symlink does not work for me. A mention in the changelog would have been nice to catch that immateriality.

xnaas:
If your setup is the new way, then the wiki is up-to-date and the symlinking is irrelevant to you anyway. The symlinks are only for making sure stuff setup the old way stays working for those already running that way.

SuperSandro2000:
Let's get away from my setup. If someone only ready the release notes he won't notice this and that was my only intention to get this change more prominent in the release notes.

PR #1160, opened 2020-06-14
WA: reload user information when a new contact is detected

### LINKED ISSUE #796: [WhatsApp] Nicks of new contacts joining the channel instead of @someone
When someone new joins the channel while the bridge is working, the messages from this new person will be shown as `@someone` because we don't know the whatsapp nick of that person.

Load information of contacts joining the group.

We could show telephone number, but #privacyissues!

**Labels**
Between bug and enhancement. :)

nathanaelhoun:
Fixes #796

I tried to solve this problem, as it was annoying for my use case (we often have new people in the groups).

Before returning an empty string, we refresh the WhatsApp contacts and if we found the one we wanted, we can return a real name.

Tested on a mattermost<->whatsapp bridge, seems to work fine.

This is my first contribution, feel free to say if it is not the good way

### DIFF
--- a/bridge/whatsapp/helpers.go
+++ b/bridge/whatsapp/helpers.go
@@ -80,8 +80,33 @@ func (b *Bwhatsapp) getSenderName(senderJid string) string {
 		// if user is not in phone contacts
 		// it is the most obvious scenario unless you sync your phone contacts with some remote updated source
 		// users can change it in their WhatsApp settings -> profile -> click on Avatar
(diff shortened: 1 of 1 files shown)


42wim:
Sorry for the close/reopen, this was needed to start the new github action checks.

42wim:
Looks good, thank you!

nathanaelhoun:
Thanks @42wim !

PR #1161, opened 2020-06-18
Only colour IRC nicks if there is one.

TheHolyRoger:
Hello, very minor change, can we please implement this though?

I'm currently using tengo scripts for various modifications, but there is currently no way with tengo to set the first character in a message sent to IRC with the colour nick option turned on in config. matterbridge will always send an IRC colour/format code first.

My workaround is to send an empty nick string with this fix, which means I can make sure the first character from msgText is the first character in a message sent to IRC.


42wim:
Sorry for the close/reopen, this was  needed to start the new github action checks.

42wim:
Thanks!

PR #1163, opened 2020-06-24
matrix: add support for global display names

jlu5:
Add support for using Matrix display names (/nick) in place of the account name.

Currently this is fairly bare-bones and doesn't support per-room nicks (/myroomnick in Riot) or disambiguation. Both of those appear to require tracking the display name for each channel/user pair, so it's probably best left for a later PR.

### DIFF
--- a/bridge/matrix/matrix.go
+++ b/bridge/matrix/matrix.go
@@ -180,9 +180,21 @@ func (b *Bmatrix) handleEvent(ev *matrix.Event) {
 			return
 		}
 
+		// Use display name (/nick) if available
+		// XXX: no support for per-room display names or disambiguation yet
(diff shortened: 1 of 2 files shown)


42wim:
This means that on every message we get we do a call to get the displayname.
That's maybe something matrix/homeservers won't like :)
I think we better should use a cache for this. 

_Looking into the code it seems we also do this for the avatar without caching (I should fix that too :)_

I propose we now do a simple non-changing cache or time-based one and for a future PR we would need to react on the state changing events of the user to update the cache/avatar. (Ofcourse if you want to do this you also may do this in this PR).

42wim:
@jlu5 any feedback?

PR #1164, opened 2020-06-24
irc: skip gIRC built-in rate limiting

jlu5:
By default, [gIRC rate limits all outgoing messages](https://github.com/lrstanley/girc/blob/master/conn.go#L435-L436). I found this behaviour particularly confusing when I tried to flood exempt the relay bot (since I own the network), so I've added a gIRC option settings to disable flood protection. Since matterbridge already implements message throttling, I don't believe this is extra layer of throttling is necessary.

(Alternatively, https://github.com/overdrivenetworks/matterbridge/commit/a351cc6f3ceb63f33bcacfc599a4084747c366b5 makes this a toggleable option)


42wim:
Sorry for the close/reopen, this was needed to start the new github action checks.

42wim:
Thanks!

PR #1167, opened 2020-06-28
Add Nextcloud Talk support

gary-kim:
~~This is still a work in progress but I welcome any feedback anyone may have :).~~

### Ready for review :)

~~I forgot about documentation, gimme a sec~~

This adds basic support for Nextcloud Talk. I'm hoping to add support for stuff like replies and such in a future PR.

### DIFF
--- a/bridge/nctalk/nctalk.go
+++ b/bridge/nctalk/nctalk.go
@@ -1,0 +1,114 @@
+package nctalk
+
+import (
+	"context"
+	"strconv"
+
+	"github.com/42wim/matterbridge/bridge"
+	"github.com/42wim/matterbridge/bridge/config"
+
+	talk "gomod.garykim.dev/nc-talk"
+	"gomod.garykim.dev/nc-talk/ocs"
+	"gomod.garykim.dev/nc-talk/room"
+	"gomod.garykim.dev/nc-talk/user"
+)
+
+type Btalk struct {
+	user  *user.TalkUser
+	rooms []Broom
+	*bridge.Config
+}
+
+func New(cfg *bridge.Config) bridge.Bridger {
+	return &Btalk{Config: cfg}
+}
+
+type Broom struct {
+	room      *room.TalkRoom
+	ctx       context.Context
+	ctxCancel context.CancelFunc
+}
+
+func (b *Btalk) Connect() error {
+	b.Log.Info("Connecting")
+	b.user = talk.NewUser(b.GetString("Server"), b.GetString("Login"), b.GetString("Password"))
+	_, err := b.user.Capabilities()
+	if err != nil {
+		b.Log.Error("Cannot Connect")
+		return err
+	}
+	b.Log.Info("Connected")
+	return nil
+}
+
+func (b *Btalk) Disconnect() error {
+	for _, r := range b.rooms {
+		r.ctxCancel()
+	}
+	return nil
+}
+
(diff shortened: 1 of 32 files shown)


bridge/talk/talk.go
```diff
@@ -0,0 +1,116 @@
...
+			if msg.MessageType != ocs.MessageComment || msg.ActorID == b.user.User {
+				continue
+			}
+			remoteMessage := config.Message{
+				Text:      msg.Message,
+				Channel:   newRoom.room.Token,
+				Username:  msg.ActorDisplayName,
+				UserID:    msg.ActorID,
+				Account:   b.Account,
+				Timestamp: time.Unix(int64(msg.Timestamp), 0),
```
42wim, line 76: You don't need to set the Timestamp here, the gateway does this

42wim:
Cool!
The WIP looks good, added a small comment.
The linter is also complaining so please take a look at that too (we don't use named returns in our codebase)

bridge/talk/talk.go
gary-kim: Okay, thank you!

gary-kim:
Tested and working. I'll write a README for the library in the next 24 hours and tag it then this will be complete!

bridge/talk/talk.go
```diff
@@ -0,0 +1,114 @@
...
+	ctx       context.Context
+	ctxCancel context.CancelFunc
+}
+
+func (b *Btalk) Connect() error {
+	b.Log.Info("Connecting")
+	b.user = talk.NewUser(b.GetString("Server"), b.GetString("Login"), b.GetString("Password"))
+	_, err := b.user.Capabilities()
+	if err != nil {
+		b.Log.Info("Connected")
```
42wim, line 37: shouldn't this be `return err` ?
The err isn't nil so it's actually an error.

bridge/talk/talk.go
```diff
@@ -0,0 +1,114 @@
...
+	} else {
+		b.Log.Info("Cannot connect")
```
42wim, line 39: This should also change then

bridge/talk/talk.go
```diff
@@ -0,0 +1,114 @@
+package talk
```
42wim, line 1: As talk is too generic, I think we should better rename this bridge and package to `nctalk`

42wim:
Thank you! some small issues and it can get merged :)

eneiluj:
Hey there. I just tried it between NC Talk and Mattermost and it works great!

I guess you already know all that and want to take care of issues later. Just listing what I see:

* Nick message prefix works both ways.
* Emojis work both ways.
* When mentioning someone in the Talk room, it appears like that in the Mattermost channel: `{mention-user1} blabla`. Could it be replaced with something like `@username blabla`?
* When joining a file in the Talk room (new file or from storage), it shows `{file}`. Maybe those could be ignored. What do you think?
* Joining a file in the Mattermost channel displays an empty message in the Talk room. I guess it's a choice made in Mattermost backend.


Pretty exciting stuff. Cheers!

bridge/talk/talk.go
gary-kim: Whoops! I have no idea what I was thinking here. Will fix

gary-kim:
Thank you for your far more comprehensive tests @eneiluj 

I honestly forgot to test how files will send. Let me try that too.

PR #1168, opened 2020-07-01
Add an option to log into a file rather than stdout

z3bra:
This gives the ability to specify a filename to write the logs to, rather than `stdout`. It is useful for running it as a daemon when the service manager doesn't handle programs logging on stdout, like OpenBSD.

A better approach would probably be logging to syslog, so the system can handle file rotation, but I am not sure how to approach it. This is already useful as-is I think. Hope it helps !

### DIFF
--- a/bridge/config/config.go
+++ b/bridge/config/config.go
@@ -250,7 +250,7 @@ func NewConfig(rootLogger *logrus.Logger, cfgfile string) Config {
 	cfgtype := detectConfigType(cfgfile)
 	mycfg := newConfigFromString(logger, input, cfgtype)
 	if mycfg.cv.General.LogFile != "" {
-		logfile, err := os.OpenFile(mycfg.cv.General.LogFile, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0644)
+		logfile, err := os.OpenFile(mycfg.cv.General.LogFile, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0600)
 		if err == nil {
 			logger.Info("Opening log file ", mycfg.cv.General.LogFile)
 			rootLogger.Out = logfile
--- a/bridge/config/config.go
+++ b/bridge/config/config.go
@@ -1,9 +1,9 @@
 package config
 
 import (
-	"os"
 	"bytes"
 	"io/ioutil"
+	"os"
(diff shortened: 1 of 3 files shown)


qaisjp:
you probably also want to close the file when matterbridge is shut down

z3bra:
Indeed. I am not sure if what I did is the best approach, as I couldn't find anything related to a "clean" program termination.

bridge/config/config.go
```diff
@@ -247,6 +249,15 @@ func NewConfig(rootLogger *logrus.Logger, cfgfile string) Config {
 
 	cfgtype := detectConfigType(cfgfile)
 	mycfg := newConfigFromString(logger, input, cfgtype)
+	if mycfg.cv.General.LogFile != "" {
+		logfile, err := os.OpenFile(mycfg.cv.General.LogFile, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0644)
```
42wim, line 253: Maybe better to make this `0600` ?

matterbridge.go
```diff
@@ -51,6 +51,15 @@ func main() {
 	cfg := config.NewConfig(rootLogger, *flagConfig)
 	cfg.BridgeValues().General.Debug = *flagDebug
 
+	// if logging to a file, ensure it is closed when the program terminates
+	defer func() {
+		if f, ok := rootLogger.Out.(*os.File); ok {
+			if err := f.Sync(); err != nil {
+				f.Close()
```
42wim, line 58: We should always close the file not only when sync has an error.
I think we can just replace this by 

```
f.Sync()
f.Close()
```
Don't need to check the error there, it's not like we're able to fix it when it errors.
Put a `// nolint:errcheck` above the `defer func()`

42wim:
Thank you! Some minor nitpicks.

If possible, could you also add documentation about the option to the `[general]` section of the `matterbridge.toml.sample` file?

bridge/config/config.go
z3bra: Makes sense indeed.

z3bra:
I also added the setting documentation in the config sample. Hope that's clear enough, though the name should be pretty self-explanatory :)

Don't hesitate to report any other issue/problem you might have with this !

PR #1169, opened 2020-07-02
fix MarkdownV2 support in Telegram

greenx:
Without this, MarkdownM2 simply does not work.
Because only constant is taken from tgbotapi, here I just assigned the desired value.
Although in tgbotapi I also sent MR.

### DIFF
--- a/bridge/telegram/handlers.go
+++ b/bridge/telegram/handlers.go
@@ -322,9 +322,9 @@ func (b *Btelegram) handleEdit(msg *config.Message, chatid int64) (string, error
 	case "Markdown":
 		b.Log.Debug("Using mode markdown")
 		m.ParseMode = tgbotapi.ModeMarkdown
-	case "MarkdownV2":
+	case MarkdownV2:
 		b.Log.Debug("Using mode MarkdownV2")
-		m.ParseMode = "MarkdownV2"
+		m.ParseMode = MarkdownV2
 	}
 	if strings.ToLower(b.GetString("MessageFormat")) == HTMLNick {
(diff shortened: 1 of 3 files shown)


z3bra:
I gave it a go, and it works. However, this triggered a few issues for me when using MarkdownV2, and some messages were not relayed because the parsing failed:

    time="2020-07-08T16:53:55+02:00" level=error msg="SendMessage failed: Bad Request: can't parse entities: Character '.' is reserved and must be escaped with the preceding '\'" func=handleMessage file="gateway/handlers.go:228" prefix=gateway
    time="2020-07-08T16:56:18+02:00" level=error msg="SendMessage failed: Bad Request: can't parse entities: Character '(' is reserved and must be escaped with the preceding '\'" func=handleMessage file="gateway/handlers.go:228" prefix=gateway
    time="2020-07-08T16:56:34+02:00" level=error msg="SendMessage failed: Bad Request: can't parse entities: Character ')' is reserved and must be escaped with the preceding '\'" func=handleMessage file="gateway/handlers.go:228" prefix=gateway
    time="2020-07-08T17:04:51+02:00" level=error msg="SendMessage failed: Bad Request: can't parse entities: Character '!' is reserved and must be escaped with the preceding '\'" func=handleMessage file="gateway/handlers.go:228" prefix=gateway

This is probably expected though. It's just really annoying to loose messages because the "parsing failed".

42wim:
Yes, they should be escaped like mentioned in https://core.telegram.org/bots/api#markdownv2-style
But it's fine to keep this for another PR.

haykam821:
Relates to #1203 and #1225.

Issue #874, opened 2019-08-13
Telegram: animated stickers


Oreolek:
Telegram introduced animated stickers in TGS format. They don't get converted to PNG. And the browsers can't show them either. Maybe convert them to APNG?

42wim:
Sure, do you know a library that can do that?

Oreolek:
No and we'll have to wait until this will be in the open source Telegram version.

TomTheDragon:
The only tool that could do something like this (right now) is the python package TGS. It could convert those files into video or GIFs.
https://gitlab.com/mattia.basaglia/tgs

BenWiederhake:
There is the python library [lottie](https://pypi.org/project/lottie/), and after a single `pip3 install lottie cairosvg` you can convert stickers like this:

```
$ lottie_convert.py example.tgs example.webp
WebP frame rendering completed
WebP Writing to file...
```

Which results in a webp file (and GitHub can't deal with WebP yet).

Can this plus tengo do this? I hope that the image is already somewhere on the filesystem, in which case Tengo would need to intercept the outgoing message, call `lottie_convert.py --input-format lottie /path/to/the.tgs.webp /path/to/the.webp`, and modify the message to point to the new file instead (`/path/to/the.webp`).

Is there anyone here who knows tengo well enough?

42wim:
Fixed in #1173

Drjacky:
@BenWiederhake I did exactly what you said but, it doesn't find the `lottie_converter.py`!
<img width="315" alt="Screen Shot 2021-02-27 at 13 31 50" src="https://user-images.githubusercontent.com/2296754/109388622-27d28100-7900-11eb-8ee4-c26646c6c62a.png">

BenWiederhake:
Well, yes, as you observed correctly, it has to be actually installed.

Drjacky:
@BenWiederhake My bad. After installed https://pillow.readthedocs.io/en/stable/installation.html, it worked But, the result wasn't good enough. I've tested this sticker to JSON:
https://t.me/addstickers/RickAndMorty
In a few frames, the picture gets black-white. For sample:
[untitled folder.zip](https://github.com/42wim/matterbridge/files/6060985/untitled.folder.zip)

PR #1173, opened 2020-07-17
Support Telegram animated stickers (tgs) format

BenWiederhake:
This is half a fix for #874

This patch introduces two new config flags:
- MediaConvertTgsToWebP
- MediaConvertTgsToPNG

These need to be treated independently from the existing MediaConvertWebPToPNG flag because Tgs→WebP results in an *animated* WebP, and the WebP→PNG converter can't handle animated WebP files yet.

Furthermore, some platforms (like discord) don't even support animated WebP files, so the user may want to fall back to static PNGs (not APNGs).

The final reason why this is only half a fix is that this introduces an external dependency, namely lottie, to be installed like this:

```
$ pip3 install lottie cairosvg
```

This patch works by writing the tgs to a temporary file in /tmp, calling lottie to convert it (this conversion may take several  econds!), and then deleting the temporary file.
The temporary file is absolutely necessary, as lottie refuses to work on non-seekable files.
If anyone comes up with a reasonable use case where /tmp is unavailable, I can add yet another config option for that, if desired.

Here's a screenshot of Discord, with `MediaConvertTgsToPNG=true`:

Bildschirmfoto_2020-07-17_23-55-31

As this requires new text in the Wiki, I'll post some suggestions below.

### DIFF
--- a/bridge/config/config.go
+++ b/bridge/config/config.go
@@ -100,6 +100,7 @@ type Protocol struct {
 	MediaDownloadSize      int    // all protocols
 	MediaServerDownload    string
 	MediaServerUpload      string
+	MediaConvertTgs        string     // telegram
 	MediaConvertWebPToPNG  bool       // telegram
 	MessageDelay           int        // IRC, time in millisecond to wait between messages
 	MessageFormat          string     // telegram
--- a/bridge/helper/helper.go
+++ b/bridge/helper/helper.go
@@ -5,7 +5,10 @@ import (
 	"fmt"
 	"image/png"
 	"io"
+	"io/ioutil"
 	"net/http"
+	"os"
+	"os/exec"
 	"regexp"
 	"strings"
 	"time"
@@ -192,7 +195,7 @@ func ParseMarkdown(input string) string {
 	return res
 }
 
-// ConvertWebPToPNG convert input data (which should be WebP format to PNG format)
+// ConvertWebPToPNG converts input data (which should be WebP format) to PNG format
 func ConvertWebPToPNG(data *[]byte) error {
 	r := bytes.NewReader(*data)
 	m, err := webp.Decode(r)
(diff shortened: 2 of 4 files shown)


BenWiederhake:
Suggested wiki page: https://github.com/42wim/matterbridge/wiki/Settings

```
## MediaConvertTgsToWebP
Convert Tgs (Telegram animated sticker) images to WebP before upload.
This is independent of MediaConvertWebPToPNG. (This is useful because the result will be
an *animated* WebP file, and the WebP-to-PNG converter fails on animated WebP files.)
https://github.com/42wim/matterbridge/issues/874

Setting: OPTIONAL, RELOADABLE \
Format: boolean \
Example: enable it

`MediaConvertTgsToWebP=true`

## MediaConvertTgsToPNG
Convert Tgs (Telegram animated sticker) images to PNG before upload.
This is useful when your bridge also contains platforms that do not support animated WebP files, like Discord.
https://github.com/42wim/matterbridge/issues/874

Setting: OPTIONAL, RELOADABLE \
Format: boolean \
Example: enable it

`MediaConvertTgsToPNG=true`
```

BenWiederhake:
I disagree with the linter.
- `switch { case …` in `maybeConvertTgs`: Uhh, alright, if I must, but it looks weird to me.
- `logger` can be `github.com/stretchr/testify/assert.TestingT`: I don't even understand what the error is about.
- (lines 230 and 234) declaration of "err" shadows declaration at lines 224: I don't understand why it complains *here* but not usually.

qaisjp:
> * I don't understand why it complains _here_ but not usually.

The reason it complains is because Go supports this:

```go
func thisPanicsAndRecovers() (err error) {
	defer func() {
		message := recover()
		fmt.Println("Recovered from this panic:", message)
	}()

	err = errors.New("Some fake error")
	panic("aaaaa thisPanicsAndRecovers")
}
```

In the above code, the deferred function will get called when `panic()` is called. The deferred function recovers from the panic, and the outer function (`thisPanicsAndRecovers`) will return the error `"Some fake error"`. We get this output:

> Recovered from this panic: Some fake error

Now see this code:

```go
func someError() error {
	return errors.New("Error from if statement")
}

func thisPanicsAndRecoversFailingLint() (err error) {
	defer func() {
		message := recover()
		fmt.Println("Recovered from this panic:", message)
	}()

	err = errors.New("Some fake error")
	if err := someError(); err != nil {
		panic("noooooo")
	}
	panic("aaaaa")
}
```

One might expect the `if err := someError()` to update the first `err` variable, and for this to be the output:

> Recovered from this panic: Error from if statement

But it doesn't, the if statement creates a brand new `err` variable (it shadows the function's `err` variable that is auto-returned). We get this UNEXPECTED output:

> Recovered from this panic: Some fake error



----

You can try it out here: https://play.golang.org/p/Y-OaeUn03-J

qaisjp:
> `logger` can be `github.com/stretchr/testify/assert.TestingT` (interfacer)

Yeah, this one is weird. I have no idea.

BenWiederhake:
Thank you @qaisjp for the explanation, I totally forgot about the auto-returned variable :)

I'll update the code soon-ish.

BenWiederhake:
Updated. (And rebased for convenience.)

Now the only remaining problem is this linter false-positive: `##[error]`logger` can be `github.com/stretchr/testify/assert.TestingT` (interfacer)`
I'm going to ignore it.

qaisjp:
`maybeCleanup` is only used once, right? (In a `defer`.) Why not just defer an anonymous function? `defer func() { ... }()`

BenWiederhake:
Nice, I wasn't aware of that syntax. Pushed it.

qaisjp:
> `logger` can be `github.com/stretchr/testify/assert.TestingT` (interfacer)

Yeah so this uses [_interfacer_](https://github.com/mvdan/interfacer):

> **Deprecated:** A tool that suggests interfaces is prone to bad suggestions, so its usefulness in real code is limited. This tool will remain available as a proof of concept, and for others to examine and learn from.

https://github.com/golangci/golangci-lint/issues/541

And it looks like wim has just turned it off 2a41abb3d199085e648e820ba2e2b638e3c89e69 (1 hr after your last push)

So I guess just make another force push and `golangci-lint` should stop complaining.

42wim:
@BenWiederhake First off sorry for the delay and thank you for this PR.

This is a tricky one wrt the external dependency.

I'm fine with adding this feature with a few changes:
- As a user normally will use either `MediaConvertTgsToWebP` or `MediaConvertTgsToPNG`, I think it's better to make this a string value. `MediaConvertTgs="png"` or `MediaConvertTgs="webp"`, this also allows for other formats if ever necessary.
- If the user has enabled this on startup and the dependency is not found, matterbridge should treat this as an error and exit on startup with a helpful message to the user. You can do this by adding a `log.Fatal` in the `New()` method of telegram for now (as we don't have support for returning an error there yet)

What do you think?

42wim:
LGTM, thanks!

42wim:
I've added your change to the wiki

PR #1190, opened 2020-08-07
improve websocket handling

NikkyAI:
this is heavily based on https://github.com/gorilla/websocket/tree/master/examples/chat

in addition to the ringbuffer i added a broadcast group
the ringbuffer is no longer cleared and contains all history
endpoints that want to get "live" messages get them from 
```
member := b.group.Join()
msg := <- member.Read
```
see https://github.com/grafov/bcast#usage-
multiple websocket connections are now handled in parallel and each gets all messages (except from each other because there is no loopback i guess)
websockets now ping the connected clients and will close connection if no pong is received
and a few other things where the websocket is closed properly.. (before the tcp connection seemed to be hanging, did not verify wireshark though, could be other issues that got fixed along the way)

do we want to make the time between pings configurable ?

i would also really like it if each websocket connection had its own buffered channel for messages
as well as the `/api/messages` and `/api/stream` endpoints, maybe https://github.com/grafov/bcast could be used?

this would allow having multiple clients connect to the same websocket without "stealing" each others messages
thus making declaring one api endpoint for each bot no longer necessary

also, can we please just delete (or deprecate) the `/api/stream` endpoint? with websockets working that hack is no longer needed

### DIFF
--- a/bridge/api/api.go
+++ b/bridge/api/api.go
@@ -46,6 +46,11 @@ type Message struct {
 	Gateway  string `json:"gateway"`
 }
 
+type MessageWrapper struct {
+	message config.Message
+	member *bcast.Member
+}
+
 func New(cfg *bridge.Config) bridge.Bridger {
 	b := &API{Config: cfg}
 	e := echo.New()
@@ -110,7 +115,7 @@ func (b *API) Send(msg config.Message) (string, error) {
 	}
 	b.Log.Debugf("enqueueing message from %s to group broadcast", msg.Username)
 	b.Messages.Enqueue(msg)
-	b.group.Send(msg)
+	b.group.Send(MessageWrapper{msg, nil})
 	return "", nil
 }
 
@@ -169,17 +174,23 @@ func (b *API) handleStream(c echo.Context) error {
 		member.Close()
 	}()
 
-	for {
+	loop: for {
 		select {
 		// block until channel has message
 		case msg := <-member.Read:
-			if err := json.NewEncoder(c.Response()).Encode(msg); err != nil {
+			messageWrapper, ok := msg.(MessageWrapper)
+			if !ok {
+				break loop
+			}
+			message := messageWrapper.message
+			if err := json.NewEncoder(c.Response()).Encode(message); err != nil {
 				return err
 			}
 			c.Response().Flush()
 		}
 		time.Sleep(200 * time.Millisecond)
 	}
+	return nil
 }
 
 func (b *API) handleWebsocketMessage(message config.Message) {
@@ -193,7 +204,7 @@ func (b *API) handleWebsocketMessage(message config.Message) {
 	b.Remote <- message
 }
 
-func (b *API) writePump(conn *websocket.Conn, member bcast.Member) {
+func (b *API) writePump(conn *websocket.Conn, member *bcast.Member) {
 	ticker := time.NewTicker(pingPeriod)
 	defer func() {
 		b.Log.Debug("closing websocket")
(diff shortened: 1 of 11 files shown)


bridge/api/api.go
```diff
@@ -110,8 +119,21 @@ func (b *API) handlePostMessage(c echo.Context) error {
...
+	loop: for {
+		select {
+		case msg := <- b.send:
+			messages = append(messages, msg)
+		default:
+			break loop
+		}
+	}
+	// TODO: get all messages from send channel
+	c.JSONPretty(http.StatusOK, messages, " ")
```
NikkyAI, line 133: this seems to return `null` instead of `[]` for a empty slice

NikkyAI:
might be useful for others
i use the following `docker-compose.yml` for testing

```
version: '3.7'
services:
  matterbridge:
    build:
      context: "./"
    restart: unless-stopped
    ports:
      - "4242:4242"
      - "4343:4343"
    volumes:
      - ./run:/etc/matterbridge
```

`run/` contains the `matterbridge.toml`

although directly hitting the run task in goland is still faster, this came in handy when i tested on a different device that did not have golang installed

NikkyAI:
what is the correct process to add a import?
eg. `"github.com/grafov/bcast"`
when i hit sync 
goland modifies ALL the vendor files it seems, i assume i should not commit that?

NikkyAI:
currently multiple clients do not see messages from each other
this is fixable but then they see all messages, including their own

this should not be a problem however.. as long as as messages sent to the api can be identified and filtered out
i suggest to set the `userid` field on the message

NikkyAI:
i added a optiona header `skip_before` that can be passed a timestamp (RFC 3339 just like the timestamp field on the messages) that will make the api only send messages from history that are newer than that

the ring buffer is also not cleared anymore so multiple requests can retrieve the same history, also the reason filtering messages is useful

this can be used like a continuation token on reconnect as well, skipping all messages that the client already processed

42wim:
Sorry for the delay :/
The main issue I have is that this is getting complex quickly and it's going to be a pain to debug.
Before I review it completely, what is your take on using a library like https://github.com/olahol/melody which abstracts this kind of thing away.

NikkyAI:
a abstraction over this to push the responsibilty of handling the complex stuff to a library seems useful,
after all this is my first time touching go since the last PR

i can try to rewrite this using melody on the weekend

NikkyAI:
melody seems to not have been touched since 2018, its code is written against `echo.v1`, so this seems not the right library,
if anyone can find a similar enough abstraction over gorilla/websockets though i will be happy to use it

NikkyAI:
i managed to make melody work, i pushed it to another branch
https://github.com/42wim/matterbridge/compare/master...NikkyAI:feature/websockets-melody

still unsure about how to handle the history buffer, i would prefer to always send all history
and pass a token along to skip messages older than what the client has already seen, this is assuming it can trust the timestamps on api messages

other alternative is to never send history on websocket connections, this makes things a lot easier in matterbridge but might make clients more complicated because they have to connect using a 2 step method if they want to get the buffer

42wim:
@NikkyAI quick look at your branch, is the bcast still necessary with melody? I thought melody would handle the broadcast?

42wim:
To summarize what we chatted about:
- a client program to test the websockets (maybe https://github.com/vi/websocat)
- keep the PR's with limited scope, so this one just for adding decent multi-client websocket support (maybe use melody, newer fork on https://github.com/z9905080/melody)

PR #1195, opened 2020-08-13
Add TLSConfig to nctalk

gary-kim:
Still need to test. Give me a bit to set up an instance with improper SSL config.


bridge/nctalk/nctalk.go
```diff
@@ -31,8 +32,18 @@ type Broom struct {
 
 func (b *Btalk) Connect() error {
 	b.Log.Info("Connecting")
-	b.user = talk.NewUser(b.GetString("Server"), b.GetString("Login"), b.GetString("Password"))
-	_, err := b.user.Capabilities()
+	tconfig := &user.TalkUserConfig{
+		TLSConfig: &tls.Config{
+			InsecureSkipVerify: b.GetBool("SkipTLSVerify"), //nolint:gosec
```
gary-kim, line 37: I put `//nolint:gosec` here since it seems like gosec didn't like this.

gary-kim:
Confirmed to be working

bridge/nctalk/nctalk.go
qaisjp: I'm wondering if we could keep the setting name as `InsecureTLSVerify`, so that people using this setting explicitly know they're opting into insecure behaviour?

gary-kim: I personally think that would be a better idea but I decided to follow what the other bridges were using. Would it be better to change that?

https://github.com/42wim/matterbridge/blob/c63f08c8113598149f312bffd3fbdf1970660639/matterbridge.toml.sample#L347-L350

42wim: @gary-kim I agree with you, keep it consistent with what exists.
As for insecure/security, that's the reason why it's default disabled, so people still need to explicit enable it.

gary-kim: Sounds good to me

42wim:
Thanks 👍

PR #1205, opened 2020-08-23
replaced gorilla with melody

NikkyAI:
also added websocat command to the `matterbridge.tml.sample`


NikkyAI:
this does **NOT** send backlog to the websockets on connection
I plan to add that in a followup PR

`api/stream` still consumes messages from the buffer
and accessing `api/messages` will clear the message buffer

websockets do not consume messages as messages are directly sent to the melody router

bridge/api/api.go
```diff
@@ -33,6 +35,31 @@ func New(cfg *bridge.Config) bridge.Bridger {
...
+		message := config.Message{}
+		err := json.Unmarshal(msg, &message)
+		if err != nil {
+			b.Log.Errorf("failed to decode message from byte[] '%s'", string(msg))
+		}
+		b.handleWebsocketMessage(message)
+	})
+	b.mrouter.HandleConnect(func(session *melody.Session) {
+		greet := b.getGreeting()
+		bytes, err := json.Marshal(greet)
```
42wim, line 50: rename bytes to data so we don't have a collision with the `bytes` package

bridge/api/api.go
```diff
@@ -83,7 +110,14 @@ func (b *API) Send(msg config.Message) (string, error) {
 	if msg.Event == config.EventMsgDelete {
 		return "", nil
 	}
-	b.Messages.Enqueue(&msg)
+	b.Log.Debugf("enqueueing message from %s on ring buffer", msg.Username)
+	b.Messages.Enqueue(msg)
+
+	bytes, err := json.Marshal(msg)
```
42wim, line 116: rename bytes to data so we don't have a collision with the bytes package

42wim (changes requested):

42wim:
some small changes and it can be merged

NikkyAI:
rebased on master and rnamed the requested bits
(also one occurrence where i forgot to return in case of error)

bridge/api/api.go
NikkyAI: done

bridge/api/api.go
NikkyAI: done

PR #1213, opened 2020-08-25
Handle broadcasts as groups in Whatsapp

escoand:
The current way to get the correct JID of a WhatsApp group is to dump all JIDs to the log and grab the right one. This is working for for groups fine but not for broadcast, as they are not print out.

According to https://www.npmjs.com/package/@noamalffasy/js-whatsapp we have these possibilities:
* Chats: `[country code][phone number]@s.whatsapp.net`
* Groups: `[country code][phone number of creator]-[timestamp of group creation]@g.us`
* Broadcast Channels: `[timestamp of group creation]@broadcast`

But the bridge does currently interprets (and prints) the only second option.

### DIFF
--- a/bridge/whatsapp/whatsapp.go
+++ b/bridge/whatsapp/whatsapp.go
@@ -184,7 +184,7 @@ func (b *Bwhatsapp) Disconnect() error {
 }
 
 func isGroupJid(identifier string) bool {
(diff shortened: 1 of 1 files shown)


42wim:
Thanks 👍

PR #1216, opened 2020-08-26
loopback messages to other websockets as well

NikkyAI:
before multiple concurrently connected websocket clients on the same api would not see messages sent by other clients
sorry i missed that in #1205


NikkyAI:
maybe messages should also be appended to the ringbuffer there so they are accessible to `/api/stream` and `/api/messages` ?

PR #1219, opened 2020-08-28
Switch to upstream gomatrix

tilosp:
This way matterbridge is able to login in to newer homeservers like [conduit](https://conduit.rs/), that only have the new login endpoint.


42wim:
Thanks for looking into this :+1:
Yet another fork that can be removed :)

PR #1220, opened 2020-08-30
Update nc-talk to version 0.1.2

tilosp:
This solves a problem where matterbridge spams a lot of old nctalk messages into bridged rooms on every restart.

See: https://github.com/gary-kim/go-nc-talk/pull/14

cc: @gary-kim


42wim:
👍

PR #1222, opened 2020-08-30
nctalk: Format rich object strings

tilosp:
Currently nctalk mentions show up as `{mention-user1}`, `{mention-user2}`, ... on bridged platforms,
with this they show up as `@somename`.


42wim:
👍

PR #1237, opened 2020-09-09
Matrix: Permit uploading files of other mimetypes

BenWiederhake:
This includes at least c-source-files, cpp-source-files, markdown-files, Rust-files, and plaintext files.

We already allow uploading arbitrary executables. (And javascript-files, coincidentally.) Not permitting these other text files would be highly unexpected.

The "weird" lower part of the diff just moves the "executable" handler to the bottom, and marks it as default.


BenWiederhake:
Codeclimate criticizes code that I only *moved*. In fact, I would argue there is less duplication than before, because the `executable` case got more general.

Therefore, I'm not yet convinced to change the PR. (If you insist that I change the PR, please tell me *how* you would like me to avoid the duplication that was already there.)

BenWiederhake:
@42wim What do you think of this PR? :)

42wim:
LGTM, thank you! sorry for the delay

PR #967, opened 2019-12-26
Add initial Microsoft Teams support

42wim:
This implements Microsoft Teams support.

It uses polling to check for new messages because the ChatMessagesDelta isn't working (https://github.com/microsoftgraph/microsoft-graph-docs/issues/6451)

Documentation on how to setup can be found on https://github.com/42wim/matterbridge/wiki/MS-Teams-setup

# working

- [x] Receiving and sending messages.
- [x] Listening on multiple channels.
- [x] Messages edited in Microsoft Teams will also be edited in bridges that support message editing (mattermost, slack, discord, ..)
- [x] Richtext/HTML messages from Microsoft Teams will be sent as Markdown to other bridges.
- [x] Receiving attachments from Microsoft Teams to other bridges
- [x] Replies from mattermost/slack to Microsoft Teams
- [x] Code snippets from Microsoft teams to other bridges

# not possible because Microsoft API limitations

The non-supported API issues are sending stuff to msteams, receiving stuff from msteams mostly works. Doesn't work:

- [x] Messages edited in other bridges also edited in MS teams
- [x] Sending HTML/codeblock/emoji messages to Microsoft Teams (they will not be rendered as such)
- [x] Sending messages with attachments and images is not supported.

PR #1240, opened 2020-09-17
MSTeams threading

matthieu-foucault:
Would love to get some early feedback on this.

TODO: 
 - [ ] Polling for replies is a bit slow, I'm wondering if there is some info somewhere as to whether a message has replies
 - [ ] Add some documentation indicating that `PreserveThreading` is now an option for MSteams.

### DIFF
--- a/bridge/msteams/msteams.go
+++ b/bridge/msteams/msteams.go
@@ -118,6 +118,16 @@ func (b *Bmsteams) sendReply(msg config.Message) (string, error) {
 	return *res.ID, nil
 }
 
+func (b *Bmsteams) getReplies(channel string, msg msgraph.ChatMessage) ([]msgraph.ChatMessage, error) {
+	ct := b.gc.Teams().ID(b.GetString("TeamID")).Channels().ID(channel).Messages().ID(*msg.ID).Replies().Request()
+	rct, err := ct.Get(b.ctx)
+	if err != nil {
+		return nil, err
+	}
+	b.Log.Debugf("got %#v replies", len(rct))
+	return rct, nil
+}
+
 func (b *Bmsteams) getMessages(channel string) ([]msgraph.ChatMessage, error) {
 	ct := b.gc.Teams().ID(b.GetString("TeamID")).Channels().ID(channel).Messages().Request()
 	rct, err := ct.Get(b.ctx)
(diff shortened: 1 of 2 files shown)


bridge/msteams/msteams.go
```diff
@@ -118,13 +118,30 @@ func (b *Bmsteams) sendReply(msg config.Message) (string, error) {
...
 	if err != nil {
 		return nil, err
 	}
 	b.Log.Debugf("got %#v messages", len(rct))
+	for _, msg := range rct {
+		replyct, replyerr := b.getReplies(channel, msg)
+		if replyerr != nil {
+			return nil, replyerr
+		}
+		rct = append(rct, replyct...)
```
matthieu-foucault, line 143: Just starting to discover Go, so this might not be the way to do this

42wim:
Thanks for taking a look at this! :+1:

> Polling for replies is a bit slow, I'm wondering if there is some info somewhere as to whether a message has replies

Yes, that's why I didn't include it in my PR because with thousands of messages this becomes very very slow.

According to the most recent graph changelog: https://docs.microsoft.com/en-us/graph/changelog#teamwork they've fixed my issue https://github.com/microsoftgraph/microsoft-graph-docs/issues/6693 wrt to `lastModifiedDateTime`

> Changed lastModifiedDateTime property in the chatMessage resource to represent the time the entity was last touched. It will always be set and never have a null value

So for the replies you can now keep state and see if this property has changed, if so there's a new reply.

matthieu-foucault:
@42wim thanks for the tip, I'll look into using `lastModifiedDateTime` at some point this week

matthieu-foucault:
@42wim sounds like the [Delta query](https://docs.microsoft.com/en-us/graph/api/chatmessage-delta?view=graph-rest-beta&tabs=http) would be very useful, but that doesn't seem to be available in the library we're using. Sounds like the code generator there needs to be fixed too (https://github.com/yaegashi/msgraph.go/issues/22). Any suggestion for the best way to move forward?

42wim:
Well, I've also looked into delta when implementing support for msteams. (I've added the methods myself then)
Delta didn't work, see https://github.com/microsoftgraph/microsoft-graph-docs/issues/6451 and https://github.com/42wim/matterbridge/pull/967

So I'd go with my previous suggestion (or you can try to add delta in a forked library, maybe it's fixed)

DAcodedBEAT (approved):

DAcodedBEAT:
Just pulled this code down and set the PreserveThreading config and it looks like it worked for me!

PR #1250, opened 2020-09-30
nctalk: append a suffix if user is a guest user

gary-kim:

### DIFF
--- a/bridge/nctalk/nctalk.go
+++ b/bridge/nctalk/nctalk.go
@@ -70,6 +70,13 @@ func (b *Btalk) JoinChannel(channel config.ChannelInfo) error {
 		return err
 	}
 	b.rooms = append(b.rooms, newRoom)
+
+	// Config
+	guestSuffix := " (Guest)"
+	if b.IsKeySet("GuestSuffix") {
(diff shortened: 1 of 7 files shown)


42wim:
I'm wondering if this shouldn't be configurable by users, eg that they can choose what to add.
eg `GuestSuffix="(Guest)"` or `GuestString` or something like that :-)

gary-kim:
I like that idea, gimme a bit

s3lph:
While you're at it, would it be possible to add a config option to show the user ID instead of the display name for non-guest users? Use case: The affected Nextcloud gets its user accounts from an LDAP directory with nicknames as UIDs and full names as display names. However, in channels which are bridged to public IRC, I'd prefer to use nicknames rather than full names.

I'd open a pull request myself, but since you're already working on something similar, I thought I'd ask first.

gary-kim:
@s3lph 
I can see how that would be useful. I'd rather not include it into this PR though so I'll get to it after this PR. Feel free to work on it yourself if you'd like, though.

42wim:
:+1:

PR #1252, opened 2020-10-01
nctalk: Add option to use User ID instead of Display Name as nick

s3lph:
As proposed in [#1250](https://github.com/42wim/matterbridge/pull/1250#issuecomment-702180455).

Use case: The Nextcloud instance I intend to use this with gets its user accounts from an LDAP directory with nicknames as UIDs and full names as display names. However, in channels which are bridged to public IRC, I'd prefer to use nicknames rather than full names.

This change only applies to non-guest users, as they are already handled specially, and their UIDs are long hex strings (like 4247797f3cbf4fb2f4ee73995f2a2c683d32284e), which is quite unfit for display.

### DIFF
--- a/matterbridge.toml.sample
+++ b/matterbridge.toml.sample
@@ -1407,6 +1407,9 @@ Password = "talkuserpass"
 
 # Suffix for Guest Users
 GuestSuffix = " (Guest)"
+# If true, use the User ID instead of Display Name for non-guest users
+# WARNING: Before using this, verify that your user IDs are indeed human-readable and intended for display.
+UserIdAsNick = false
 
 ###################################################################
 #
--- a/bridge/nctalk/nctalk.go
+++ b/bridge/nctalk/nctalk.go
@@ -76,6 +76,7 @@ func (b *Btalk) JoinChannel(channel config.ChannelInfo) error {
 	if b.IsKeySet("GuestSuffix") {
 		guestSuffix = b.GetString("GuestSuffix")
 	}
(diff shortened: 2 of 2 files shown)


matterbridge.toml.sample
```diff
@@ -1407,6 +1407,9 @@ Password = "talkuserpass"
 
 # Suffix for Guest Users
 GuestSuffix = " (Guest)"
+# If true, use the User ID instead of Display Name for non-guest users
```
gary-kim, line 1410:
```suggestion
# If true, use the User ID instead of Display Name for non-guest users (default false)
```

gary-kim (approved):
LGTM. I have one suggestion if you want to implement it but your choice :)

42wim:
I've been thinking about this, I think it's better we fix this in a general matter instead of adding yet another config option.

What about we add {USERID} (and {USERNAME} as an alias to {NICK}) as an option in `RemoteNickFormat`?
(and in future also {USERDISPLAYNAME} etc ..)

s3lph:
Come to think of it, this does sound like the more sensible approach. I just had a look at the code, the `nctalk` bridge already uses the same value as `config.Message.UserID`, so this would work. I'll update the PR tomorrow.

s3lph:
I had some more thoughts about how to best implement this: I agree that it would be a good solution to extend the RemoteNickFormat. However, due to the special handling required for nctalk guest users, I think in this case this approach makes more sense.

PR #1272, opened 2020-10-21
Allow tengo to drop messages using msgDrop

42wim:
Can be used to for example drop empty messages

```
if msgText == "blockthismessage" {
   msgDrop=true
}
```


TheHolyRoger:
Thank you @42wim ! Tested and working perfectly, using it now.

42wim:
good to hear :)

Issue #441, opened 2018-05-30
{NICK} Uses Matrix Username, Not Nickname


benrob0329:
**Describe the bug**
When bridging a matrix room, the user's username is displayed on other platforms rather than their nickname.

**To Reproduce**
Bridge matrix with another platform, change your nickname to be different from your username.

**Expected behavior**
Bridge should use the nickname for the {NICK} formatting option.

**Screenshots/debug logs**
Riot: ![screenshot_20180530_135738](https://user-images.githubusercontent.com/19180682/40738560-e46452fa-6411-11e8-96bb-d90b3076c478.png)
Discord: ![screenshot_20180530_135806](https://user-images.githubusercontent.com/19180682/40738568-e7660e80-6411-11e8-8fec-257f070eb088.png)

**Environment (please complete the following information):**
 - OS: Debian 9
 - Matterbridge version: version: 1.10.1 dba259e

**Additional context**
```toml
[discord.unix]                                                                                                                                                                               
Token=redacted                                                                                                                         
Server=redacted                                                                                                                                                                          
WebhookURL=redacted                                               
RemoteNickFormat="{NICK}"                                                                                                                                                                    
                                                                                                                                                                                             
[discord.uofft]                                                                                                                                                                              
Token=redacted                                                                                                                         
Server=redacted                                                                                                                                                                      
WebhookURL=redacted                                                  
RemoteNickFormat="{NICK}"                                                                                                                                                                    
                                                                                                                                                                                             
[discord.bnet]                                                                                                                                                                                                                                                                                                       
Token=redacted                                                                                                                  
Server=redacted                                                                                                                                                               
WebhookURL=redacted                                                
RemoteNickFormat="{NICK} [{PROTOCOL}]"                                                                                                                                                       
                                                                                                                                                                                             
[discord.landfill]                                                                                                                                                                           
Token=redacted                                                                                                                        
Server=redacted                                                                                                                                                                      
WebhookURL=redacted                                                   
RemoteNickFormat="{NICK}"                                                                                                                                                                    
                                                                                                                                                                                             
[xmpp.at]                                                                                                                                                                                    
Server="xmpp.is:5222"
#Jid your userid
Jid=redacted
Password=redacted
Muc=redacted
Nick=redacted
RemoteNickFormat="[{PROTOCOL}] {NICK}: "
ReplaceMessages=[ ["@everyone","@everyone"], ["@here","@here"] ]

[matrix.org]
[...]

PR #1282, opened 2020-11-06
matrix: send the display name instead of the user name

nightmared:
This PR set the default behavior to sending the matrix display name instead of the user name, as asked by #441.

For example, this means '@thobys:server' will be translated to the nickname I chose on `server`.

In addition, if the user didn't set a display name, this will fallback to the user name.

To turn off this behavior, we use the option `UseUserName`, as it seems to be the very meaning this option was intended for in the discord bridge.

### DIFF
--- a/bridge/matrix/matrix.go
+++ b/bridge/matrix/matrix.go
@@ -2,10 +2,7 @@ package bmatrix
 
 import (
 	"bytes"
-	"encoding/json"
-	"errors"
 	"fmt"
-	"html"
 	"mime"
 	"regexp"
 	"strings"
@@ -48,20 +45,23 @@ type matrixUsername struct {
 	formatted string
 }
 
-func newMatrixUsername(username string) *matrixUsername {
-	mUsername := new(matrixUsername)
+// SubTextMessage represents the new content of the message in edit messages.
+type SubTextMessage struct {
+	MsgType string `json:"msgtype"`
+	Body    string `json:"body"`
+}
 
-	// check if we have a </tag>. if we have, we don't escape HTML. #696
-	if htmlTag.MatchString(username) {
-		mUsername.formatted = username
-		// remove the HTML formatting for beautiful push messages #1188
-		mUsername.plain = htmlReplacementTag.ReplaceAllString(username, "")
-	} else {
-		mUsername.formatted = html.EscapeString(username)
-		mUsername.plain = username
-	}
+// MessageRelation explains how the current message relates to a previous message.
+// Notably used for message edits.
+type MessageRelation struct {
+	EventID string `json:"event_id"`
+	Type    string `json:"rel_type"`
+}
 
-	return mUsername
+type EditedMessage struct {
+	NewContent SubTextMessage  `json:"m.new_content"`
+	RelatedTo  MessageRelation `json:"m.relates_to"`
+	matrix.TextMessage
 }
 
 func New(cfg *bridge.Config) bridge.Bridger {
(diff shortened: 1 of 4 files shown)


bridge/matrix/matrix.go
```diff
@@ -216,6 +225,73 @@ func (b *Bmatrix) handlematrix() {
...
+	b.RLock()
+	for k, v := range b.NicknameMap {
+		if now.Sub(v.lastUpdated) > 10*time.Minute {
+			toDelete = append(toDelete, k)
+		}
+	}
+	b.RUnlock()
+
+	b.Lock()
+	if len(toDelete) != 0 {
```
42wim, line 242: you can remove this if, if the len is 0 the for loop will not run

bridge/matrix/matrix.go
```diff
@@ -216,6 +225,73 @@ func (b *Bmatrix) handlematrix() {
 	}()
 }
 
+func (b *Bmatrix) setDisplayName(mxid string, displayName string) string {
```
42wim, line 228: maybe cacheDisplayName is better as function name (just a suggestion)

bridge/matrix/matrix.go
```diff
@@ -216,6 +225,73 @@ func (b *Bmatrix) handlematrix() {
...
+}
+
+func (b *Bmatrix) getDisplayName(mxid string) string {
+	if b.GetBool("UseUserName") {
+		return mxid[1:]
+	}
+
+	b.RLock()
+	if val, present := b.NicknameMap[mxid]; present {
+		if time.Since(val.lastUpdated) < 10*time.Minute {
```
42wim, line 263: We don't need the time check here as you're already cleaning up outdated nicks, you can remove this if?

bridge/matrix/matrix.go
```diff
@@ -216,6 +225,73 @@ func (b *Bmatrix) handlematrix() {
...
+}
+
+func (b *Bmatrix) handleMemberChange(ev *matrix.Event) {
+	// Update the displayname on join messages, according to https://matrix.org/docs/spec/client_server/r0.6.1#events-on-change-of-profile-information
+	if ev.Content["membership"] == "join" {
+		dn, dnPresent := ev.Content["displayname"]
+		dnStr, dnValidString := dn.(string)
+		if dnPresent && dnValidString {
+			b.setDisplayName(ev.Sender, dnStr)
+		}
```
42wim, line 291: not tested, but shouldn't this work also?

```golang
if dn,ok := ev.Content["displayname"].(string);ok {
    b.setDisplayName(ev.Sender, dn)
}
```

42wim (changes requested):

42wim:
@nightmared thanks for you work! 👍 
I've got a few comments

bridge/matrix/matrix.go
nightmared: Oh, I didn't know one could chain the presence check and the cast check like, that's nice ;)

42wim:
looks good, can you rebase as I just merged your previous PR ? :)

42wim (approved):

42wim:
Could you also add the `UseUsername` option to the discord section in matterbridge.toml.sample ?

nightmared:
Hmm, maybe I should move some functions to a new `helpers.go` to make `codeclimate` happy.

42wim:
LGTM, thank you.
As it is a breaking change in behavior I'm going to put it in release 1.20.0 instead of 1.19.1
I'll merge it after 1.19.1 is released.

PR #1283, opened 2020-11-06
Add the {SHORTPROTOCOL} string interpolation for RemoteNickFormat

nightmared:
This new interpolation string for `RemoteNickFormat` allows for building shorter nicknames by only showing the first character in the name of the protocol.
As is noted in the `matterbridge.toml.sample` file, naming conflict can happen between different protocols, so users should take care of not having any conflicting protocols when using that option.

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -340,12 +340,13 @@ func (gw *Gateway) modifyUsername(msg *config.Message, dest *bridge.Bridge) stri
 		nick = strings.Replace(nick, "{NOPINGNICK}", msg.Username[:i]+"​"+msg.Username[i:], -1)
 	}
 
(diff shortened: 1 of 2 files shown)


qaisjp:
I think this could be achieved with a tengo script?

nightmared:
Hmm, you're right, I haven't looked up enough before opening a PR, let's close this.

qaisjp:
Thank you very much for the pull request though :) If we didn't have tengo it would've been perfect :)

nightmared:
Thanks for answering so quickly !
For future reference, here's a possible tengo script that does exactly that (assuming a user didn't put the protocol name in his username, abetter script would be to use last_index and to override the rest of the username :-D):
```
text := import("text")
msgUsername = text.replace(msgUsername, inProtocol, inProtocol[0], 1)
```

PR #1287, opened 2020-11-06
discord: update messages via the webhook instead of deleting messages and creating new ones

### LINKED ISSUE #1259: [Slack->Discord] Msg edit sends new message
Looking through the issues, this sounds related to #588, but the issue insinuates that Slack message edits still sometimes get through. Did this break recently, with the new Slack API changes?

**Describe the bug**
On a Slack<->Discord bridge, edited Slack->Discord message are resent (1 message -> 2 messages). 
Meanwhile, when a Discord->Slack message is edited, it is edited on Slack (1 message -> 1 message).

**To Reproduce**
On a Slack<->Discord bridge,
- send a message on Slack
- edit the Slack message
- observe both the original and edited message on Discord

**Expected behavior**
Having sent (and edited) one message on Slack, there shouldn't be two messages on Discord. The original message should be replaced by the edited message on Discord.

 - OS: linux 64bit
 - Matterbridge version: 1.17.5 8eb5e3cb

**Additional context**
(please tell me if you can't reproduce this and I'll attach more info)

nightmared:
When using the webhook, the previous method to edit a message was to
delete the old one via the classical API, and to create a new message
via the webhook. While this works, this means that editing "old" messages
lead to a mess where the chronological order is no longer respected.

This uses an hidden API explained in https://support.discord.com/hc/en-us/community/posts/360034557771
to achieve a proper edition using the webhook API.

The obvious downside of this approach is that since it is an
undocumented API, there is no stability guarantee at all :/

Do we consider that acceptable ?

fixes #1259

### DIFF
--- a/bridge/discord/discord.go
+++ b/bridge/discord/discord.go
@@ -34,8 +34,6 @@ type Bdiscord struct {
 	membersMutex  sync.RWMutex
 	userMemberMap map[string]*discordgo.Member
 	nickMemberMap map[string]*discordgo.Member
-	webhookCache  map[string]string
-	webhookMutex  sync.RWMutex
 }
 
 func New(cfg *bridge.Config) bridge.Bridger {
@@ -43,7 +41,6 @@ func New(cfg *bridge.Config) bridge.Bridger {
 	b.userMemberMap = make(map[string]*discordgo.Member)
 	b.nickMemberMap = make(map[string]*discordgo.Member)
 	b.channelInfoMap = make(map[string]*config.ChannelInfo)
-	b.webhookCache = make(map[string]string)
 	if b.GetString("WebhookURL") != "" {
 		b.Log.Debug("Configuring Discord Incoming Webhook")
 		b.webhookID, b.webhookToken = b.splitURL(b.GetString("WebhookURL"))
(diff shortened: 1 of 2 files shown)


qaisjp:
"Undocumented endpoints" become "endpoints" by adding it to https://github.com/discord/discord-api-docs

I think this pull request is a reasonable stopgap though before someone goes and adds it to the docs, and then adds it to discordgo, and then updates vendor here.

I'll test this and see how it goes.

nightmared:
I am currently in the process of trying to clean up the mess I made, so to sum it up:
- functions `getCacheID` & `updateCacheID` are no longer required, as the message ID no longer changes
- I am looking at storing the chanID of the last used when using a global webhook, so as to only change the channel via the webhook API when changing channel (as I wonder if I may be rate limited by Discord because of that call). That's my interpretation of the comment "// TODO: this isn't necessary if the last message from this webhook was sent to the current channel" anyway, so I will see if I can make it work.

qaisjp:
> I am looking at storing the chanID of the last used channel when using a global webhook, so as to only use WebhookEdit when really changing channel (as I wonder if I may be rate limited by Discord because of that call). That's my interpretation of the comment "// TODO: this isn't necessary if the last message from this webhook was sent to the current channel" anyway, so I will see if I can make it work.

I'd recommend against this as sometimes it gets out of sync, I'd recommend just deleting that comment.

If you still want to do it, though, I'd recommend doing it in a separate pull request.

Source: I used to do this for my own discord/irc bridge, and sometimes messages would end up in the wrong channel.

nightmared:
Fair enough, I deleted the cache for the channel ID, and I will try to reintroduce that change in another PR, when I will have tested it a bit more to ensure it works well (there was indeed a potential TOCTOU issue if multiple calls to `Send` were executed concurrently, between checking the channel ID and sending the message, I will work on this).

qaisjp:
This works really well! 👍

PR #1288, opened 2020-11-11
xmpp: add NoTLS option to allow plaintext XMPP connections

zhoreeq:
The NoTLS option allows the bridge to work with XMPP servers that do not use TLS, like private intranet servers


42wim:
Thanks!

PR #1308, opened 2020-11-22
Add UserID to RemoteNickFormat and Tengo

### LINKED ISSUE #487: {USERNAME} placeholder in config

JeremyRand:
Fixes #487.  I used `{USERID}` instead of `{USERNAME}` for less ambiguity because the term "Username" is already used in Tengo to refer to the nickname.

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -337,21 +337,21 @@ func (gw *Gateway) modifyUsername(msg *config.Message, dest *bridge.Bridge) stri
 			}
 			i++
 		}
-		nick = strings.Replace(nick, "{NOPINGNICK}", msg.Username[:i]+"​"+msg.Username[i:], -1)
+		nick = strings.ReplaceAll(nick, "{NOPINGNICK}", msg.Username[:i]+"​"+msg.Username[i:])
 	}
 
-	nick = strings.Replace(nick, "{BRIDGE}", br.Name, -1)
-	nick = strings.Replace(nick, "{PROTOCOL}", br.Protocol, -1)
-	nick = strings.Replace(nick, "{GATEWAY}", gw.Name, -1)
-	nick = strings.Replace(nick, "{LABEL}", br.GetString("Label"), -1)
-	nick = strings.Replace(nick, "{NICK}", msg.Username, -1)
-	nick = strings.Replace(nick, "{USERID}", msg.UserID, -1)
-	nick = strings.Replace(nick, "{CHANNEL}", msg.Channel, -1)
+	nick = strings.ReplaceAll(nick, "{BRIDGE}", br.Name)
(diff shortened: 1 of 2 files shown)


JeremyRand:
Not sure what to do about the `wrapperFunc` lint error; I just copied the style of the surrounding code.

42wim:
Thanks!

The linting tells you to use `ReplaceAll` instead of `Replace`, only new code gets linted (as there is too much old code that doesn't pass the linter). Could you change that?

I don't really see how it fixes #487 ? It's still a useful PR though.

JeremyRand:
> The linting tells you to use `ReplaceAll` instead of `Replace`, only new code gets linted (as there is too much old code that doesn't pass the linter). Could you change that?

Okay, I'll do that; might take me a day or two to update the PR.

> I don't really see how it fixes #487 ? It's still a useful PR though.

Maybe I'm missing something -- that issue is about the RemoteNickFormat not being able to access the user ID, no?  I.e. if I have a Matrix user, RemoteNickFormat only can access the display name (e.g. `Jeremy_Rand`) as opposed to the user ID (e.g. `@Jeremy_Rand:matrix.org`).  That's what this PR fixes (the user ID can be accessed independently of the display name).  Am I misunderstanding what that issue is talking about?

42wim:
> Maybe I'm missing something -- that issue is about the RemoteNickFormat not being able to access the user ID, no? I.e. if I have a Matrix user, RemoteNickFormat only can access the display name (e.g. `Jeremy_Rand`) as opposed to the user ID (e.g. `@Jeremy_Rand:matrix.org`). That's what this PR fixes (the user ID can be accessed independently of the display name). Am I misunderstanding what that issue is talking about?

In mattermost the userid is a random ascii string like `wsq4jko6ctyfjybyxnuy6h114y`
I think matrix is one of the few that doesn't have a generated userid.

JeremyRand:
> In mattermost the userid is a random ascii string like `wsq4jko6ctyfjybyxnuy6h114y`
> I think matrix is one of the few that doesn't have a generated userid.

Ah, interesting, I wasn't aware of this.  Should I edit the commit message to remove the reference to that issue, then?  (Matrix is all I need for my purposes, so I'll probably pass on trying to implement Mattermost support for this use case.)

42wim:
I removed the fix for mattermost and squashed your commits, thanks for this PR!

PR #1309, opened 2020-11-23
matrix: show mxids in case of clashing usernames

### LINKED ISSUE #1302: Two Matrix users can use the same display name and Matterbridge makes no distinguishion
**Describe the bug**

Two users at Matrix set the same display name and Matterbridge shows them both the same way.

**To Reproduce**

1. Join a Matterbridged room using two accounts.
2. Run `/myroomnick Nick` as both.
3. Send messages to the room using both accounts.

**Expected behavior**

Generally Matrix clients (Element) resolve this by showing the MXID in brackets after duplciate display names:

* Nick (@account2:example.net)
* Nick (@account1:example.com)

**Screenshots/debug logs**

* image
  * This screenshot could be a bit better as one Element didn't yet notice the other using the same name while clearing cache would likely bring out the brackets due to collision.
* image
* Debug log: N/A

 - OS: Linux
 - Matterbridge version: v1.20.0

**Additional context**

I think this was introduced by https://github.com/42wim/matterbridge/pull/1282, CC: @nightmared. Otherwise thank you for making Matrix users more readable through Matterbridge, this has been waited for 👍🏻

nightmared:
Fixes #1302.

### DIFF
--- a/bridge/matrix/helpers.go
+++ b/bridge/matrix/helpers.go
@@ -3,6 +3,7 @@ package bmatrix
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"html"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


bridge/matrix/helpers.go
```diff
@@ -82,22 +83,39 @@ func (b *Bmatrix) getDisplayName(mxid string) string {
...
+
+	b.Lock()
 	for k, v := range b.NicknameMap {
+		// to prevent username reuse across matrix servers - or even on the same server, append
+		// the mxid to the username when there is a conflict
+		if v.displayName == displayName {
+			conflict = true
+			// TODO: it would be nice to be able to rename previous messages from this user.
+			// The current behavior is that only users with clashing usernames and *that have spoken since the bridge last started* will get their mxids shown, and I don't know if that's the expected behavior.
+			v.displayName = fmt.Sprintf("%s (%s)", displayName, v.mxid)
```
42wim, line 99: `k` is already the mxid, you don't need to use `v.mxid`
I'd rewrite the for loop to `for mxid,v := range b.NicknameMap...`

bridge/matrix/helpers.go
```diff
@@ -82,22 +83,39 @@ func (b *Bmatrix) getDisplayName(mxid string) string {
...
+		displayName = fmt.Sprintf("%s (%s)", displayName, mxid)
+	}
+
 	for _, v := range toDelete {
 		delete(b.NicknameMap, v)
 	}
+
 	b.NicknameMap[mxid] = NicknameCacheEntry{
 		displayName: displayName,
+		mxid:        mxid,
```
42wim, line 118: not needed, as mxid is the key

bridge/matrix/matrix.go
```diff
@@ -22,6 +22,7 @@ var (
 
 type NicknameCacheEntry struct {
 	displayName string
+	mxid        string
```
42wim, line 25: not needed as mxid is the key

42wim:
I think we can simplify it a bit

42wim:
Thanks!

PR #1319, opened 2020-11-30
Refactor guild finding code (discord)

qaisjp:


bridge/discord/discord.go
```diff
@@ -94,29 +94,48 @@ func (b *Bdiscord) Connect() error {
...
 	b.channelsMutex.Unlock()
-	if !guildFound {
-		msg := fmt.Sprintf("Server \"%s\" not found", b.GetString("Server"))
-		err = errors.New(msg)
-		b.Log.Error(msg)
-		b.Log.Info("Possible values:")
+
+	// If we couldn't find a guild, we print extra debug information and return a nice error
+	if b.guildID == "" {
+		err = errors.Errorf("Discord server %#v not found", b.GetString("Server"))
```
42wim, line 124: can you make this `err = fmt.Errorf("Discord server %#v not found", b.GetString("Server"))` so that we don't pull in a new package just for this.

bridge/discord/discord.go
```diff
@@ -94,29 +94,48 @@ func (b *Bdiscord) Connect() error {
...
-			guildFound = true
+		// Skip, if the server name does not match the visible name or the ID
+		if guild.Name != serverName && guild.ID != serverName {
+			continue
+		}
+
+		// Complain about an ambiguous Server setting. Two Discord servers could have the same title!
+		// For IDs, practically this will never happen. It would only trigger if some server's name is also an ID.
+		if b.guildID != "" {
+			return errors.Errorf("There are multiple Discord servers with the same name %#v, expected to see only one", serverName)
```
42wim, line 109:
```suggestion
			return fmt.Errorf("There are multiple Discord servers with the same name %s, expected to see only one", serverName)
```

42wim: this way we don't need to pull in another package

bridge/discord/discord.go
```diff
@@ -94,29 +94,48 @@ func (b *Bdiscord) Connect() error {
...
 		}
+
+		// Getting this guild's channel could result in a permission error
+		b.channels, err = b.c.GuildChannels(guild.ID)
+		if err != nil {
+			return errors.Wrapf(err, "could not get %#v's channels", b.GetString("Server"))
```
42wim, line 115:
```suggestion
			return errors.Wrapf(err, "could not get %s's channels", b.GetString("Server"))
```

qaisjp:
That package's `errors.Wrap` function is used frequently in my `feature/transmitter` branch: https://github.com/42wim/matterbridge/compare/master...qaisjp:feature/transmitter.

And this package is actually already pulled in by multiple things we depend on, and this just makes it available inside matterbridge:

https://github.com/42wim/matterbridge/blob/ad95e356877babdf38953a521f54b303e108d490/go.sum#L665-L669

Do you still want me to eliminate this dependency?

qaisjp:
Hmm, looks like there's a "new" way of wrapping errors: https://blog.golang.org/go1.13-errors

I'll move the code I've written to do what that says, and get rid of github.com/pkg/errors

bridge/discord/discord.go
qaisjp: `%#v` will format strings with proper quoting and escaping, so it looks like `with the same name "Server Name Here", expected to see only one`, instead of `with the same name Server Name Here, expected to see only one`. Are you sure you want `%s`?

42wim: No good point for %#v 👍

42wim:
wrt pkg/errors, if you need it you can use it but I think there was no reason to use it here for what you are/were using it for :)

PR #1333, opened 2020-12-09
nctalk: update go-nc-talk

gary-kim:
Update go-nc-talk to fix a few bugs.


gary-kim:
Also, what is the standard way to report a fatal error in a bridge? There is now an error report for access to the room being lost but I'm not sure if that should quit Matterbridge or just error.

42wim:
No, we don't expect fatal errors ;-)
At the moment, does only get checked at startup and will disable the bridge, at runtime there is no option.
I think the best thing to do (at the moment) is to log the error. If access to the room is being lost, this is something the owner probably has done himself?

PR #1345, opened 2020-12-31
Reject cross-channel message references (discord)

qaisjp:
Discord message references have been designed in a way for this to
support cross-channel or even cross-guild references in the future.

This will ensure the ParentID is *not* set when the message refers to a
message that was sent in a different channel.


42wim (approved):

Gothicania:
> Discord message references have been designed in a way for this to
> support cross-channel or even cross-guild references in the future.
> 
> This will ensure the ParentID is _not_ set when the message refers to a
> message that was sent in a different channel.

Gothicania:
commit b657252eb31d80ffe14cbf9c028550b79645a60c

PR #1347, opened 2020-12-31
Refactor "msg-parent-not-found" to config.ParentIDNotFound

qaisjp:


42wim:
Any reason for  `ParentNotFoundID` instead of  `ParentIDNotFound` ?

qaisjp:
I didn't like `ID` and `Not` smushing together to `IDN` 🙂 

Imo both works:

- `ParentNotFoundID`, "this is _the ID_ when the parent is not found" 
- `ParentIDNotFound`, "this is _[the value]_ when the parent _ID_ is not found"

I can change it if you prefer `ParentIDNotFound`.

42wim:
> Imo both works:
> 
> * `ParentNotFoundID`, "this is _the ID_ when the parent is not found"

I don't like this because you expect it to be an ID and it's not really an ID :)

> * `ParentIDNotFound`, "this is _[the value]_ when the parent _ID_ is not found"
> 
> I can change it if you prefer `ParentIDNotFound`.

Yes please :)

42wim (approved):

PR #1352, opened 2021-01-04
[Dockerfile] Unify/sync apk index cache control

PeterDaveHello:
The second stage is using a single `apk` command with `--no-cache` which is better.


42wim:
thanks 👍

Issue #1357, opened 2021-01-09
Telegram Image bridge caption missing (fixable)


Humorhenker:
**Is your feature request related to a problem? Please describe.**
The telegram Bridge Handler does not add Captions to images it uploads. The telegram go libary used is able to do that but it would need a little tweak to expose the option to us. As the bridge just sends an extra message with the caption after the image it is no problem for images with captions but if an image got no caption no one is able to tell who send it in the first place because there is no Nick added (as there is no caption).

**Describe the solution you'd like**
In handleUploadFile (bridge/telegram/handlers.go) tgbotapi.NewPhotoUpload is called. This function does not feature the caption parameter which it could if we would change go-telegram-bot-api/telegram-bot-api/helpers.go to allow that.
As go-telegram-bot-api/telegram-bot-api/helpers.go seems to be just a mirror for an external dependency it would not be wise to just edit it i suppose so how would i go about making my change?

Greetings.

Humorhenker:
After some more investigation i realized that we dont need to edit the tgbotapi dependency at all. I created a somewhat hacky edit to add the captions inline with the photo. Which would add the username even if there is not caption in the first place. See pr #1358

PR #1358, opened 2021-01-11
Adding caption to send telegram images see issue #1357

Humorhenker:
As you can see this is a fairly hacky solution. Someone propably got a better way. Just wanted to start


Humorhenker:
Some similar fix would be needed for all other mediatypes as well so maybe it would nice to just include them in one commit. But we first gotta figure out who to do it the right way.

Humorhenker:
I have added bit more code to also handle other telegram media types. CodeQL got some minor quality points which i will improve but overall i would say that should fix the problem now

42wim:
Thank you! (sorry for the delay)

PR #1360, opened 2021-01-12
Add jpe as valid image filename extension for telegram

Humorhenker:
When bridging from Whatsapp to Telegram images are sent as files by the telegram bot. That seems to be the case because they have a .jpe fileextension which is a valid jpeg fileextension. So the bridge should recognize them as images.


42wim:
👍

PR #1395, opened 2021-02-09
Handle Rocket.Chat attachments

tadeokondrak:
Continuing on from https://github.com/42wim/matterbridge/pull/756's code.

### DIFF
--- a/bridge/helper/helper.go
+++ b/bridge/helper/helper.go
@@ -51,6 +51,30 @@ func DownloadFileAuth(url string, auth string) (*[]byte, error) {
 	return &data, nil
 }
 
+// DownloadFileAuthRocket downloads the given URL using the specified Rocket user ID and authentication token.
+func DownloadFileAuthRocket(url, token, userID string) (*[]byte, error) {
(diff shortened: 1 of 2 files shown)


42wim:
Hi @tadeokondrak thank you for working further on the previous PR.

Some remarks:
- vendored code can not be changed, for your rocketchat changes you'll need to open a PR against https://github.com/matterbridge/Rocket.Chat.Go.SDK (branch matterbridge)
- what exactly are those vendored mattermost changes needed for?

tadeokondrak:
I've removed the vendored code changes. One of them was completely useless because I mixed things up, and the other was fixing the `title_link_download` flag for this to work:
```go
	if !file.TitleLinkDownload {
		b.Log.Infof("File is not intended to be downloaded.")
		return nil
	}
```
From what I can tell, this means that clicking on the URL in Rocket.Chat will download the file, but I don't think attachments without this flag need to be ignored, since they are still shown in the channel.
I've removed the check of this flag along with the other changes.

42wim:
Thanks!

PR #1399, opened 2021-02-11
Add IgnoreMessageNotContains Option

realPy:
A little improvement: We can ignore message contains but its not possible to ignore message not contains :)  
For my use, i want only forward message begin with some tags


42wim:
Thanks for your PR, but this sort of thing should be handled by tengo scripts.
Feel free to submit a PR with a tengo script for the contrib directory !

See https://github.com/42wim/matterbridge/wiki/Tengo for more information.

PR #1401, opened 2021-02-14
Fix matterbridge crashing when using the XMPP webhook

Polynomdivision:
I recently deployed matterbridge with the XMPP webhook enabled on my server, but whenever someone wrote a message in a bridged channel that is not XMPP, the bridge would crash.

For me, it turned out that it was an invalid certificate. Nonetheless, matterbridge should tell you in the logs why it was not able to
POST the webhook.

This PR adds a log entry with the POST failure reason and prevents the bridge from "crashing" the bridge when a POST fails.


42wim:
👍

PR #1402, opened 2021-02-14
Use valid transmitter Log default (discord)

qaisjp:
Using a logger created by `log.NewEntry(nil)` would crash. (matterbridge does not encounter this issue as it updates the Log field manually.)


qaisjp:
Thanks!

PR #1405, opened 2021-02-17
Fix webhooks for channels with special characters

Polynomdivision:
During my deployment of the XMPP webhook, I noticed that certain channels were not getting messages. Turns out, that that channel has a '#' in its URI, meaning that the webserver of prosody was interpreting it as an URL fragment and thus ignoring it.

This PR encodes the channel name to prevent such behaviour.


42wim:
Thank you!

PR #1406, opened 2021-02-17
Change workflow from go 1.16.0-rc1 to go 1.16.x

qaisjp:
https://golang.org/doc/devel/release.html#go1.16 released yesterday, and now checks are failing because the rc is no longer available https://github.com/42wim/matterbridge/runs/1920474638


42wim:
thanks!

Issue #1381, opened 2021-01-30
XMPP media is presented as link


allilengyi:
**Describe the bug**
When sending an image from protocol A to xmpp then one receives a clickable link but there is no media offered to download in the client. Sure, when clicking on the link, I can view the image in the browser. 

**Expected behavior**
I expect a file offered to download in xmpp clients

**Screenshots/debug logs**
Screenshot from XMPP client Dino
![Bildschirmfoto vom 2021-01-30 23-18-19](https://user-images.githubusercontent.com/47537028/106369421-79174100-6351-11eb-9efe-c51cc1e30606.png)

Relevant part from the matterbridge log
```
[0911] DEBUG xmpp:         [handleXMPP:bridge/xmpp/xmpp.go:237] == Receiving xmpp.Chat{Remote:"xr_berlin_events@conference.jabber.systemli.org/XR Berlin Events", Type:"groupchat", Text:"https://xmpp-media.extinctionrebellion.de/xrberlinevents/afe34065/file_10.jpg", Subject:"", Thread:"", Ooburl:"", Oobdesc:"", ID:"0a0484bdf90bd9c5", ReplaceID:"", Roster:xmpp.Roster(nil), Other:[]string{}, OtherElem:[]xmpp.XMLElement(nil), Stamp:time.Time{wall:0x0, ext:0, loc:(*time.Location)(nil)}}
```

**Environment (please complete the following information):**
 - OS: debian 9
 - Matterbridge version: 1.21.0 cf13fff7

**Additional context**
Might this be an issue with the apache config? Permissions? ports? I checked that file sending/receiving works with the xmpp server that is used to bridge. 
Matterbridge config file.
```
[general]
#IgnoreFailureOnStart=true
MediaDownloadPath="/var/www/matterbridge/xrberlinevents"
MediaServerDownload="https://xmpp-media.extinctionrebellion.de/xrberlinevents"
MediaDownloadSize=5000000

[xmpp.xr_berlin_events]
Server="jabber.systemli.org:5222"
#Jid your userid
Jid="xr_berlin@jabber.systemli.org"
Password=""
Muc="conference.jabber.systemli.org"
Nick="XR Berlin Events"
RemoteNickFormat=""

[telegram.tgbot]
#See https://core.telegram.org/bots#6-botfather 
#and https://www.linkedin.com/pulse/telegram-bots-beginners-marco-frau
Token=""
RemoteNickFormat=""
MessageFormat="HTMLNick"

[...]

PR #1410, opened 2021-02-24
Make function SendOOB in vendor/github.com/matterbridge/go-xmpp/xmpp.…

allilengyi:
Resolve issue 42wim#1381.
Make function SendOOB in vendor/github.com/matterbridge/go-xmpp/xmpp.go to send meesage *with* actual body according to https://github.com/42wim/matterbridge/issues/1381#issuecomment-774032477

Remove the appendage `":" + fileInfo.URL` in bridge/xmpp/xmpp.go. This will not modify the message so that the link to the file is not sent in the text message. Instead the media file is available as inline display in modern XMPP clients.
Moreover, send the media attachement first, then the message.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,7 @@ And more...
 ### 3rd party via matterbridge api
 
 - [Discourse](https://github.com/DeclanHoare/matterbabble)
+- [Facebook messenger](https://github.com/powerjungle/fbridge-asyncio)
 - [Facebook messenger](https://github.com/VictorNine/fbridge)
 - [Minecraft](https://github.com/elytra/MatterLink)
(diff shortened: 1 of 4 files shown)


allilengyi:
Please discard all changes in bridge/xmpp/xmpp.go except those from line 300-400. These are not intended.

Humorhenker:
You will have to discard all non intended changes yourself but you can just update your fork and this pull request will update aswell. Be sure to edit the libraries back to the original.

42wim:
We can't change vendored code here, for changes against go-xmpp please open a PR on https://github.com/matterbridge/go-xmpp against the `matterbridge` branch

PR #1418, opened 2021-03-01
rudamentary support for Signal protocol via signald

petermolnar:
Signald <https://gitlab.com/signald/signald> is a standalone Signal client written in Java that can be interacted with via Unix socket and newline separated JSON strings.

These commits establish a crude and basic interaction with the signald daemon where by giving a group ID, one can relay messages from and to a Signal group.

Anything beyond this functionality, including registration, device linking, joining a group, etc. is missing.

To test the functionality one has to set up signald with linking or registration in advance.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "vendor/gitlab.com/signald/signald-go"]
+	path = vendor/gitlab.com/signald/signald-go
+	url = https://gitlab.com/signald/signald-go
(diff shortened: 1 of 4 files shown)


42wim:
@petermolnar why did you close this?

petermolnar:
It failed every test possible, so I thought I'd revisit my changes first, then open a pull request that might actually work.

petermolnar:
Truth is, I should have kept it open :/

PR #1428, opened 2021-03-16
discord: declare GUILD_MEMBERS privileged intent (fix nick tracking after connect)

jlu5:
Without this declared, it seems that Discord will not send any member update
events after connection, even if the privileged gateway intent is enabled for
the bot in settings. This causes nick tracking to get out of sync when people
change their nicks after the bot connects.

See: https://discord.com/developers/docs/topics/gateway#gateway-intents


42wim:
Thanks!

PR #1438, opened 2021-03-29
Add possibility for using MxId/Token with Matrix

### LINKED ISSUE #1429: Support Matrix login via token
Matrix homeservers with SSO will possibly have password login disabled. Using a pre-created access token would make using Matterbridge possible in these cases as well. Currently the only option is username/password.

jaywink:
Add possibility for using MxId/Token with Matrix

Makes it possible to configure a Matrix bot to use Matrix ID + Access token instead of username/password. This makes it possible to use the bot in environments where password login is disabled (for example SSO environments).

Matrix user ID's are commonly referred to as "MXID's". I thought about (ab)using "Login" here but it felt like a bad idea given it's used as "username" for the password login. None of the other configuration items felt fitting.

Closes #1429


jaywink:
Not sure what to do about the linting failure, doesn't feel related to the PR, unless I'm missing something?

bridge/matrix/matrix.go
```diff
@@ -75,22 +75,30 @@ func New(cfg *bridge.Config) bridge.Bridger {
...
-		User:       b.GetString("Login"),
-		Password:   b.GetString("Password"),
-		Identifier: matrix.NewUserIdentifier(b.GetString("Login")),
-	})
-	if err != nil {
-		return err
+	if b.GetString("MxID") != "" && b.GetString("Token") != "" {
+		b.mc, err = matrix.NewClient(
+			b.GetString("Server"), b.GetString("MxID"), b.GetString("Token"),
+		)
```
42wim, line 81: The linting problem is because you don't handle the error, you need to add an
```
if err != nil {
         return err
}
```
So that we fail when the NewClient method fails.

42wim (changes requested):

42wim (approved):

42wim:
Thanks! sorry for the delay.

PR #1444, opened 2021-04-02
Add MatterBukkit and Forge / Bukkit distinction

Millesimus:
Just would like to add MatterBukkit to the list. It works great, but it took me a little to find it, since MC Forge is more prominent,


42wim:
Thanks :+1:

PR #1462, opened 2021-05-05
discord: Add AllowMention to restrict allowed mentions

### LINKED ISSUE #476: disallow @everyone and @here in discord
`@everyone` and `@here` are not configurable to be filtered when sent to discord
a bot user can be configured to being disallowed to ping everyone,
but there is no such setting when using webhooks

confiig entries
```toml
allowPingEveryone=false # default false
allowPingHere=false # default true
allowPingUser=false # default true
```

also related to #460 for translating user mentions

possibly also 
```toml
IgnorePing="someUser someOther Bot.+"
```
similar to `IgnoreMessages`

I have tried to filter it through `IgnoreMessages`, but that does not work for messages coming in from api endpoints
and there is no general setting for `IgnoreMessages` that that applies to every platform

alexandregv:
As discussed in #476, since Discord now has a [native way to disable mentions for both bots and webhooks](https://discord.com/developers/docs/resources/channel#allowed-mentions-object), it is probably better to use this instead of `ReplaceMessages`.  

Benefits:
- Safer handling: this is handled by Discord itself and cannot be dodged by any trick, which `ReplaceMessages` can be
- Better display: with this method, mentions are still nicely displayed and clickable, but do not send a ping to users

I added it in a way to be compatible with existing configs, so the behavior remains the same by default (all mentions allowed).

If this PR is accepted, the [Wiki page](https://github.com/42wim/matterbridge/wiki/Section-Discord-%28basic%29) should be updated with the new `matterbridge.toml.sample` file. I can do it if needed.

Fixes #476


bridge/config/config.go
```diff
@@ -85,91 +85,94 @@ type ChannelMember struct {
...
+	BindAddress             string // mattermost, slack // DEPRECATED
+	Buffer                  int    // api
+	Charset                 string // irc
+	ClientID                string // msteams
+	ColorNicks              bool   // only irc for now
+	Debug                   bool   // general
+	DebugLevel              int    // only for irc now
+	DisablePingEveryoneHere bool   // discord
+	DisablePingRoles        bool   // discord
+	DisablePingUsers        bool   // discord
```
haykam821, line 98: I think it would better to keep this in line with the documentation's terminology, and invert the value accordingly:

```suggestion
	ParseEveryoneMentions bool   // discord
	ParseRoleMentions        bool   // discord
	ParseUserMentions        bool   // discord
```

alexandregv: That's what I did first, but I then inverted it so that the default bool value (false) reflect the current behavior (all allowed), for compatibility.
I guess I can get it working with config.IsKeySet()

alexandregv: Resolved by https://github.com/42wim/matterbridge/pull/1462/commits/cb6d9598ed78c5cae77bcc10683ce50d4a0e60e3

haykam821 (approved):

qaisjp:
Can we replace `EveryoneHere` with just `Everyone` — it's more consistent with Discord API also more likely to be less confusing across platforms

42wim:
@alexandregv thank you!
I'm not a fan of introducing yet again 3 bools though, what do you think if we make it an array? Also maybe call it `AllowMention`.

eg:
```
AllowMention=["user","role"]
```

alexandregv:
@42wim That's indeed better with an array, and reflects how Discord API handles it. I changed it in that way.

42wim:
@alexandregv perfect, thank you very much. 👍 
If you still want to update the wiki it'd be much appreciated :)

alexandregv:
Thanks for merging!
I just edited the [Wiki](https://github.com/42wim/matterbridge/wiki/Section-Discord-(basic)) :)

PR #1487, opened 2021-05-23
Add a MessageClipped option to change clipped message. Closes #1359

### LINKED ISSUE #1359: Allow changing or disabling the "\<clipped message\>" message on long lines

I don't like seeing the "\<clipped message\>" message as I think it can mostly be figured out from context and it's so long as opposed to something like `…`.

I would like to be able to change or disable the "\<clipped message\>" message like changing the edit message is possible with the `EditSuffix` option.

I haven't thought of other alternatives.

**Additional context**

This might also be regression of `MessageSplit=true` having broken at some point judging by https://github.com/42wim/matterbridge/issues/281#issuecomment-346903684.

I am running Matterbridge 1.21.0 cf13fff7 and am also using `EditSuffix=" ✏️"` (which inspires me to try to change it).

42wim:


xnaas:
I think this one still needs to be added to the wiki, @42wim?

42wim:
@xnaas yes feel free to add it please :-)
How it works is you just set the text it should display instead of the default one.

`MessageClipped="<toolong>"`

PR #1488, opened 2021-05-24
[whatsapp] make DocumentMessage handler use FileName attribute.

AvinashReddy3108:
referenced: https://github.com/Rhymen/go-whatsapp/blob/master/message.go#L582


42wim:
Thanks!

PR #1496, opened 2021-05-27
matrix: content does not have body for redactions

gary-kim:
While testing, I noticed that deletions from Matrix were running into an error. It seems to be because `m.room.redaction` does not have `content.body` according to the spec but the Matterbridge implementation expects it.

Redaction spec here: https://matrix.org/docs/spec/client_server/r0.6.1#m-room-redaction

This change makes it not required for redactions.


42wim:
Thanks!

PR #1506, opened 2021-06-02
nctalk: separate display name

gary-kim:
This adds support to add a separate display name on the Nextcloud Talk side. At the moment, this only works when configured via Nextcloud Talk.

Works with: https://github.com/nextcloud/spreed/pull/5667


gary-kim:
@42wim Could you take a look?

42wim:
Sorry for the delay was off my radar.

As this is a setting that needs to be enabled on the nextcloud side, is there any way to have it check nextcloud? So that it'll be automatically active if it's enabled? (without adding an extra option in the matterbridge config)

gary-kim:
This seemed to be the most reliable way to go about it. Nextcloud actually generates the config if support for it is enabled so if it is supported, it should be in the config automatically.

42wim:
Thanks!

PR #1541, opened 2021-07-08
Matrix: use per-room nicknames

### LINKED ISSUE #1336: [Matrix] Matterbridge always uses the latest room specific display name everywhere
**Describe the bug**

When I change my room display name in one room, I suddenly see it as my name through the Matterbridge instance everywhere.

**To Reproduce**

1. Bridge two Matrix rooms
2. `/devtools`, explore room state, `m.room.id`, change your display name, send the event
    * it may also be possible to use `/myroomnick`
3. start talking in another room
4. observe your name having changed there too through matterbridge.

**Expected behavior**

I expect to see the original name in where I didn't change my name.

**Screenshots/debug logs**

I cannot debug logs now.

 - OS: Debian 10 GNU/Linux
 - Matterbridge version: https://github.com/42wim/matterbridge/releases/tag/v1.20.0 (I cannot SSH right now)

**Additional context**

* I am assuming Element is being used as the Matrix client.
* This may also affect avatars, but I don't have a Discord where I would see that.
* If I am missing something important from this issue to help with reproduction, please let me know and I will add it ASAP.

nightmared:
Fixes #1336

### DIFF
--- a/bridge/matrix/helpers.go
+++ b/bridge/matrix/helpers.go
@@ -51,17 +51,19 @@ func interface2Struct(in interface{}, out interface{}) error {
 	return json.Unmarshal(jsonObj, out)
 }
 
-// getDisplayName retrieves the displayName for mxid, querying the homserver if the mxid is not in the cache.
-func (b *Bmatrix) getDisplayName(mxid string) string {
+// getDisplayName retrieves the displayName for mxid, querying the homeserver if the mxid is not in the cache.
+func (b *Bmatrix) getDisplayName(channelID string, mxid string) string {
 	if b.GetBool("UseUserName") {
 		return mxid[1:]
 	}
 
 	b.RLock()
-	if val, present := b.NicknameMap[mxid]; present {
-		b.RUnlock()
+	if channel, channelPresent := b.NicknameMap[channelID]; channelPresent {
+		if val, present := channel[mxid]; present {
+			b.RUnlock()
 
-		return val.displayName
+			return val.displayName
+		}
 	}
 	b.RUnlock()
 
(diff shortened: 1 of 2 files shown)


42wim:
Not sure how this works? There's only one global displayname?
https://matrix.org/docs/spec/client_server/r0.6.1#get-matrix-client-r0-profile-userid-displayname

Seems like the per-room displayname is still being worked on - https://github.com/vector-im/element-web/issues/2458 / https://github.com/matrix-org/matrix-doc/pull/3189

nightmared:
In element you can already update your per-room nick with the commande `/myroomnick`.

42wim:
Sure, but the API call here is fetching the global nick `mc.GetDisplayName`, there is no way to fetch the nick per room using the matrix API ?
Or am I missing something?

nightmared:
Hmm yeah I'm not really sure, I guess this is fixed by getting the state when syncing:
```
GET /_matrix/client/r0/sync

Note: This endpoint supports lazy-loading. See Filtering for more information. Lazy-loading members is only supported on a StateFilter for this endpoint. When lazy-loading is enabled, servers MUST include the syncing user's own membership event when they join a room, or when the full state of rooms is requested, to aid discovering the user's avatar & displayname.
```
I will have to check that it also works when display names are changed while matterbridge is stopped.

nightmared:
So, from what I understand, a server send the update when a user changes its displayname (as per `https://matrix.org/docs/spec/client_server/r0.6.1#events-on-change-of-profile-information`), and the server also sends that information when the bridge starts, sending all the events since that last MarkRead() (I guess that's the point where it starst sending events again). And we use that information in `handleMemberChange` to update the displayname for the given room.

42wim:
For the current code (and your PR) this means that there's a 10 minute delay before the nickname also changes in the other rooms because we remove the cache every 10 minutes. And we'll ask for the nickname again (using the matrix API call)

https://github.com/42wim/matterbridge/blob/23e785deb98b2673ea193b939d244b3d4b3e2dde/bridge/matrix/helpers.go#L106-L108

So I don't think this fixes the issue ?

nightmared:
Oh yeajh, I haven't tried over sessions running long enough to hit that case. You're right, I will have to rethink this code. I'm not sure yet this is simple to do in matterbridge because there is no data persistence across restarts.

nightmared:

Issue #1187, opened 2020-08-03
Matrix: Support for encrypted chatrooms


BenWiederhake:
The bridge currently doesn't support encryption, because gomatrix doesn't support it, and doesn't plan to support it anytime soon: https://github.com/matrix-org/gomatrix/issues/31#issuecomment-565988610

By now, Rooms on Matrix are encrypted by default, which is why I "reopen" the old issue (#170).

Maybe it's a good idea to switch to mautrix-go, because [it supports E2EE](https://github.com/tulir/mautrix-go/issues/3), and it's based on gomatrix (so hopefully migration will be easy).

I could try to migrate it myself, but before I put any effort into that I'd first like to hear what @42wim thinks about this.

In case anyone is interested, this is the current situation:
- Sending to an unencrypted Matrix room works fine
- Receiving from an unencrypted Matrix room works fine
- Sending to an encrypted Matrix room works, but the message is displayed as "(!) Unencrypted" to other Matrix users
- Receiving from an encrypted Matrix room doesn't work, `./matterbridge -debug` shows no output whatsoever.

qaisjp:
Have you tried setting this up? https://github.com/matrix-org/pantalaimon

Not convenient that it uses Python, though.

TheHolyRoger:
@BenWiederhake have you done any work on this yet? would be happy to move to mautrix-go since gomatrix aren't puling their finger out

BenWiederhake:
I haven't, because I haven't heard anything from go-matrix. I'd be very happy if you can do this.

TheHolyRoger:
I had a look last night and it really doesn't look too difficult, and I'm a GO noob... @42wim any thoughts on this? I'd be happy to help out, I learn quick :)

42wim:
@TheHolyRoger i'm open for PR's to fix this, I don't mind going to mautrix-go but it should also support the extra stuff I've added to matrix-go.
I don't have the time at the moment to do this myself.

Mikaela:
If I understand this blog post correctly, Matrix-org itself isn't using go-matrix anymore having switched to mautrix-go.

* https://matrix.org/blog/2020/09/08/gsoc-report-e-2-e-encryption-for-go-neb

BenWiederhake:
Excellent! Even more reason to switch to mautrix-go.

unode:
Keeping an eye on this as we would be interested in having support for encrypted rooms. Thanks to everyone making it possible.

[...]

Issue #137, opened 2017-03-26
Use matrix Application Service API (matrix)


42wim:
Look into using matrix application service api for the matrix bridge.
(https://matrix.org/docs/guides/application_services.html)

t3chguy:
I believe we have no go helper for Application Services yet, but gomatrix iirc includes the AS extension to the CS API so would just need a http server listening on specific routes to receive the pushed events from Synapse.

xludx:
Could this be of any help: https://github.com/tulir/matrix-appservice-go

benrob0329:
Perhaps using something like [matrix-appservice-webhooks](https://github.com/turt2live/matrix-appservice-webhooks) could be better.

MayhemBill:
Has there been any update to this?

42wim:
@MayhemBill no, I think the number of users who would actually use this would be rather small, you'll need to setup your own homeserver for this. So it's not priority for me at the moment.

But if someone sends me a PR that adds support for this via eg https://github.com/tulir/mautrix-appservice-go , I will accept it gladly.

TheHolyRoger:
Is using swagger not a better approach? I might look into this.

https://github.com/go-swagger/go-swagger
https://matrix.org/docs/api/client-server/

t-my:
Application API requires hosting matrix homeserver and accessing api's directly. Remote access is forbidden for security reasons so this might be unfeasible.

w3bb:
With Dendrite, a small homeserver entering beta, and various lightweight implementations supporting the appservice API now this might be more worth it. You can easily run a small bridging server on a 5 buck VPS.

w3bb:
I might give this a shot

hellcp:
This would be really useful for us

PR #1543, opened 2021-07-09
Matrix: switch to the mautrix-go library

nightmared:
Replaces https://github.com/matrix-org/gomatrix with https://github.com/tulir/mautrix-go as it appear to be the successor (see https://github.com/42wim/matterbridge/issues/1187#issuecomment-688986854).

This could be a stepping stone for #137 and #1187.

Note that for reducing rebases this currently depends of #1541, however I will happily change that if #1541 is rejected.

Hopefully one day we will have encryption support with appservices for matrix ;)


nightmared:
FYI I'm currently looking at adding a very basic appservice support for matterbridge using this patch, you can see what it looks like at https://github.com/nightmared/matterbridge/tree/mautrix_appservice.

TheHolyRoger:
@nightmared you bloody beast thank you! I can help review and test this, I did briefly try to tackle it before.

Does this support encryption already?

nightmared:
I haven't yet tried it against encrypted room, there is probably more work to be done to support it (like some mechanism to store keys across sessions maybe ? I have to admit I don't know much of the way keys are managed in the matrix world, I would have to take a look. Do you know about it maybe ?)

42wim:
Hi @nightmared thanks for your work! I'll try to review it this week

poVoq:
Would be great to have this merged *sorry for nagging*

nightmared:
I think I'm going to work on https://github.com/42wim/matterbridge/pull/1544 on the next few evenings (that draft PR depends on this PR), which will offer me the occasion to test this a bit further.
In any case, if you have tested this and it worked (and even more if it didn't), please share ;)

w3bb:
> I haven't yet tried it against encrypted room, there is probably more work to be done to support it (like some mechanism to store keys across sessions maybe ? I have to admit I don't know much of the way keys are managed in the matrix world, I would have to take a look. Do you know about it maybe ?)

A lot of bots just use [Pantalaimon](https://github.com/matrix-org/pantalaimon/).

nightmared:

PR #1544, opened 2021-07-10
WIP: Matrix: support for appservice

nightmared:
Builds on top of the work done in #1543.

It seems to work for me, in a discord<->matrix setting.

To reproduce, here is my configuration:
```
$ cat appservice_synapse.yml
id: "Matterbridge"
url: "http://192.168.1.13:1234"
as_token: "<RANDOM HERE>"
hs_token: "<RANDOM HERE>"
sender_localpart: "_bridge_bot"
namespaces:
  users:
    - exclusive: true
      regex: "@__bridge_.*"
  aliases:
    - exclusive: false
      regex: "#__bridge_.*"
  rooms: []
```

Note that the `@__bridge_` part is currently hardcoded in the source of matterbridge, I haven't devised a smart way to extract that information properly from the namespaces.users.regex yet.

I have a room (hosted on another matrix server, so that works withotu any change, which is a lot nicer than the pain I expected) with a local alias that matches the `#__bridge_' I defined namespaces.aliases.regex (that part isn't hardcoded, you should be able to change it).

You can see an example here: https://nightmared.fr/files/2021-07-10-1625933690.png

Finally my matterbridge configuation:
```
[matrix]
[matrix.main]
Server="https://matrix.montaigne.nightmared.fr"
Login="matterbridge"
Password="a bad password was here"

[discord]
[discord.main]
Token="YEEAAH A TOKEN"
Server="some number"
RemoteNickFormat="{NICK}[{PROTOCOL}]"
AutoWebhooks=true

[general]
RemoteNickFormat="<{NICK}> "
UseAppService=true
AppServiceHost="192.168.1.13"
AppServicePort=1234
AppServiceConfigPath="appservice_synapse.yml"

[[gateway]]
    name="test1"
    enable=true

    [[gateway.inout]]
    account="matrix.main"
    channel="#test:some_server"

    [[gateway.inout]]
    account="discord.main"
    channel="ID:some number"

[[gateway]]
    name="test2"
    enable=true

    [[gateway.inout]]
    account="matrix.main"
    channel="#test2:some_server"

    [[gateway.inout]]
    account="discord.main"
    channel="ID:some number"
```

Reviews/tests would be much appreciated!

Work that remains to be done:
- [x] showing usernames properly when in appservice mode (example: https://nightmared.fr/files/2021-07-10-1625934431.png)
- [ ] fixing the harcoded `@__bridge_` mentionned earlier
- [ ] accepting rooms not defined in the config (that is the whole point of app services after all, yet currently it only relays messages for rooms that are already defined in the config AND that have an alias matching the regexp in namespaces.aliases.regex)


nightmared:

PR #1547, opened 2021-07-13
Fix XMPP parseNick function

tytan652:
The length of the split shall be higher than 1, if it equals 1 a crash will happen because of the following split which use the second value of the array created by the first split.


42wim:
Good catch, thanks!

Issue #1459, opened 2021-04-29
Allow converted animated telegram stickers to be smaller


Funatiker:
**Is your feature request related to a problem? Please describe.**
If `lottie_convert.py` is present and set to be used, it is basically used at fully quality. This means that the  tens of kilobytes sized gzipped lottie/JSON files turn to several megabytes sized webp files. This wastes quite a bit of mobile traffic, and storage space. Depending on the hardware, it also takes several seconds to render. This leads to stickers arriving in a different context in the conversation.

**Describe the solution you'd like**
Currently `lottie_convert.py` is called with no parameters regarding quality: https://github.com/42wim/matterbridge/blob/0450482e6ebed8a379f2d43f0ee264ea0d130a31/bridge/helper/helper.go#L283 One could lower the size of the WebP files by skipping a few frames and lowering the quality.

```bash
# current setting: 4.2 MB WebP vs 20 kB tgs
$ lottie_convert.py --input-format lottie --output-format webp file_106_tgs.tgs file_106_tgs_orig.webp
# rendering a fourth of all frames, at 25% quality: 736 kB
$ lottie_convert.py --webp-skip-frames 4 --webp-quality 25 --input-format lottie --output-format webp file_106_tgs.tgs file_106_tgs_low.webp
# rendering a fifth of all frames, at 15% quality: 532 kB
$ lottie_convert.py --webp-skip-frames 5 --webp-quality 15 --input-format lottie --output-format webp file_106_tgs.tgs file_106_tgs_lower.webp
# rendering a sixth of all frames, at 10% quality: 424 kB
$ lottie_convert.py --webp-skip-frames 6 --webp-quality 10 --input-format lottie --output-format webp file_106_tgs.tgs file_106_tgs_lowest.webp
```

One could either change the current call of `lottie_convert.py` to one with a lower quality setting, or add a setting that allows picking changing the quality setting.

**Describe alternatives you've considered**
Obviously, the converting animated vector graphics to a sequence of pixelated graphics is a workaround for clients that don't support animated vector formats. Theoretically, one could consider conversion to a classic video format. `lottie_convert.py` should support that  (see https://mattbas.gitlab.io/python-lottie/script_lottie_convert.html ) my version 0.6.7 from pip didn't. But of course the best alternative (but not scope of this project) would be patching lottie/tgs support into the clients. Thus, sticking to the current variant but with lower quality settings seems to be the most pragmatic approach.

**Additional context**
[...]

PR #1569, opened 2021-08-13
Convert .tgs with better code

Benau:
As discussed in https://github.com/42wim/matterbridge/issues/1459 I made the PR, and it seems that it's not very difficult to keep the old code

Right now "making tgs smaller" is not yet done, but it should be easier to add after this PR is merged (and I think it's better to not  mix too many things in same PR)


42wim:
@Benau Thanks, looks good, could you make the linter happy and then we can merge it?

Benau:
```go
	if err != nil {
 		return fmt.Errorf("Failed to run libtgsconverter.ImportFromData: %s", err.Error())
 	}

 	*data = blob
 	return nil
```
I guess this should fix the error now?

bridge/helper/libtgsconverter.go
```diff
@@ -0,0 +1,34 @@
...
+	return nil
+}
+
+// ConvertTgsToX convert input data (which should be tgs format) to any format supported by libtgsconverter
+func ConvertTgsToX(data *[]byte, outputFormat string, logger *logrus.Entry) error {
+	options := libtgsconverter.NewConverterOptions()
+	options.SetExtension(outputFormat)
+	blob, err := libtgsconverter.ImportFromData(*data, options)
+	if err != nil {
+		return fmt.Errorf("Failed to run libtgsconverter.ImportFromData: %s", err.Error())
```
42wim, line 21:
```suggestion
		return fmt.Errorf("failed to run libtgsconverter.ImportFromData: %s", err.Error())
```

bridge/helper/libtgsconverter.go
```diff
@@ -0,0 +1,34 @@
+// +build cgo
+
+package helper
+
+import (
+	"fmt"
```
42wim, line 6:
```suggestion
	"fmt"
	
```

bridge/helper/libtgsconverter.go
```diff
@@ -0,0 +1,35 @@
...
+
```
42wim, line 7:
```suggestion

```

42wim:
@Benau thanks!

PR #1572, opened 2021-08-20
Not needed newline character in mumble bridge multiline messages

powerjungle:
Because it looks ugly and serves no purpose.

Before:
Screenshot 2021-08-20 235634

After:
Screenshot 2021-08-20 235705

Message being sent:
Screenshot 2021-08-20 235729


42wim:
Thanks!

PR #1574, opened 2021-08-25
[Fix]: Fixed the bug in adding replies on Mattermost thread from Slack.

avas27JTG:
Fixed the bug in adding replies on Mattermost thread from Slack.


avas27JTG:
@42wim Please review

42wim:
@avas27JTG what issue are you referring too?

avas27JTG:
> @avas27JTG what issue are you referring too?

There was a issue (server was stopping) when adding a reply from **slack** on a parent message from **mattermost**.

PR #1589, opened 2021-09-21
Better error handling on Zulip

alexmv:
We noticed that matterbridge bots were making thousands of requests to Zulip servers with unauthenticated credentials.  This is an attempt to stem that tide.

Specifically, it catches failures when registering, which were previously completely ignored -- most of the unauthenticated requests were to `yourserver.zulipchat.com` which is the example host in the configuration, which isn't currently a valid Zulip realm:
https://github.com/42wim/matterbridge/blob/53cafa9f3d0c8be33821fc7338b1da97e91d9cc6/matterbridge.toml.sample#L1564-L1582

It also catches the 401's when fetching events, which were not rate-limited previously.

Finally, it provides a more specific user-agent, which will make it easier to track down errors like this in the future.  It does so by doing a small re-arrangement of packages to make the version information available via a package which is not `main`.

I made commits directly into `vendor/github.com/matterbridge/gozulipbot/` directly, rather than open a PR against https://github.com/matterbridge/gozulipbot.  It looks like the `gozulipbot` code in this repo has already diverged from the other repository, so it looked easier to fix it here.  Regardless, however, the key flaws of failing to check for failure during queue registration and failing to rate-limit 401 responses _also_ apply to the `matterbridge/gozulipbot` repository, and should also be applied there.

I'm happy to open a second PR there if you'd prefer; let me know how you'd like to proceed, as it looks like there are changes other than these which need backporting out of `vendor/` and into `matterbridge/gozulipbot`.


alexmv:
Gentle poke on this?

alexmv:
I've pushed the gozulipbot-specific changes to matterbridge/gozulipbot#1; this PR shouldn't be closed, because there are changes that are not contained within the vendored gozulipbot.

I've rebased, dropping the gozulipbot-only changes, and re-pushed.  This is _not expected to pass tests_ until it is rebased past a commit which pulls in the gozulipbot changes.

PR #1640, opened 2021-11-18
Allow binding to IP on IRC

inzanity:
Add configuration option "Bind" that is passed on to girc, allowing
to choose which IP address to use on systems that have multiple ones.


42wim:
Thanks, could you also update the https://github.com/42wim/matterbridge/blob/master/matterbridge.toml.sample file with the new option?

42wim:
Thanks!

Issue #1649, opened 2021-12-06
Forward from a telegram group with image, only shares image


soloam:
**Describe the bug**
When I forward a message from a group (message with text and image) to another with the bridge, it only forwards to discord (only tested in discord) the image, not the text

**To Reproduce**
Add a gateway with discord and telegram, forward to a telegram group a message with text and image, and the message will be send by the matterbridge to discord, but only the image

**Expected behavior**
send both text and image

**Screenshots/debug logs**

Telegram:
![image](https://user-images.githubusercontent.com/11949987/144870629-324eee3f-f2a6-48f2-8d8f-7d14b6359bab.png)

Discord:
![image](https://user-images.githubusercontent.com/11949987/144870704-b5fea9a2-a865-44c4-9ad3-172863fa214d.png)

**Environment (please complete the following information):**
 - OS: freebsd
 - Matterbridge version: 1.23.2 73ec02ab

soloam:
I did a debug, and the this is the result:

`
[0047] DEBUG discord:      [handleEventWebhook:bridge/discord/webhook.go:139] Processing webhook sending for message &config.Message{Text:"Forwarded from unknown: ", Channel:"ID:xxxxxxxxx", Username:"[telegram] <NAME> ", UserID:"xxxxxxxxx", Avatar:"", Account:"telegram.mytelegram", Event:"", Protocol:"telegram", Gateway:"mygateway_test", ParentID:"", Timestamp:time.Date(2021, time.December, 7, 13, 39, 56, 470053548, time.Local), ID:"", Extra:map[string][]interface {}{"file":[]interface {}{config.FileInfo{Name:"file_2.jpg", Data:(*[]uint8)(0xc00000d5c0), Comment:"✔️ Xiaomi Mijia Mi Smart Standing Fan 2 Lite Fan |EU| 🇪🇺️\n🥇 Price: $52.49 (47.05€)\n🔥 The lowest price (in this store)! 🔥 Best before: 61.49 USD\n❇️ #Aliexpress Direct link ✂️ https://bit.ly/3ouSJed\n✏️ PROMOTION CODE ⭕️ 12DEC7\n\n✅ Similar products: https://bit.ly/3EAZGA4\n✅ Coupons: http://bit.ly/2I1VXQb", URL:"", Size:0, Avatar:false, SHA:""}}}}
`

It looks like the text is in telegram comment! This should be forwarded as message

PR #1652, opened 2021-12-07
[fix] telegram: Add's comments to message

soloam:
This is a change to handle comments in telegram messages!

Some messages in telegram have comments added to the message! This normally is the description in images or links. This changes appends the comment to the message if available.

This should fix the issue in #1649


42wim:
hi @soloam did you test this? the code doesn't build

soloam:
I'm sorry for the confusion, I pushed the wrong branch! I corrected the branch and updated the new change! Discord was ignoring the image caption if a text was in the image! The caption contains important information from the extra! It should be forwarded!

42wim:
But this change will remove any text ?

bridge/discord/webhook.go
```diff
@@ -82,10 +82,8 @@ func (b *Bdiscord) webhookSend(msg *config.Message, channelID string) (*discordg
 				ContentType: "",
 				Reader:      bytes.NewReader(*fi.Data),
 			}
-			content := ""
-			if msg.Text == "" {
-				content = fi.Comment
-			}
+			content := fi.Comment
```
42wim, line 85:
```suggestion
			content := fi.Comment
			if msg.Text != "" {
				content = msg.Text
			}
```

42wim: can you test with this change, so we only add the comment when there is empty text

42wim (changes requested):

soloam:
That was the original behavior! The problem is that when it's forwarded (for example telegram writers "forwarded from") or anything is written on the text, the comments are removed.

42wim:
Ok, as I'm not using telegram I take you for your word :)

42wim (approved):

PR #1664, opened 2021-12-18
Reply support for Matrix

vpzomtrrfrt:
Implements sending and receiving reply messages for Matrix bridges

In doing this, I moved the protocol stripping downstream from FindCanonicalMsgID since the previous behaviour made the ParentID value inconsistent

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -454,9 +454,9 @@ func (gw *Gateway) SendMessage(
 		msg.Channel = rmsg.Channel
 	}
 
-	msg.ParentID = gw.getDestMsgID(rmsg.Protocol+" "+canonicalParentMsgID, dest, channel)
+	msg.ParentID = gw.getDestMsgID(canonicalParentMsgID, dest, channel)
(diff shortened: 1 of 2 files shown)


gateway/gateway.go
```diff
@@ -454,9 +454,9 @@ func (gw *Gateway) SendMessage(
 		msg.Channel = rmsg.Channel
 	}
 
-	msg.ParentID = gw.getDestMsgID(rmsg.Protocol+" "+canonicalParentMsgID, dest, channel)
+	msg.ParentID = gw.getDestMsgID(canonicalParentMsgID, dest, channel)
```
vpzomtrrfrt, line 457: not sure exactly what's going on here

The previous code caused multiple protocol names to be included in the message ID, which seems wrong

Also removed the protocol from the message ID in the fallback case since that gets sent to the destination as invalid

42wim:
Thanks! could you fix the small linting issue, I'll try to do a review the coming week.

42wim:
Thanks! 🚀

PR #1668, opened 2021-12-30
vk: Remove GroupID

SevereCloud:
- using `longpoll.NewLongPollCommunity(vk *api.VK)` instead `longpoll.NewLongPoll(vk *api.VK, groupID int)`
- add `lp.Shutdown()` for `Disconnect()`

### DIFF
--- a/bridge/vk/vk.go
+++ b/bridge/vk/vk.go
@@ -34,6 +34,7 @@ type user struct {
 
 type Bvk struct {
 	c            *api.VK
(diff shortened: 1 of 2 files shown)


42wim:
Thanks for this PR, could you give a reason why the group ID needs to be removed? (I'm not using VK myself)
Is this not longer used or being autodetected?

SevereCloud:
Being autodetected

42wim:
Thanks!

PR #1703, opened 2022-01-31
Telegram Bot Command /chatId

DavyJohnes:
For regular users it is not easy task to get ID of telegram group, this bot command helps to do it.


42wim:
Thanks, good idea! 👍

PR #1707, opened 2022-02-02
Add support for telegram captions

lekoOwO:


lekoOwO:
I think this should be the real fix for https://github.com/42wim/matterbridge/issues/1649

42wim:
For #1649 it's actually the comment as can be seen in the debug log there
Do you have messages where it's the caption?

lekoOwO:
> For #1649 it's actually the comment as can be seen in the debug log there
> Do you have messages where it's the caption?

Oh I see.
For normal pictures like the one we send in group/supergroup/private chat, it should be captions.
Captions also applies to document, audio, video.... too.

42wim:
We already do this here: https://github.com/42wim/matterbridge/blob/master/bridge/telegram/handlers.go#L350 
The caption gets set as a `Comment` and in receiving bridges this comment will be added to the text, see example for discord https://github.com/42wim/matterbridge/blob/master/bridge/discord/discord.go#L339

Where specifically are you not getting the captions? Could you paste a debug log?

PR #1753, opened 2022-03-01
Fix Telegram channel title in forwards

ValdikSS:
Forward from channels requires different handling than forward from the regular users.
This patch fixes the issue: it prints channel title instead of "forwarded from unknown".


42wim:
Thanks!

Issue #909, opened 2019-09-26
Complex-formatted text with URLs from Telegram gets randomly corrupted


ValdikSS:
**Describe the bug**
Long messages (posts forwarded to bridged chat from telegram channel) with multiple URLs gets corrupted: the URLs are misplaced and appear in the middle of the text, rewriting original text in this place.

See the screenshot.

`с суммарным полезным объёмом ~45TB и начинаете ЭКСПЛУАТАЦИЮ.`
became
`с суммарным полезным объёhttps://t.me/TrTrTr8holes/119ПЛУАТАЦИЮ.`

This URL should be in `обои от стен отклеиваются`

P.S. music.yandex.ru link on top is missing space before the link. I'm not sure if it's relevant to this bug.

**To Reproduce**
Open https://t.me/TrTrTr8holes/236 and forward it into bridged chat.

**Expected behavior**
The URLs are placed correctly

**Screenshots/debug logs**
![Screenshot_20190926_212647-fs8](https://user-images.githubusercontent.com/3054729/65714927-bce6a000-e0a4-11e9-997a-58b1a6a4c830.png)

**Environment (please complete the following information):**
 - OS: Linux, x86_64
 - Matterbridge version: 1.15.1 87788f3

**Additional context**
```
[telegram.bridged]
…
RemoteNickFormat="{LABEL}<b>&lt;{NICK}&gt;</b> "
UseInsecureURL=false
MessageFormat="HTMLnick"
Label="[T] "
```

42wim:
Issue is probably in https://github.com/42wim/matterbridge/blob/727fa9f92971a29eaa1790fb249bd3a7baa919f6/bridge/telegram/handlers.go#L367-L388 if anyone wants to debug.

https://github.com/go-telegram-bot-api/telegram-bot-api/issues/231 is also related

ValdikSS:
Fixed by https://github.com/42wim/matterbridge/pull/1765

PR #1765, opened 2022-03-16
Fix for complex-formatted Telegram text

ValdikSS:
This patchset fixes issue #909.
> Long messages (posts forwarded to bridged chat from telegram channel) with multiple URLs gets corrupted: the URLs are misplaced and appear in the middle of the text, rewriting original text in this place.

### DIFF
--- a/bridge/telegram/handlers.go
+++ b/bridge/telegram/handlers.go
@@ -516,7 +516,7 @@ func (b *Btelegram) handleEntities(rmsg *config.Message, message *tgbotapi.Messa
 
 		if e.Type == "pre" {
 			offset := e.Offset + indexMovedBy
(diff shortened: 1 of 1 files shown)


ValdikSS:
Tested with Latin, Russian and Chinese, so I guess no need to explicitly encode as utf16.

ValdikSS:
Turned out that Unicode emojis break the markup, fixed by processing everything in utf16.

42wim:
@ValdikSS thanks a lot for looking into this!
could you just run a `gofmt -s` on handlers.go so the linter is happy and I can merge it

ValdikSS:
> @ValdikSS thanks a lot for looking into this! could you just run a `gofmt -s` on handlers.go so the linter is happy and I can merge it

Done, please check.

42wim:
Thanks!

PR #1773, opened 2022-03-21
Update lrstanley/girc dep

42wim:


poVoq:
I am seeing some strange connection issues to IRC channels bridged by Matterbridge since updating from 1.24 to 1.25.

It seems like the nicks Matterbridge now creates with RELAYMSG are somehow malformed, causing modern IRC clients to have connection issues or crashing when trying to connect.

Any idea what might be the cause? I am pretty sure it is related to this PR as previously it worked without problems.

42wim:
Open an issue with debug logs also try to run with 1.24 to see that it actually doesn't happen there.
It would be strange because that code isn't touched.

PR #1774, opened 2022-03-22
Add Dockerfile_whatsappmulti for building with WhatsApp Multi-Device support

tpxtron:


42wim:
Thanks!

Waldelf:
Hey,
I'm not so firm with linux and docker - but I kept running a matterbridge on docker for about a year now, quite stable.
But I don't get the new whatsapp bridge starting - what do I have to do to get a docker container with the new whatsapp_multi bridge already compiled inside?

PR #1775, opened 2022-03-22
Use slack real name as user name

sas1024:
Added new option to use Slack real name as user name.


42wim:
I think we better keep the same naming convention and also use `UseFullName` here like with telegram.
In the comments you can refer to the slack real name name

sas1024:
@42wim done! :)

42wim:
👍

PR #1776, opened 2022-03-22
Preserve threading from telegram replies

sas1024:
I've tested it between Slack and Telegram - all fine.

Telegram replies appears as thread replies in Slack, and Slack thread replies appears as Telegram replies in Telegram.


42wim:
Thanks! 🚀

PR #1779, opened 2022-03-28
Telegram: Multiple media in one message

sas1024:
In this PR:

1. Fix for quoting when message have media with caption.
2. Send message with multiple media as one message with media group.
3. Preserve threading for telegram messages with files.

I've test this changes via Slack and Telegram.

Old behavior: Sending one message with 3 images/files from Slack produces 3 telegram messages with one image/file each.
New behavior: Sending one message with 3 images/files from Slack produces one telegram message with 3 images/files.


42wim:
Also good job! 👍 thanks!

PR #1781, opened 2022-03-30
Slack: Preserve threading for messages with files

sas1024:
This PR allow to preserve threading when Slack receive files.

I've tested it between Telegram and Slack, all work fine.

**Old behavior:** Send file From Telegram to Slack will produce file in Slack, but if I quote this file in Telegram and send message, this message will not received by Slack. If I reply to message with this file in Slack, Telegram message will be not attached as reply on message with file.

**New behavior:** Send file From Telegram to Slack will produce file in Slack, if I quote this file in Telegram and send message, this message will received by Slack. If I reply to message with this file in Slack, Telegram message will be attached as reply on message with file.

### DIFF
--- a/bridge/slack/slack.go
+++ b/bridge/slack/slack.go
@@ -321,7 +321,7 @@ func (b *Bslack) sendRTM(msg config.Message) (string, error) {
 	}
 
 	// Upload a file if it exists.
-	if msg.Extra != nil {
+	if len(msg.Extra) > 0 {
 		extraMsgs := helper.HandleExtra(&msg, b.General)
 		for i := range extraMsgs {
 			rmsg := &extraMsgs[i]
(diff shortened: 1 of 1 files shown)


bridge/slack/slack.go
```diff
@@ -471,13 +472,22 @@ func (b *Bslack) uploadFile(msg *config.Message, channelID string) {
...
 		if err != nil {
 			b.Log.Errorf("uploadfile %#v", err)
-			return
+			return "", err
 		}
 		if res.ID != "" {
 			b.Log.Debugf("Adding file ID %s to cache with timestamp %s", res.ID, ts.String())
 			b.cache.Add("file"+res.ID, ts)
+
+			//search for message id by uploaded file in private/public channels, get thread timestamp from uploaded file
```
42wim, line 481:
```suggestion
			// search for message id by uploaded file in private/public channels, get thread timestamp from uploaded file
```

(to fix the gofumpt lint)

42wim:
Very nice, thank you!

PR #1783, opened 2022-03-31
Ignore sending file with comment, if comment contains IgnoreMessages value

sas1024:
Now `IgnoreMessages` setting check and attached file comment too.

How I test it:
In Slack bridge configuration I set `IgnoreMessages="^~~"`

**Old Behavior:**
Sending message with attached file from Slack to Telegram with file comment "~~ test comment" sends this file to Telegram.

**New Behavior:**
Sending message with attached file from Slack to Telegram with file comment "~~ test comment" not sends this file to Telegram.


42wim:
👍

Issue #1780, opened 2022-03-30
Matrix to Discord replies buggy since version 1.2.4


BrotherIAmPinnedHere:
**Describe the bug**
As of version 1.2.4, replies seem thoroughly broken when done on Matrix and bridged to Discord.  Ordinarily, replies look something like this: 

![image](https://user-images.githubusercontent.com/16894072/160778887-290dc305-8f93-427a-8ef7-9303aaab5214.png) (ignore the first two 'test's)

After version 1.2.4, the quote without preservethreading: true ends up coming through as 

```
[thread:]
test
```
With preservethreading: true, replies will show up correctly on Discord, but puppeting will break and they will appear to have come from the bridge bot its self rather than the user who made the comments. 
(usernames removed as the users did not consent to being posted)

![image](https://user-images.githubusercontent.com/16894072/160780648-293cedc6-2d18-4303-b2ed-7c8861a47cbd.png)

**To Reproduce**
Uncertain. Running Ubuntu 18.04. There does not appear to be anything untoward in the SystemD logs for Matterbridge and I didn't see anything strange when debugging. I am using replacemessage to prevent abuse of pings through the bridge and a general media upload limit of 8 MB , but other than that basically everything is default.

BrotherIAmPinnedHere:
Is there a setting to restore the original reply system after V1.2.4? Seemed to be the best alternative without Discord directly supporting it.

42wim:
What version are you talking about because 1.2.4 is very old

PR #1791, opened 2022-04-03
Change discord non-native threading behaviour

42wim:
Sorta regression introduced by 9a8ce9b17e560433731eb5efa3cee7ced0b93605
which changes the way we get replies of matrix.

This causes issues like https://github.com/42wim/matterbridge/issues/1780
We "fix" this by mimicking the old behaviour when "PreserveThreading" is
disabled.


Lucki:
I was unaware this setting even exists for Discord and Matrix, the wiki (https://github.com/42wim/matterbridge/wiki/Settings) only lists this setting available for slack.

Issue #474, opened 2018-08-04
Slack > Mattermost Attachments


Tjird:
**Is your feature request related to a problem? Please describe.**
As far as I know not, looked into the issues that already exist.

**Describe the solution you'd like**
Solution doesn't exist, it's a feature request.

**Describe alternatives you've considered**
Alternative doesn't exist, it's a feature request.

**Additional context**
Add support to send the whole attachment through the message. Now a day if the message doens't have text it is looking to the text in the attachment part. I have tested it with a webhook and user and doens't work atm.

42wim:
If I understand the problem correct, you want the slack message attachments (https://api.slack.com/docs/message-attachments) to be sent to mattermost? (we're not talking about file attachments?)

Are you getting attachments which are missing the `fallback` property so that you don't get a message on mattermost?

My issue with this is that the only bridge that supports slack attachments beside slack is mattermost.
I want to keep stuff as generic as possible between the different bridges, so I'm not going to add this option at the moment.

Tjird:
Yes, I want that the whole slack attachments part is being sent through the bridge. And yes we are not talking about file attachments. The message has got the fallback property filled in, I'm getting a message but not the message that I got in slack. 

I can understand your issue about this, it's easier the get everything nearly the same.

Tjird:
@42wim I have done some more research and it looks like the bot is sending only the Title and Text in the attachment part through, is this correct? 

I also have a question on your answer, is it not possible to add an argument in the config file that sounds like `SlackAttachments`. If this argument is set to true is sends the whole attachment part through and if it is set to false it stays the same as now a day. Is this an option?

42wim:
Correct, if the `Text` is not empty it uses the `Text` and `Title`. If the `Text` is empty it uses `Fallback`

Yes, everything is possible, it just takes work :-)

zjoasan:
I don't know if this is the issue for me, but some of the info/messages from github-bot on slack wont cross the bridge to mattermost: 

Slack message:
[osmc/osmc] New comment by samnazarko on issue #534: [Installer] An error occurred while writing the image! Please consult logfile
I've updated to Catalina's .1 point release, so can test this shortly.

Mattermost post: 
I've updated to Catalina's .1 point release, so can test this shortly.

Missing all the data on who posted and on which issue wher issue number and description is a link to issue on github.

We prefer the Slacks Github-bot to the Mattermost Github-bot.

zjoasan:
[...]

Issue #265, opened 2017-09-19
Mattermost->Slack: links in attachments need rewrite


ninja-:
If you have a configuration problem, please first try to create a basic configuration following the instructions on [the wiki](https://github.com/42wim/matterbridge/wiki/How-to-create-your-config) before filing an issue.

Please answer the following questions. 

### Which version of matterbridge are you using?
run ```matterbridge -version```
1.2.1-dev

### Please describe the expected behavior.
Links in attachments should be rewritten from format of
```
[text](link)
```

to
```
<link|text>
```

### Please describe the actual behavior. 
Links are not rewritten and are displayed incorrectly on slack.

42wim:
I'm trying to test this, but if I put a link in an attachment it, it just shows fine on slack.
What app does this ? And if you connect that app to slack does it behave correct ?

Could you run with -debug, reproduce such a message and paste the debug output here ?

42wim:
I guess you're using a mattermost plugin for jenkins, which isn't slack compatible.


I've looked some more into it (writing it down for future reference and/or for anyone who wants to work on it) and possible solutions for this is:

For each "slack_attachment" received by mattermost webhook, check if pretext, text or fields contain markup, if so add mrkdown_in for that field.

```Valid values for mrkdwn_in are: ["pretext", "text", "fields"]. Setting "fields" will enable markup formatting for the value of each field.```

https://api.slack.com/docs/message-formatting#message_formatting for more info

ninja-:
@42wim I am using GitLab Slack integration + mattermost webhook link.
Probably mattermost converts it to it's own format when it's coming as a slack hook, which is kind of normal.
All other formattings like bold etc are passed properly by matterbridge.
I think doing a regexp replacement for the link format would be a proper solution here like I mentioned?
But I am not sure at which level it should be done for best compatibility with other bridges like mattermost->discord etc...

ninja-:
I think it's best to normalize it to slack formatting at mattermost bridge level, so we do the link format regexp there.

ninja-:
@42wim mrkdown_in is set correctly here so that's not a problem, it's just that the link format is different on mattermost and that gets converted when a webhook is received by mattermost.

So what we can do
here https://github.com/42wim/matterbridge/blob/master/bridge/mattermost/mattermost.go#L240
1) Cast to array
2) For each
- get mrkdown_in field
- get each field's content for fields in mrkdown_in
- apply link regexp on that field

[...]

PR #1807, opened 2022-04-20
Improve Slack attachments formatting (slack)

chugunov:
While I was tuning bridge between Slack and Mattermost, I faced with a problem, that Slack message with the attachment (not a file attachment, but a message, that bot or integration generates) lost `TitleLink` and `Footer` attributes at Mattermost. 

I made a quick fix for replacement of Slack link format `<link|text>` to Markdown format `[text](link)`, that could be a breaking change for other bridges that doesn't support Markdown format. Also add `Footer` to the result message.

Can be linked to #265 #474


42wim:
Thanks!

PR #1808, opened 2022-04-20
Telegram: Do not apply any markup to URL entities

ValdikSS:
handleEntities code uses simple modification offset which does not
allow to detect whether the offset is placed before or after
the element in already modified string.
This works fine is most cases as Telegram server always sort the
elements by offset, in ascending order.
However, this is not the case when the modification, for example bold
text, is applied to the URL. In this case, the offset of URL and
bold entity is equal, which raises the issue.

This commit introduces additional hack for this case, stripping
any entities which intersect with URL.

### DIFF
--- a/bridge/telegram/handlers.go
+++ b/bridge/telegram/handlers.go
@@ -517,8 +517,8 @@ func (b *Btelegram) handleEntities(rmsg *config.Message, message *tgbotapi.Messa
 	}
 
 	indexMovedBy := 0
+	prevLinkOffset := -1
 
-	// for now only do URL replacements
 	for _, e := range message.Entities {
 
 		asRunes := utf16.Encode([]rune(rmsg.Text))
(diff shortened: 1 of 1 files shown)


bridge/telegram/handlers.go
```diff
@@ -537,34 +537,37 @@ func (b *Btelegram) handleEntities(rmsg *config.Message, message *tgbotapi.Messa
...
 			indexMovedBy += len(url.String()) + 3
+			prevLinkOffset = e.Offset
 		}
 
-		if e.Type == "code" {
-			offset := e.Offset + indexMovedBy
-			rmsg.Text = string(utf16.Decode(asRunes[:offset])) + "`" + string(utf16.Decode(asRunes[offset:offset+e.Length])) + "`" + string(utf16.Decode(asRunes[offset+e.Length:]))
-			indexMovedBy += 2
-		}
+		if e.Offset != prevLinkOffset {
```
42wim, line 543:
```suggestion
		if e.Offset == prevLinkOffset {
			continue
		}
```

To keep the linter happy and make it a bit less complex

42wim (changes requested):

ValdikSS:
Complexity did not decrease :D

42wim:
But linter is happy, and we have 1 nested if less :)

42wim:
Thank you @ValdikSS

PR #1818, opened 2022-04-29
Fix telegram message deletion request

sas1024:
This commit fix telegram message deletion behavior.

In old behavior Send method used, message deleted in telegram, but method will return error.

I changed it to [recommended way](https://github.com/go-telegram-bot-api/telegram-bot-api/issues/514).


42wim:
Thanks!

PR #1836, opened 2022-05-21
Whatsapp changes

Y6b:
-First part of deletion code
-Audio message type added
-Quoted message sent
-Instead of number or "someone" using pushname if available


Y6b:
All ok

Y6b:
All is ok

Issue #1612, opened 2021-10-17
Unwanted join notifications from one Discord server to another


halian:
**Describe the bug**
My MatterBridge instance has two Discord servers, each bridged to mutually exclusive sets of IRC channels. When someone joins Discord B, a notification is posted by the bot to Discord A.

**To Reproduce**
Unknown. Nothing obviously leading to the bug can be found in the configuration file.

**Expected behavior**
Notifications to be posted to Discord A only when someone joins IRC A, and same for Discord B/IRC B.

**Screenshots/debug logs**
```
[0043] DEBUG discord:      [memberAdd:bridge/discord/handlers.go:183] <= Sending message from discord.safiralliance to gateway
[0043] DEBUG discord:      [memberAdd:bridge/discord/handlers.go:184] <= Message is config.Message{Text:"Ketsuban joins", Channel:"", Username:"system", UserID:"", Avatar:"", Account:"discord.safiralliance", Event:"join_leave", Protocol:"", Gateway:"", ParentID:"", Timestamp:time.Time{wall:0x0, ext:0, loc:(*time.Location)(nil)}, ID:"", Extra:map[string][]interface {}(nil)}
[0043] DEBUG discord:      [memberAdd:bridge/discord/handlers.go:183] <= Sending message from discord.alden-ic to gateway
[0043] DEBUG discord:      [memberAdd:bridge/discord/handlers.go:184] <= Message is config.Message{Text:"Ketsuban joins", Channel:"", Username:"system", UserID:"", Avatar:"", Account:"discord.alden-ic", Event:"join_leave", Protocol:"", Gateway:"", ParentID:"", Timestamp:time.Time{wall:0x0, ext:0, loc:(*time.Location)(nil)}, ID:"", Extra:map[string][]interface {}(nil)}
[0043] DEBUG discord:      [memberAdd:bridge/discord/handlers.go:183] <= Sending message from discord.alden-ooc to gateway
[0043] DEBUG discord:      [memberAdd:bridge/discord/handlers.go:184] <= Message is config.Message{Text:"Ketsuban joins", Channel:"", Username:"system", UserID:"", Avatar:"", Account:"discord.alden-ooc", Event:"join_leave", Protocol:"", Gateway:"", ParentID:"", Timestamp:time.Time{wall:0x0, ext:0, loc:(*time.Location)(nil)}, ID:"", Extra:map[string][]interface {}(nil)}
[...]

PR #1846, opened 2022-06-11
discord: Ignore events from other guilds, add nosendjoinpart support

wlcx:
- Filter all incoming discord events and ignore ones originating from Guild IDs other than the one we have configured. This is necessary because discord bots receive events for *all* discord guilds that they are present in. Fixes https://github.com/42wim/matterbridge/issues/1612
- Add support for `nosendjoinpart` to the discord bridge


42wim:
Thanks! 👍

PR #1855, opened 2022-06-29
Fixed message html entities escaping when sending to Telegram

sas1024:
Hello!

If you send message (for example, from Slack) with following content to Telegram:
> ```I'm just <br /> unescaped html with ````<b>some</b> html```` inside code block```

Message will be failed to to send with following error:
`SendMessage failed: Bad Request: can't parse entities: Unsupported start tag "br" at byte offset`

This PR just fixes this issue with Telegram messages sending with message `MessageFormat="HTML"`.

### DIFF
--- a/bridge/telegram/html.go
+++ b/bridge/telegram/html.go
@@ -2,7 +2,6 @@ package btelegram
 
 import (
 	"bytes"
-	"html"
 
 	"github.com/russross/blackfriday"
 )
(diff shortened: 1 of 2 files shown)


42wim:
Thanks!

sas1024:
@42wim there is typo in milestone :) 1.52.3 -> 1.25.3

PR #1856, opened 2022-07-06
Update discord.go

greenlighttec:
Added bolding around the name, and split the name vs message with ": " on line 338

### DIFF
--- a/bridge/discord/discord.go
+++ b/bridge/discord/discord.go
@@ -335,7 +335,7 @@ func (b *Bdiscord) handleEventBotUser(msg *config.Message, channelID string) (st
 	}
 
 	m := discordgo.MessageSend{
-		Content:         msg.Username + msg.Text,
+		Content:         "**" + msg.Username + "**: " + msg.Text,
 		AllowedMentions: b.getAllowedMentions(),
 	}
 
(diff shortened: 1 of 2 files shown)


42wim:
You should use `RemoteNickFormat` config for changing the nick appearance.

greenlighttec:
I mean I'll look at that, but the current code is purposely stringing the name and message together without any kind of separation. Surely that cannot be intended design?

greenlighttec:
Ok the `RemoteNickFormat` changes the display for ALL messages. I'm specifically fixing an issue in threads (with `PreserveThreading` set to true) that the name of the person sending the message (Since the sender is the bot name) and the message itself is concatenated together and looks dumb. My changes will fix this and make the threads easier to read. 

image

You can see the Bot messages look fine, but the thread reply looks ridiculous.

greenlighttec:
Okay, I compiled the changes I suggested and tested and found that they didn't actually do anything lol. Definitely not what I was trying to achieve. I spent some more time actually reading through the code and I think i'm in the wrong place to influence the messages being sent when a thread is occurring. You can ignore this for now if I find out where to do the change I'll make a new pull request.

jonaswouters:
Did you ever solve this @greenlighttec because I think this was a clean solution

greenlighttec:
https://github.com/mspgeek-community/matterbridge

Issue #1393, opened 2021-02-08
NOTICEs are only visible on IRCs


Mikaela:
**Describe the bug**

I have matterbridge in 5 IRC networks (excluding Twitch), XMPP and Telegram. I see NOTICEs in IRC networks, but not in Matrix (which supports them natively) or XMPP or Telegram. 

**To Reproduce**

1. Matterbridge 2 IRC networks together and some other network.
2. Send a NOTICE.

**Expected behavior**

The NOTICE is visible on all other protocols, if it doesn't have an equivalent message type, then it's shown as a normal message.

**Screenshots/debug logs**

**Environment (please complete the following information):**
 - OS: Debian GNU/Linux 10 (buster)
 - Matterbridge version: version: 1.22.0 afcd362c

**Additional context**

The message type exists so automatic messages aren't send in response to them. https://tools.ietf.org/html/rfc1459#section-4.4.2

At times it looks like NOTICEs aren't visible even to other networks, I am not sure if this is related to RELAYMSG.

See also https://github.com/42wim/matterbridge/pull/1305.

PR #1864, opened 2022-07-24
[Matrix] switch to mautrix-go + initial support for application services

### LINKED ISSUE #137: Use matrix Application Service API (matrix)

nightmared:
Hello,

This is a revival of #1541, #1543 and #1544.

Fixes #137.

Before considering reviewing this, I would really like for other people to confirm that it works for them, to detect some of the numerous issues that must be lurking inside this changeset.

If you want to test the appservice support, you can use the same example configuration than in #1544.

The work that remains to be done mainly revolves around application services:
- [ ] supporting more complex regexps for the usernames
- [ ] accepting rooms not explictly hardcoded in the config (that is the whole point of app services after all, yet currently it only relays messages for rooms that are already defined in the config)
- [ ] testing this enough to be confident it works as expected


poVoq:
>  accepting rooms not explictly hardcoded in the config (that is the whole point of app services after all, yet currently it only relays messages for rooms that are already defined in the config)

That seem unlikely to happen given Matterbridge config heavy approach, but it would be really nice if this appservice improvement could properly create ghost users Matrix side through the bridge.

nightmared:
Agreed on both counts!

As for the "proper ghost accounts", that's actually what it does:

poVoq:
I am not a Go deleloper and thus don't have an dev environment to compile this easily. But if you can share a x86_64 binary I am willing to test it.

42wim:
@nightmared thanks for your work! :)
Could you rebase on master? I've updated dependencies.

I'll try to test later

Mikaela:
Does this also bring support for IRC `NOTICE` -> Matrix `m.notice` (https://github.com/42wim/matterbridge/issues/1393) or should that be done in a separate PR?

I don't know if the previous library supported `m.notice`.

nightmared:
> I am not a Go developer and thus don't have an dev environment to compile this easily. But if you can share a x86_64 binary I am willing to test it in a IRC-Matrix-Discord setting.

Sure thing, here is one: https://nightmared.fr/files/matterbridge-2022-08-14.

You can verify its integrity with sha256sum: 
```
5d9511d784b6d4716af917026b79ab9921f65397ca8660baa40bce73b22c8f39  dev/matterbridge/matterbridge
```

Sorry about the time this took me, I have discovered the joys of COVID19 these past few days, and unsurprisingly that doesn't really help focusing on code (or anything really) ^_^
Anyway, I have rebased this on master, and I will take a look to see if I can tackle the CodeClimate issues.

nightmared:
> Does this also bring support for IRC `NOTICE` -> Matrix `m.notice` (#1393) or should that be done in a separate PR?
> 
> I don't know if the previous library supported `m.notice`.

No, I haven't added support for `m.notice` in this PR, because if I'm being honest, I didn't even knew this feature existed!
That being said, I don't think there should be any blocker for implementing this.

nightmared:
For golangci-lint, while I could reproduce locally the golangci-lint messages about the extraneous spaces in `// nolint`, I do not have these error messages even though I also have golangci-lint v1.48.0:
```
  Error: File is not `gofmt`-ed with `-s` (gofmt)
  Error: File is not `gofmt`-ed with `-s` (gofmt)
  Error: File is not `gofmt`-ed with `-s` (gofmt)
  Error: File is not `gofmt`-ed with `-s` (gofmt)
  Error: File is not `gofmt`-ed with `-s` (gofmt)
```
  
Does anyone happen to know why that is ?

Akselmo:
Hi, is this still being worked on? I have had so much problems with other bridges yet matterbridge works really well. Would love to have this feature.

nightmared:
@Akselmo Yes, this is definitely still ongoing :)

I'm currently waiting for user feeback to know if it works for other people, and if the current design seems acceptable or if I have to go back to the drawing board.

So if you happen to have some feedback, or to know people who might, I woul be happy to hear some or discuss this further!

Akselmo:
> @Akselmo Yes, this is definitely still ongoing :)
> 
> I'm currently waiting for user feeback to know if it works for other people, and if the current design seems acceptable or if I have to go back to the drawing board.

Ah, I could try building this with your PR and see if it works for me.

Akselmo:
Alright I got it to work, here's some feedback!

- I can't use the internal id's for channel names (probably has something to do with the appservice regex, changed it to `#.*` and all my channels work with it)
- Avatars do not get over the bridge in either direction. The stable build has the webhook support and the avatars get over bridge to Discord at least. I tried with `localhost` and my matrix domain in the appservice settings. No dice.
  - To add, the bridgebot is sending its *own* avatar from matrix to discord, instead of the user.
- The `RemoteNickFormat="<{NICK}> "` can't be anything else but that. I tried to remove the `<>` but it stopped working. I tried to remove them because the nick looks like this: `&lt;Aks&gt;`

Anyhow so far so good and I prefer this way much more over the other bridge bots I've tried.

nightmared:
Thanks for the feedback!

- For the internal ID, that's concerning. Could you tell more about what it meant for the rooms IDs to not work? To you not receive messages?
- For the avatar, this is a known problem: https://github.com/mautrix/go/pull/83. This should be fixed once mautrix-go will have published a new version of the library, that we would apply by updating the vendored dependency.
- Not being able to change the Nickname format is surprising. I will perform some tests locally to see if I can reproduce (and fix the `<` -> `&lt;` conversion) - probably not today though.

Akselmo:
>     * For the internal ID, that's concerning. Could you tell more about what it meant for the rooms IDs to not work? To you not receiving messages?

If the room ID started with `!randomlettershere:address.org` it wouldn't work: It couldn't find the address. But I think this was because the appservice regexp has `#` in the beginning, so it was trying to look for `#!randomlettershere" and failed because of that.

It's not *that* big of a deal since using the public address of the channels worked. I tend to use the internal addresses/ID's in case I change the public address, so it still keeps pointing to right place.

>     * Not being able to change the Nickname format is surprising. I will perform some tests locally to see if I can reproduce (and fix the `<` -> `&lt;` conversion) - probably not today though.

I could try changing it again, it was a bit odd. Sometimes when closing the bot and then starting it up again it wouldn't launch the appservice and worked in the non-appservice way: Instead of ghost accounts it would just relay messages with `<name> bla` way.

Akselmo:
I also noticed custom emojis wont go over the bridge either. I guess this is also something to do with Mautrix?

nightmared:
What do you mean by "custom emojis"? Reactions to messages?
If you mean reactions, AFAIK they weren't shared across bridges prior to this MR either (but I could be wrong of course).

Akselmo:
> What do you mean by "custom emojis"? Reactions to messages? If you mean reactions, AFAIK they weren't shared across bridges prior to this MR either (but I could be wrong of course).

I mean actual emojis and stickers.

Here's an example of Discord custom emoji to Matrix

Discord:
image

Matrix:
image


And from Matrix to Discord

Matrix
image

Discord
image

This bridge has managed to solve this issue, so maybe theres something we could use? https://github.com/matrix-org/matrix-appservice-discord

---

Also, something new I noticed!

After enabling appservice, someone on the Matrix channel has to talk first for the appservice to be active.

image

After enabling bot, the first message there came from Discord, without the appservice. But after I said something in the Matrix channel, then tried again from Discord, the appservice was enabled.

Akselmo:
Also it seems I can't delete messages on discord from matrix side.

nightmared:
Thanks for all these reports, I will take a look at them (mostly this week-end I think)!

Akselmo:
No problem, I don't know Go but I can also look into some issues if I got time.

nightmared:
> After enabling appservice, someone on the Matrix channel has to talk first for the appservice to be active.
>
> After enabling bot, the first message there came from Discord, without the appservice. But after I said something in the Matrix channel, then tried again from Discord, the appservice was enabled.

Yes, that is a bit sad but I have seen that issue too. This is a consequence of the appservice design: the server only pushes events to the application service when there is new messages. Which means at least one message must be received before the application service kicks in and replace the classic syncer for handling messages from the room.

The same logic means we can receive twice the first message send on a matrix room after a bridge restart: one from the classical syncer and a second one from the appservice. This lead to that first message being sent (e.g. to a Discord gateway) and then being resent instantly (the message have the same ID, so Discord actually show the message as "(edited)" instead of showing it twice).

Work remaining on this subject:
- [ ] look into handling graciously appservice failures (if the server is reconfigured to disable the application service, right now you need to restart the bridge for it to start treating messages from the "classical syncer" again - we mute the classical syncer for rooms where we already receive events. This ensures we don't treat events twice)
- [ ] test if we could force the server to ping the appservice listener on every room, hence solving the "first message" boostraping problem described earlier (by sending some fake event like a typing notification maybe?)
- [ ] try to detect and fix the first matrix message being sent twice

nightmared:
Alright, to sum up the work done or remaining to do:
- [x] Adding bidirectionnal avatar support
- [x] Convert transparently between IRC NOTICE and matrix] m.notice messages
- [x] Fix the deletion of messages from [matrix] when received over application services
- [x] Fix various issues with the displayname/avatar caching
- [ ] Add the option to have a different "Nickname template" for appservice messages and for messages sent over the "classical syncer"
- [ ] emojis and stickers support (cf. https://github.com/42wim/matterbridge/pull/1864#issuecomment-1234552002)
- [ ] look into handling graciously appservice failures (if the server is reconfigured to disable the application service, right now you need to restart the bridge for it to start treating messages from the "classical syncer" again - we mute the classical syncer for rooms where we already receive events. This ensures we don't treat events twice)
- [ ] test if we could force the server to ping the appservice listener on every room, hence solving the "first message" boostraping problem described earlier (by sending some fake event like a typing notification maybe?)
- [ ] detect and fix the first matrix message being sent twice -> could be done by merging gracefully messages from the application service and the "classical syncer", which would also be useful for the first caveat described below. This will probably require even more caching ;)
- [ ] support more complex regexps for the usernames in the application service configuration file
- [ ] testing this patcch series enough to be confident it works as expected
- [ ] document better the options/how to configure the application serviced

Planned, then rejected:
- Accepting rooms not explictly hardcoded in the config (that is the whole point of app services after all, yet currently it only relays messages for rooms that are already defined in the config).
  Rationale: not in the spirit of matterbridge explicit configuration

Known caveats:
- If the bridge is down while the homeserver receives messages on bridged room, the homeserver will apply some exponential backoff trying to reach the application service endpoint (https://github.com/matrix-org/synapse/blob/d736d5cfadcc9a56523fcb1cfe8cb1d2be47a4ec/synapse/appservice/scheduler.py#L489-L493), which means that the bridge will only use the classical syncer for up to 9 minutes. Worse, as we disable message receiving over the classical syncer if the application service worked at some point since the bridge started, this means that the bridge would be non-functional for that duration!

Thanks to @Akselmo for the feedback!

Akselmo:
> Thanks to @Akselmo for the feedback!

No problem. :) 

>Accepting rooms not explictly hardcoded in the config (that is the whole point of app services after all, yet currently it only relays messages for rooms that are already defined in the config).
Rationale: not in the spirit of matterbridge explicit configuration

Personally I don't mind this at all. I prefer having explicit control over what is bridged and what is not. Helps with channels that shouldn't be shared from one system to another.

> test if we could force the server to ping the appservice listener on every room, hence solving the "first message" boostraping problem described earlier (by sending some fake event like a typing notification maybe?)

This could work. Alternatively, sending a message "Brige is now enabled" could work. Could also make the bot delete the message after some time. This is not a *good* solution however since it'll then show notifications for channels even after deleting the message, at least on Discord side. The Matrix side of the bot could just say "Re-enabling bridge" as deletion reason as well.

Anyhow, will gladly do more testing, just ping me here and I'll help. :)

42wim:
> For golangci-lint, while I could reproduce locally the golangci-lint messages about the extraneous spaces in `// nolint`, I do not have these error messages even though I also have golangci-lint v1.48.0:
> 
> ```
>   Error: File is not `gofmt`-ed with `-s` (gofmt)
>   Error: File is not `gofmt`-ed with `-s` (gofmt)
>   Error: File is not `gofmt`-ed with `-s` (gofmt)
>   Error: File is not `gofmt`-ed with `-s` (gofmt)
>   Error: File is not `gofmt`-ed with `-s` (gofmt)
> ```
> 
> Does anyone happen to know why that is ?

It seems that the last linter issues is because you have a space between `//nolint: somelinter` instead of `//nolint:somelinter`

nightmared:
Note: I rebased on top of master, which includes #1875, but I just threw away these modifications (read: I haven't bothered to re-implement them yet). Please let me know (once the discussion over there about whether this generates too much state events or not is settled) what would be preferred.

nightmared:
The last commit adds sticker support, but only in one direction, as I don't have a Discord Nitro account to test/implement the other side.

@Akselmo you can test the new version if you want, but the issues outlined in https://github.com/42wim/matterbridge/pull/1864#issuecomment-1236121443 have not been fixed so far, so not much have changed!

nightmared:
Sorry, I haven't had time to work on this the last few weeks, and probably won't for another week or two :/

(But I'm still commited to making this merged someday !)

nightmared:
I am about to restart working on this issue, do not consider the last push, it is only a rebase against master.

nightmared:
> You think this will be ready for a merge soon?

Sorry for the late reply!

https://github.com/42wim/matterbridge/pull/1864#issuecomment-1236121443 is an adequate status of the progress of this PR.
However, given the (lack of) time I spend on this issue, this PR will probably never be merged if we wait for all these enhancements to be developed.

As such, I believe it would be best to merge the PR as-is (fixing eventual errors and style issues of course), and to open dedicated PR for each of these issues, considering they are enhancements that we would love to see, and not regressions.

This would allow other developers to look into adding support for them, which is understandably not very feasible with the current PR.

What do you think?

poVoq:
Can you please share another binary for testing? I am running into probably very basic issues in compiling it that are likely just due to me not being used to the go tool-chain at all and I would rather spend the time actually testing the bridge than trying to debug go compilation which I am not using at all otherwise. Thanks!

gabrc52:
How do you enable puppeted users? Messages seem to come from the normal bot user

poVoq:
> How do you enable puppeted users? Messages seem to come from the normal bot user

This has some hints: https://github.com/42wim/matterbridge/pull/1864/files#diff-0b4fb334dec5e2cf5c073565d97d11ffb59c157269ff451f25df8ef26fac4d9f

But so far I didn't manage to fully configure it due to a lack of time. Probably tomorrow.

gabrc52:
Here are some notes from my testing:

Because the non-Matrix platform I use is Mattermost, I will say "mattermost", but it likely means any non-Matrix platform that Matterbridge supports.

## 1: user puppeting

* If you send a message on mattermost (non-matrix) first, then messages are bridged as a normal user
* If you send a message on matrix first, then a message on mattermost, messages are correctly bridged as a puppeted user

Steps to reproduce:
1. Start the bridge
2. Send a message on the non-Matrix platform
3. The message is bridged to Matrix as a normal user (bug)
4. Stop the bridge
5. Start the bridge
6. Send a message on Matrix
7. Send a message on the non-Matrix platform
8. The message is bridged to Matrix as a puppeted user, as expected

It's probably this item in the todo list: 

>  test if we could force the server to ping the appservice listener on every room, hence solving the "first message" boostraping problem described earlier (by sending some fake event like a typing notification maybe?)

I am not sure why that workaround would be needed. The appservice should be listening to messages on the claimed prefixes, no matter what, and sending messages as bridged users is completely independent from the REST API of the appservice.

Sending messages as bridged users should only involve making the same queries to the Matrix homeserver REST API, [but with a new parameter on the query string: `?user_id=@_irc_user:example.org`](https://spec.matrix.org/v1.6/application-service-api/#identity-assertion), except instead of logging in with the normal user (bot) account token, you use the `as_token`.

Might this be a bug in `mautrix-go`? Would it be worth asking in the Maunium room? (I've never touched the Mautrix library but I've built a bridge in Python without a library).

## 2: images

PDF file was properly bridged from Mattermost to Matrix
Sending images works on Matrix->Mattermost but not Mattermost->Matrix

* Attempting to send an image from Mattermost->Matrix without a caption sends 2 empty messages
* Attempting to send an image from Mattermost->Matrix with a caption sends the caption first then an empty message)
* It appears that images do not properly upload. It appears that the content of the “image” that is sent is just the string “{}”

Element (seems completely blank):

image

Fluffychat (attempts to load image but fails because it only contains `{}`)

image

## 3: deletions

* Bidirectional edits WORK
* Deletions partially work (Matrix->Mattermost turns into “deleted”, Mattermost->Matrix nothing happens)

poVoq:
@gabrc52 can you share your config for the appservice part? I finally got time to try this and the non-appservice part works for Matrix <-> IRC, but even after enabling the appservice it continues sending it as a bot (yes I tried sending from Matrix first). Sadly nothing in the logs either side and matterbridge starts up fine without complaining about a problem with the appservice part.

I am not a big expert with appservice configurations though, so maybe I did something wrong in the config files.

Thanks!

gabrc52:
@poVoq my file is

```yaml
id: "matterbridge"
url: "http://127.0.0.1:29777"
as_token: "REDACTED"
hs_token: "REDACTED"
sender_localpart: "matterbridge"
namespaces:
  users:
    - exclusive: true
      regex: "@_mattermost_.*"
  aliases:
    - exclusive: false
      regex: "#_mattermost_.*"
  rooms: []
push_ephemeral: true
de.sorunome.msc2409.push_ephemeral: true
```

`hs_token` and `as_token` can be generated with `pwgen -s 64 1` and must be different from each other.

Since it is only listening to rooms starting with `#_mattermost_`, to have the integration work, you must set their alias to something starting with that (or with the prefix you choose)

poVoq:
Thanks a lot for sharing your config!

> Since it is only listening to rooms starting with `#_mattermost_`, to have the integration work, you must set their alias to something starting with that (or with the prefix you choose)

Hmm, that would explain why it fails with the suggested config as well:

```
aliases:
   - exclusive: false
     regex: "#__bridge_.*"
```

But I am a bit surprised about this as Matterbridge is usually used to bridge existing channels that are not named in a special way like this. Can I just remove that part for it to work with any Matrix channel specified in the Matterbridge config file? Or what would be a regex that works with any channel name?

> Off-topic: Also keep in mind if you just want to bridge IRC and Matrix, you're better off using a dedicated Matrix bridge. https://matrix.org/bridges/#irc. The purpose of this project, at least imho, is to have a powerful multibridge that works with as many platforms as possible. If you just have two platforms in mind you want to bridge, the Matrix ecosystem offers multiple bridges.

Yes I do run an appservice-irc already... the problem arises from double bridging multiple IRC channels (and a specific ToS clause of libera.chat) that requires an alternative bridge for now.

gabrc52:
> Can I just remove that part for it to work with any Matrix channel specified in the Matterbridge config file? Or what would be a regex that works with any channel name? Edit: or do I need to create local alias names for each room to be bridged?

I believe you can either set aliases to a catch-all regex (`.*`) or add

```
rooms:
   - exclusive: false
     regex: ".*"
```

to bridge all rooms (which would include DMs as well). I have not tested it, though.

> Matterbridge is usually used to bridge existing channels that are not named in a special way like this

Yeah, this is showing one of the differences between Matrix [app services](https://spec.matrix.org/v1.6/application-service-api/) and bots with accounts. Since this PR is still under development, it could be improved so that for rooms that don't match the regex, it receives messages using the bot method, and sends them using the app service method.

There's other Matrix bridges with puppeted users that let you link any rooms without needing to have any specific alias ([such as the Discord one](https://t2bot.io/discord/) [matrix-appservice-discord](https://github.com/matrix-org/matrix-appservice-discord)), so this is not a Matrix restriction per se.

poVoq:
I noticed that Matterbridge fails to start if the Synapse server is not immediately available. This can be of course fixed via adapting the systemd script, but I think given the http nature of Matrix and the inherent slowness of it, a bit more failure resistance would be cool.

gabrc52:
For those looking for a Mattermost<->Matrix bridge only, I have done another one: https://github.com/gabrc52/mattermost-to-matrix

poVoq:
Any news on merging this?

I have been using it for a while now and it seems to work fine.

nightmared:
Since I've not worked on this in a long time (nearly a year), and that the PR would have to be updated against code changes in "upstream" matterbridge, perhaps it would be best to close this PR?
Considering this PR haven't seen any update in that long period of time, I would think users are currently running multiple bridges (like matrix-appservice-discord and another bridge) when they need to bridge rooms on three or more platforms together.
Is there really any user of matterbridge with this PR?

poVoq:
I tried using it, but it didn't work for more than one channel at a time so I went back to other bridging options.

poVoq:
@antonizoon maybe this is something for your fork? It was almost working, just multi-room support was borked in my tests.

PR #1870, opened 2022-08-04
freenode to libera

Glandos:
The freenode network doesn't really work anymore

### DIFF
--- a/matterbridge.toml.simple
+++ b/matterbridge.toml.simple
@@ -1,7 +1,7 @@
 #WARNING: as this file contains credentials, be sure to set correct file permissions
 [irc]
     [irc.freenode]
-    Server="irc.freenode.net:6667"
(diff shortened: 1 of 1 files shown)


matterbridge.toml.simple
```diff
@@ -1,7 +1,7 @@
 #WARNING: as this file contains credentials, be sure to set correct file permissions
 [irc]
     [irc.freenode]
```
Kufat, line 3: You'll want to change this line as well since you changed account to "irc.libera" elsewhere.

Kufat (changes requested):

matterbridge.toml.simple
Glandos: You're right. I thought I already double checked, but it seems I was too confident…

42wim:
thanks

PR #1872, opened 2022-08-06
Enhance IRC join/part message support

Kufat:
A help channel on my IRC network uses matterbridge to link to a help channel on a related Discord. We've noticed a problem where people on the Discord end will respond to help requests from users who've already left. Enabling join/part message support is an option, but it'd be excessively chatty if every join/part was shown.

This adds a new feature that will selectively show join/part/nick change/quit messages if either a user was active recently or the message is inherently important (i.e. kicks and kills.) New config options are **ShowActiveUserEvents** and **ActivityTimeout**, explained below in the sample config. It also addresses an issue in the original code where a quit message handler would send an invalid channel in the bridge message because IRC QUIT events aren't populated with a channel name. (Neither are nick changes, which are also supported now.)

This is literally the first thing I've ever written in Go. I attempted to make it consistent with both your code style and language idioms, but please let me know if there's anything that needs to be addressed.

### DIFF
--- a/bridge/irc/handlers.go
+++ b/bridge/irc/handlers.go
@@ -296,6 +296,7 @@ func (b *Birc) handlePrivMsg(client *girc.Client, event girc.Event) {
 		b.activeUsers[event.Source.Name] = time.Now().Unix()
 	}
 	b.Remote <- rmsg
+	b.cleanActiveMap()
 }
 
 func (b *Birc) handleRunCommands() {
--- a/bridge/irc/handlers.go
+++ b/bridge/irc/handlers.go
@@ -80,33 +80,66 @@ func (b *Birc) handleInvite(client *girc.Client, event girc.Event) {
 	}
 }
 
+func isKill(quitmsg string) bool {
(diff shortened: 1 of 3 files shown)


Kufat:
Ignoring exhaustruct and snakecase complaints because the Message struct isn't initialized fully anywhere in the IRC code and because snakecase is canonical for IRC numeric enums.

42wim (commented):
Sorry for the delay in reviewing.
Very good for your first Go code :)

The `getPseudoChannel` is an issue, because now you're returning just the first channel that exists, but it's not sure that this specific user is on that channel. Right?

Kufat:
This needs a little work for some corner cases involving quits.

antonizoon:
Hi, I'm going to try to see if this feature can be merged, but it looks like it has conflicts. do you still use it and can it be updated to the latest master branch?

I'm going to start maintaining a fork of matterbridge to accept and merge pull requests regarding irc -> matrix -> discord since our organization needs this bridge to work.

However, I can only test irc -> matrix -> discord so will need more people to help provide pull requests proven to fix the whatsapp and telegram issues. If anyone else is able to help test whatsapp and telegram please provide tested pull requests.

https://github.com/bibanon/matterbridge/

PR #1889, opened 2022-09-05
Optionally suppress URL embeds for Discord

Kufat:


42wim:
So I've been thinking about this, and this seems like a tengo script can also solve.
My stance on adding new options to matterbridge is that if it can be done with a tengo script, it should be done that way.

Have you considered this as an option and it didn't work ? or maybe you didn't know about the scripting ?
More info on https://github.com/42wim/matterbridge/wiki/Tengo

Kufat:
I wasn't aware of the Tengo scripting feature, no.

It looks like Tengo scripts are global rather than per-gateway and don't have any access to the config, which wouldn't work for the use case needed on my network. (I'm sure I could bodge together something with various channel names and the desired behavior for each hardcoded, but that'd be inelegant.)

42wim:
@Kufat sorry for the late reply, you can match on the gateway in the tengo scripts, I think it should work?
Wrt access to the config, for this change you also need access to the config to enable it? It's the same as adding a script? or am I missing something ?

PR #1896, opened 2022-09-24
Add StripQuote option (telegram)

jx11r:
Quoted messages are a good option, but sometimes they flood the chat when there are too many line breaks.
image

With this PR:
image

Works better with `QuoteLengthLimit`:
image


42wim:
@jx11r I think this can be something that could better be fixed with tengo, instead of yet another option.

jx11r:
> @jx11r I think this can be something that could better be fixed with tengo, instead of yet another option.

You're right, thanks.

PR #1897, opened 2022-09-24
Fix empty messages on IRC

jx11r:
> image

instead of:

> image


jx11r:
Well, I just noticed this: https://github.com/42wim/matterbridge/wiki/Tengo
It should fix the empty messages on IRC, but I wonder why this isn't the default behavior. I mean that users wouldn't want empty messages by default (especially on IRC, where they look bad).

jx11r:
I could also fix the empty messages here: [outmessage.tengo](https://github.com/42wim/matterbridge/tree/master/internal/tengo/outmessage.tengo)

42wim:
I think it's valid to remove empty lines on irc.

PR #1903, opened 2022-10-23
discord: workaround not being able to name your webhook

55hgc:
Since ? ago probably a couple months at a guess you are not able to call a Discord webhook "Discord" or "Clyde"
Not sure if this affects matterbridge itself but it definitely breaks downstream projects like "go-discord-irc" that try to create a webhook named that and fail


42wim:
@55hgc did you test your code? I don't think those matches work correctly

PR #1909, opened 2022-11-03
Fallback to use API for getting channel ID when not in the cache

hloeung:
See https://github.com/42wim/matterircd/issues/479


hloeung:
Closing off, it's in the upstream project (didn't realise it was separate) - https://github.com/matterbridge/matterclient/pull/3

PR #1915, opened 2022-11-06
mumble: Implement sending of EventJoinLeave both to and from Mumble

s3lph:


42wim:
Thanks a lot for your work @s3lph

PR #1920, opened 2022-11-14
Fix telegram attachment comment formatting and escaping

sas1024:
Bugfix for telegram attachment comment formatting and escaping.

**How to reproduce this bug:**
You need bridge from slack to telegram.
Just send attachment with url in comment - message with attachment will not reach telegram.


42wim:
Thanks!

Issue #1469, opened 2021-05-10
Inline media in WhatsApp


ciroiriarte:
**Is your feature request related to a problem? Please describe.**
This would improve the user experience when integrating Telegram & WhatsApp

**Describe the solution you'd like**
I would like to be able to have inline/playable media instead of attached files.

**Describe alternatives you've considered**
Downloading files and playing with external players

**Additional context**
Tested with Android native clientes for sending (Telegram) & receiving (WhatsApp). This seems to work the other way around (receiving in Telegram looks inline).

Receiving audio notes and videos in WhatsApp from Telegram, I received them as file attachments:
file_1.mp4.mp4 for fideo
file_3.oga.ogg.oga for audio notes
animation.gif.mp4.mp4 for GIF animation

Tested with official container: "Running version 1.22.2-dev a0bca42a"

ilmaisin:
Also, if there is text in the same message with the media, the text is omitted and only the file lands in WhatsApp.

ilmaisin:
> Also, if there is text in the same message with the media, the text is omitted and only the file lands in WhatsApp.

Opened #1925 for this.

ilmaisin:
#1928 fixes this, so this can be closed.

PR #1928, opened 2022-11-24
Improve attachment handling

ilmaisin:


ilmaisin:
I could merge those video, audio, document and picture functions. With regards to the audio messages, maybe the contents of the ogg file should be examined in order to make sure it is really an OPUS audio file.

42wim:
LGTM, a follow-up PR for the opus would be nice :)

ilmaisin:
Closes #1469.

PR #1934, opened 2022-12-08
implement parsing of quotes/replies from whatsapp

ilmaisin:


ilmaisin:
Parsing of the replies of Whatsapp is entirely doable, but sending them from matterbridge to Whatsapp gets tricky, since WA wants the user ID of the original message in addition to the message ID.

yousefmansy1:
Hello @ilmaisin
I expanded on your PR (#1974) to start getting replies from other platforms to display on WhatsApp properly.

However I think I might be making the call to WhatsApp incorrectly.
Can you take a look?

yousefmansy1:
@42wim @ilmaisin I suppose we can close this one since #1974 got merged now.

PR #1927, opened 2022-11-20
Allow captions for document attachments, too

ilmaisin:
This seems to be sufficient to fix https://github.com/42wim/matterbridge/issues/1925

PR #1935, opened 2022-12-09
Handle incoming document captions from whatsapp

ilmaisin:
Whatsapp document attachments may now have captions. Basically #1928 or #1927 but to the another direction.


42wim:
Thanks!

PR #1972, opened 2023-02-16
Provide a multi-room multi-bridge example

BenWiederhake:
One of the questions most often asked in the support chatroom is along the lines of "How do I configure multiple rooms?" and "How can I run multiple bridges?".

This PR provides a sample config file that demonstrates how these features work. The idea is that in the future we can just point at this file; or perhaps it answers the question before it is even asked.


42wim:
Much appreciated, thanks!

PR #1974, opened 2023-02-18
Handle Whatsapp threading/replies

yousefmansy1:
Handle Whatsapp threading/replies, based on commit 6afa93e537c53b371db37f35a3546ff0fb669416 from #1934
In this change we are updating whatsapp message IDs to include sender JID as this is necessary to reply to a message 
- https://github.com/tulir/whatsmeow/issues/88#issuecomment-1093195237 
- https://github.com/tulir/whatsmeow/discussions/148#discussioncomment-3094325

image

### DIFF
--- a/bridge/whatsappmulti/handlers.go
+++ b/bridge/whatsappmulti/handlers.go
@@ -86,6 +86,12 @@ func (b *Bwhatsapp) handleTextMessage(messageInfo types.MessageInfo, msg *proto.
 		}
 	}
 
+	parentID := ""
+	if msg.GetExtendedTextMessage() != nil {
+		ci := msg.GetExtendedTextMessage().GetContextInfo()
+		parentID = getParentIdFromCtx(ci)
+	}
+
 	rmsg := config.Message{
 		UserID:   senderJID.String(),
 		Username: senderName,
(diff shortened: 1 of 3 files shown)


yousefmansy1:
@ilmaisin, have you tried sending messages back to WhatsApp?
I couldn't figure out how to get it to work. 
Based on your contribution history it looks like you know more than I do with the whatsmeow API.

yousefmansy1:
Finally figured out the issue.
I was 90% of the way there, you actually also need to add a `QuotedMessage` to the context info.

Only figured it out by taking a look at the mautrix-whatsapp repo and tracing through what they're doing.
https://github.com/mautrix/whatsapp/blob/e0a9b0e8bb25b70196b78079dca2eeef35d526b6/portal.go#L3719-L3726

yousefmansy1:
Replies should work across the board now.
@qaisjp do you think you can take a look?

Or perhaps @ilmaisin since you've made some commits to the whatsappmulti bridge.

(will post screenshots later its late I'm tired lol)

ilmaisin:
> Replies should work across the board now. @qaisjp do you think you can take a look?
> 
> Or perhaps @ilmaisin since you've made some commits to the whatsappmulti bridge.
> 
> (will post screenshots later its late I'm tired lol)

Should your PR now work in both directions? I can probaly try it this weekend. I have my draft PR running in my instance, but it is rx-only.

yousefmansy1:
> 
> Should your PR now work in both directions? I can probaly try it this weekend. I have my draft PR running in my instance, but it is rx-only. 

Yes it works in both directions. 👍

ilmaisin:
If a Telegram user replies to a message orignating in Whatsapp, the reply that is sent back to Whatsapp includes both the "proper" quote as well as a textual quote. This may be a matter of another PR, though.
matterbridge-wa-quote

yousefmansy1:
> If a Telegram user replies to a message orignating in Whatsapp, the reply that is sent back to Whatsapp includes both the "proper" quote as well as a textual quote. This may be a matter of another PR, though. matterbridge-wa-quote

I believe this might be unavoidable, as then you have `QuoteDisable=false` it will append reply information directly into the message text. 
The receiving bridges will simply relay whatever is in the message.Text body.
https://github.com/42wim/matterbridge/blob/24f6747516a9a80f82068656df50b3972928ba56/bridge/telegram/handlers.go#L113-L119

This would require some more significant rework of how replies work to resolve.

For now you'll just have to use `QuoteDisable=true`

42wim:
Great job @yousefmansy1 !

PR #1976, opened 2023-02-19
Return a message ID when sending with a webhook.

### LINKED ISSUE #1975: Discord image threading is broken
**Describe the bug**
Sending a reply to an image from a remote does not relay

**To Reproduce**
Send a File from a remote -> discord.
Replies from remote -> discord are not relayed
Replies from discord to discord -> remote do not preserve threading

**Expected behavior**
Threading is preserved when relaying, and replies should bridge correctly.

**Screenshots/debug logs**
image
```
[0019] ERROR gateway:      [handleMessage:gateway/handlers.go:228] SendMessage failed: HTTP 400 Bad Request, {"code": 50035, "errors": {"message_reference": {"message_id": {"_errors": [{"code": "NUMBER_TYPE_COERCE", "message": "Value \"telegram 729\" is not snowflake."}]}}}, "message": "Invalid Form Body"}
```
- OS: linux/docker
- Matterbridge version: 1.26.1

**Additional context**
```
[discord]
	[discord.throwaway]
		Server="xxxxxxxxxxxxxxxxxxxxx"
		Token="xxxxxxxxxxxxxxxxxxxxx"
		RemoteNickFormat="{NICK}\n"
		AutoWebhooks=true
		UseLocalAvatar=["telegram.bot"]
		ShowEmbeds=true
		EditSuffix=" "
		PreserveThreading=true
		QuoteDisable=true

[telegram]
	[telegram.bot]
		Token="xxxxxxxxxxxxxxxxxxxxx"
		RemoteNickFormat="<b>{NICK}:</b>\n"
		MessageFormat="HTMLNick"
		DisableWebPagePreview=false
		EditSuffix=" "
		UseFullName=true
		PreserveThreading=true
		QuoteDisable=true
		MediaConvertWebPToPNG=true

[whatsapp]
	[whatsapp.googlevoice4] # ! 2/5
		Number="xxxxxxxxxxxxxxxxxxxxx"
		SessionFile="./session-xxxxxxxxxxxxxxxxxxxxx-4.gob"
		QrOnWhiteTerminal=false
		QuoteDisable=true

[[gateway]]
name="test"
enable=true

	[[gateway.inout]]
		account="discord.throwaway"
		channel="ID:xxxxxxxxxxxxxxxxxxxxxx"

	[[gateway.inout]]
		account="whatsapp.googlevoice4"
		channel="xxxxxxxxxxxxxxxxxxxxxx@g.us" # annoucements chat
	
	[[gateway.out]]
		account="whatsapp.googlevoice4"
		channel="xxxxxxxxxxxxxxxxxxxxxx@g.us" # test chat

	[[gateway.inout]]
		account="telegram.bot"
		channel="-xxxxxxxxxxxxxxxxxxxxxx"
```

yousefmansy1:
Return a message ID when sending with a webhook

Resolves #1975

Error came from that a message ID was not being returned when sending with a message ID.

image


bridge/discord/webhook.go
```diff
@@ -84,7 +84,7 @@ func (b *Bdiscord) webhookSend(msg *config.Message, channelID string) (*discordg
 			}
 			content := fi.Comment
 
-			_, e2 := b.transmitter.Send(
+			res, err = b.transmitter.Send(
```
yousefmansy1, line 87: Hypothetically we will overwrite a result and error from a sent message from line 63. 
However, this should not be possible as the text for a message sent stays in the file caption not the `msg.Text`

qaisjp: This might depend on the bridge that created the `config.Message`, and produce an issue for bridges that actually do set `msg.Text`.

I think we need a better one-to-many relationship here, instead of assuming that 1 message on the sending-bridge side corresponds to 1 message on the receiving-bridge side.

yousefmansy1: Yeah in that regard this would be an incomplete fix for the existing issue.

Also if there are multiple files in the Extra["file"] then we'd also only return the reference of the last file sent.
In order to fix this we'd have to refractor this all send methods to return an array of message id's rather than a single entry, something that would require some larger refactoring.

42wim: A quick (dirty) fix is to 

```suggestion
			res2, err := b.transmitter.Send(
			...
			if msg.Text == "" {
			    res = res2
			}
```
This needs to be handled better but needs quite a refactoring to have this done correctly, but if this fixes some of the issues it's fine by me.

yousefmansy1:
Updated with your fix @42wim 
I'll need to create an issue tracking this problem where one message entity is sent as more than one message on a destination.

Without it it will break break threading and message deletes with some dangling messages.

PR #1979, opened 2023-02-22
Use AuthorSignature`as a fallback username

yousefmansy1:
Use `AuthorSignature` as a fallback username.

This comes in handy for announcement type channels where neither the `SenderChat` or `From` structs contain name information.

Also Tweak username logic as when using a full name the username field can be " " instead of "", lead to a bug where name would be empty instead of "unknown".

Original error found when sending from telegram discord webhook would complain about invalid name. Other platforms would show no name. 
```log
time="2023-02-22T02:00:28Z" level=error msg="Could not send text (t) for message &config.Message{Text:"t", Channel:"ID:986795919334866995", Username:" \n", UserID:"-1001896329849", Avatar:"", Account:"telegram.bot", Event:"", Protocol:"telegram", Gateway:"test", ParentID:"", Timestamp:time.Date(2023, time.February, 22, 2, 0, 27, 970481068, time.Local), ID:"", Extra:map[string][]interface {}{}}: execute failed: HTTP 400 Bad Request, {"code": 50035, "errors": {"username": {"_errors": [{"code": "BASE_TYPE_BAD_LENGTH", "message": "Must be between 1 and 80 in length."}, {"code": "USERNAME_INVALID", "message": "Username cannot be \"\""}]}}, "message": "Invalid Form Body"}" prefix=discord
```

Post fix:
image

**Note**: requires `Sign Messages` enabled
image

Otherwise shows `unknown`: (Previous behavior would just show nothing when using `UseFullName`)
image


42wim:
Thanks again :)

PR #1986, opened 2023-02-26
Support handling delete message from whatsapp

yousefmansy1:
Support handling delete message from whatsapp.

This is where the pull request's start to get messy.
The message ID's here can depend on the status of #1974

PR #1991, opened 2023-02-27
Allow message deletes from any platform

yousefmansy1:
Allow message deletes from any platform

This allows for message deletes to be synced from any of the gateway's channels, not just the "canonical"/source channel.

Note, the required changes in whatsappmulti handler depends on #1974 
This is because when deleting a message that is not from the matterbridge user the sender jid is also required.

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -466,7 +466,8 @@ func (gw *Gateway) SendMessage(
 
 	// exclude file delete event as the msg ID here is the native file ID that needs to be deleted
 	if msg.Event != config.EventFileDelete {
-		msg.ID = gw.getDestMsgID(rmsg.Protocol+" "+rmsg.ID, dest, channel)
+		canonicalMsgID := gw.FindCanonicalMsgID(rmsg.Protocol, rmsg.ID)
+		msg.ID = gw.getDestMsgID(canonicalMsgID, dest, channel)
 	}
 
 	// for api we need originchannel as channel
--- a/bridge/whatsappmulti/whatsapp.go
+++ b/bridge/whatsappmulti/whatsapp.go
@@ -361,7 +361,7 @@ func (b *Bwhatsapp) Send(msg config.Message) (string, error) {
 			return "", nil
 		}
 
-		_, err := b.wc.RevokeMessage(groupJID, msg.ID)
+		_, err := b.wc.SendMessage(context.Background(), groupJID, b.wc.BuildRevoke(groupJID, extendedMsgID.Sender, extendedMsgID.MessageID))
 
 		return "", err
 	}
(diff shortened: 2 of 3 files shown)


gateway/gateway.go
```diff
@@ -466,7 +466,8 @@ func (gw *Gateway) SendMessage(
 
 	// exclude file delete event as the msg ID here is the native file ID that needs to be deleted
 	if msg.Event != config.EventFileDelete {
-		msg.ID = gw.getDestMsgID(rmsg.Protocol+" "+rmsg.ID, dest, channel)
+		canonicalMsgID := gw.FindCanonicalMsgID(rmsg.Protocol, rmsg.ID)
```
yousefmansy1, line 469: to be honest this is a bit of a hacky workaround and I don't necessarily like doing this pretty expensive operation on basically every message recived.

Hypothetically we can work around this 

In practice this inefficiency would only matter in really large message caches (and we're mostly limited by I/O bandwidth).

I still don't really like it though.
I'd rather it be more of a graph-like structure where any message maps to the rest of the messages rather than 1 canonical message to the sent mesages.

yousefmansy1: This is less of an issue with the persistent message maps in #1996

ilmaisin:
What this should exatly do and do not do? I have this PR applied in my instance, but cannot observe any propagation of message deletions between a Whatsapp and Telegram group.

yousefmansy1:
> What this should exatly do and do not do? I have this PR applied in my instance, but cannot observe any propagation of message deletions between a Whatsapp and Telegram group.
> 
> Edit: And what the relation with #1986 is? Should I have that too applied for it to work?

Yes you will have to have  #1986 merged in as well as WhatsApp deletes wont propagate at all since those events are not handled.

(telegram APIs dont have events for any message deletes so that's also inevitable.)

In theory this should work as:
1. Message is sent from platform A
2. Message is deleted from platform B
3. Message is deleted on all platforms.

If you want to test this PR without merging  #1986 as well
try sending a message from telegram to discord.
delete the message from discord, observe original message on telegram is deleted (make sure bot is admin)

yousefmansy1:
> I merged that PR. When I delete a message sent by myself from Whastapp, something seems to appear according to log, but it doesn't get deleted from Telegram. My bot should be admin. How could I troubleshoot this?

Hmm, can you share the log that you said appears?

Presumably if you send a message from WhatsApp, then delete it on WhatsApp, it will delete on telegram?

yousefmansy1:
To expand, this CR is a relatively simple hack in which when it gets all the corresponding message IDs with a message those are only stored as a map from original ("canonical") message  -> messages[]
there is a method that already existed will find the original message sent for any message.

In reality what we are doing is almost pretending the message delete came from the source platform (at least with regards to the message map.) 

In #1996 I have two separate maps, one to store the mapping of message -> canonical message and another that does canonical message -> messages[]
We're getting O(1) complexity for some extra storage usage, in this case a worthy trade off as the data we're dealing with is quite light.

PR #1997, opened 2023-03-01
Add Platform Agnostic Commands

yousefmansy1:
Add opt/in/out commands

Adds some commands that can be run regardless of the platform sent on (even works with DMs to the bot)


42wim:
If features can be done using Tengo script I'd rather have it done that way instead of adding extra features in matterbridge.
See https://github.com/42wim/matterbridge/wiki/Tengo

Ofcourse you can't use bbolt db, but if you really need state you can save it in a json file for example

yousefmansy1:
@42wim  this branch is a bit of a mess rn but to answer your question.
No, this cant be done with tango scripting.
Its much more than that.

bridge/config/config.go
```diff
@@ -179,9 +180,10 @@ type Protocol struct {
 }
 
 type ChannelOptions struct {
-	Key        string // irc, xmpp
-	WebhookURL string // discord
-	Topic      string // zulip
+	Key            string // irc, xmpp
+	WebhookURL     string // discord
+	Topic          string // zulip
+	WelcomeMessage string // all
```
yousefmansy1, line 186: this isnt actually being used will remove

PR #2000, opened 2023-03-05
Add proper support for Mumble servers with no message length limit and fix not being able to embed from Mumble to other platforms

mvoolt:
I've also removed SetSelfMuted because it would send 2 event messages with that in.


bridge/mumble/mumble.go
```diff
@@ -249,7 +249,7 @@ func (b *Bmumble) processMessage(msg *config.Message) {
 
 	// If there is a maximum message length, split and truncate the lines
 	var msgLines []string
-	if maxLength := b.serverConfig.MaximumMessageLength; maxLength != nil {
+	if maxLength := b.serverConfig.MaximumMessageLength; *maxLength != 0 {
```
42wim, line 252: Why do you remove the nil check? If the pointer is nil, it'll panic

mvoolt: I wasn't aware of that. This is my first time doing Golang.

42wim: No problem, why did you close it?

mvoolt:
The merge branch commit felt inappropriate so I had to redo the pull request: #2008

PR #2008, opened 2023-03-09
Add proper support for Mumble servers with no message length limit and fix not being able to embed from Mumble to other platforms

mvoolt:
I've also removed SetSelfMuted because it would send 2 event messages with that in.


42wim:
@mvoolt looks good, the linter has a complaint though, can you run `gofmt -s`  on `mumble.go` ?

mvoolt:
@42wim Done, This is my first time doing golang.

42wim:
👍

PR #2019, opened 2023-03-15
support telegram chat Join/Leave updates

yousefmansy1:
Support telegram chat Join/Leave updates.

For some reason during my testing the only non empty field for the user's name is the first name.

Also optimistically we'd reuse the logic from `handleUsername()`.


yousefmansy1:
@42wim fixed the gofumpt issues.
I'll try to make sure I run the `golangci-lint` stuff before pushing to avoid this lol

42wim:
Thanks @yousefmansy1 !

yousefmansy1:
> Thanks @yousefmansy1 !

When I was using this on my testing branch there was actually a bug where message can be nil!!!

will add a simple nil check for this, as it will cause matterbridge to crash.

PR #2029, opened 2023-03-30
Add Tengo script to autolink on Mumble

mvoolt:
I've thought of adding this to the internal Tengo script but I'll just upload it to contrib folder for now.


42wim:
Thanks!

PR #2035, opened 2023-04-03
Remove gitter bridge

42wim:
See https://blog.gitter.im/2023/02/13/gitter-has-fully-migrated-to-matrix/


Mikaela:
FWIW the README.md and GitHub project description still refer to Gitter

42wim:
@Mikaela well, technically it still supports gitter, I think it can stay there, like we support twitch too (but it's also just irc)

ajhalili2006:
Maybe the wiki should be updated too to reflect backend changes?

PR #2036, opened 2023-04-04
[Bugfix] Add nil check for group update in telegram

yousefmansy1:
Fix bug in telegram handler from #2019 

Some group updates actually have a nil message so the current usage may throw an exception.


42wim:
👍

PR #2077, opened 2023-09-16
Fix typos in matterbridge.toml.sample

Quantum-0:
Hello! I was exploring this example config file and found the word "syntaxa" in 10 places. I suppose it is a typo and there should be "syntax" instead of "syntaxa"


42wim:
👍

PR #2090, opened 2023-10-23
irc: do not bridge CTCP commands other than ACTION

patrickxia:
CTCP commands other than ACTION are designed for client-to-client interaction on IRC networks. Drop such messages when we receive them.

Also get rid of a "CTCP_ACTION" handler in the handler registration. The registration

1) can't do anything (if anything, we wanted the string constant
   girc.CTCP_ACTION, which is "ACTION")
2) doesn't do anything in this context, because CTCP handlers are
   registered separately:
   https://github.com/lrstanley/girc/blob/f47717952bf9258e02eac14f1b9723bcf084e618/ctcp.go#L205

The PRIVMSG handler already listens to all CTCPs.

Thanks to @lexande for the bug report (found on a live instance).


42wim:
Thanks 👍

Issue #2104, opened 2023-11-23
Discord bridge does not delete messages when they're deleted on Matrix when using webhooks


thestinger:
When we delete messages on Matrix, they don't get deleted on Discord when webhooks are enabled. This works fine with webhooks disabled. The permission gives it the ability to delete webhooks so either this functionality is missing or it's broken.

thestinger:
Renamed the issue since this isn't specific to Matrix -> Discord but rather happens with anything.

PR #2109, opened 2023-12-13
Add support for incoming webhook deletions on Discord

cryox-dev:
Resolves https://github.com/42wim/matterbridge/issues/2104


thestinger (approved):
This works well and we use it in production for GrapheneOS.

PR #2123, opened 2024-03-07
Correctly split three-or-more byte sequences of UTF-8

BenWiederhake:
The underlying bug was the assumption that [`uft8.DecodeLastRuneInString`](https://pkg.go.dev/unicode/utf8#DecodeLastRuneInString) returns some kind of number of bytes that, when stripped from the end, leaves the string with a correct ending.

In reality, this function always returns the constant value 1 if the last rune is not valid.

Therefore, if there are two or more partial bytes of a three-or-more byte rune, this used to give the wrong result.

Found while trying to implement a related feature.


42wim:
Thanks 👍

PR #2125, opened 2024-03-20
fix: update post types and include system removals in skip logic

NeilHanlon:
I believe that previously, these post types were grouped in these three categories, but have since been broken out to be more verbose.

I ran into this today as we have join/part as well as adding/removal from teams turned _off_, but found that when an Admin removes a user from the team, a different message type is generated.


42wim:
Thanks!

PR #2138, opened 2024-04-27
Clear existing IRC event handlers before connecting to new ones

bd808:
Clear IRC event handlers that we will be registering for the new connection before registering new handlers. This prevents duplicate event handlers in the case where we are connecting via a BNC and are seeing a reconnect. Attempting to clear handlers when none have been set is a no-op.

Fixes 42wim#1564
Co-authored-by: Andreas Vögele <andreas@andreasvoegele.com>


42wim:
LGTM, thanks!

PR #2143, opened 2024-05-14
Move deprecated/inactive 3rd party bridges to their own sections

erentar:
Most of the bridges in the 3rd party section were deprecated or inactive for longer than 2 years. I moved these into their own section such that other people trying to set up bridges do not waste time clicking on all of these to see which ones worked


42wim:
Thank you!

PR #2165, opened 2024-07-10
Explicitly announce the sender of each file in Telegram, simplifying moderation

BenWiederhake:
Before:
Bildschirmfoto_2024-07-10_20-25-33

After:
Bildschirmfoto_2024-07-10_20-26-14

This feature is off-by-default, and can be enabled whenever it's needed.

I'm sure there are some situations where it *shouldn't* create an additional message, but this solves my problem for now. Maybe you find it helpful, too.


bridge/telegram/telegram.go
```diff
@@ -159,6 +159,11 @@ func (b *Btelegram) Send(msg config.Message) (string, error) {
 		}
 		// check if we have files to upload (from slack, telegram or mattermost)
 		if len(msg.Extra["file"]) > 0 {
+			if !b.GetBool("AnnounceFileSender") {
```
42wim, line 162: shouldn't that be `if b.GetBool` ?

BenWiederhake: Indeed!

BenWiederhake:
Changes since last push:
- Rebased
- Fixed a boolean inversion bug

PR #2173, opened 2024-08-19
get rid of vendor directory

siddarthkay:
## Summary

checking in the `vendor` folder for a `golang` project is similar to committing `node_modules` in a `js` project.
This PR gets rid of the vendor directory from this repository and adds `/vendor` to `.gitignore`

## Why

Having a `vendor` directory checked in makes reviewing and rebasing upgrading PRs with a `golang` bump or a dependency upgrade a nightmare due to the amount of files changed. 
Since most IDEs are able to generate the `vendor` directory automatically I think its better to not check in the `vendor` directory.

I have no clue about the maintainers take on this opinion but I wanted to start the conversation with this PR.
Thank you for your time and patience 🙏🏻


siddarthkay:
@42wim : what do you think about this proposal ?

IntGrah:
Isn't the `vendor` directory is usually committed to ensure reproducible builds? It's nowhere near as heavy as `node_modules`

siddarthkay:
@IntGrah : I agree its not heavy but the number of files are insanely large.
Rebasing and upgrading go versions is a nightmare.
Look at the diff of this PR.

Screenshot 2024-08-27 at 4 16 06 PM

for reproducibility purposes there are package managers like `nix` that do the job. Committing the `vendor` folder doesn't help much for that use case I believe.

42wim:
Thanks, but the vendor folder is there for reproduceability and when upstream libraries would suddenly be gone, it's not going to get removed.

jakubgs:
You could just use forks of these dependencies if you worry about original repos disappearing. I do think committing `vendor` folder is a bad practice, but if it solves a problem for that's fine.

Issue #2175, opened 2024-08-21
telegram whatsapp bridge: [Client ERROR] Client outdated (405) connect failure


fabio841:
0000]  INFO main:         Running version 1.26.0 
[0000]  INFO router:       Parsing gateway gateway1
[0000]  INFO router:       Starting bridge: whatsapp.mywhatsapp 
13:36:59.744 [Client ERROR] Client outdated (405) connect failure
[0003]  INFO whatsapp:     QR channel result: err-client-outdated
[0005]  INFO whatsapp:     WhatsApp connection successful
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x12f5042]

goroutine 1 [running]:
github.com/42wim/matterbridge/bridge/whatsappmulti.(*Bwhatsapp).Connect(0xc000358fc0)
	/root/go/pkg/mod/github.com/42wim/matterbridge@v1.26.0/bridge/whatsappmulti/whatsapp.go:119 +0x622
github.com/42wim/matterbridge/gateway.(*Router).Start(0xc000108780)
	/root/go/pkg/mod/github.com/42wim/matterbridge@v1.26.0/gateway/router.go:76 +0x35f
main.main()
	/root/go/pkg/mod/github.com/42wim/matterbridge@v1.26.0/matterbridge.go:65 +0x39e

Why have this error ?
I have launch command ./matterbridge -conf matterbridge.toml

fabio841:
solve incredible need update go!!!!!!
First check version. send command
go version
if is old need update.

use: 

git clone https://github.com/udhos/update-golang
cd update-golang
sudo ./update-golang.sh

restart ubuntu

later send command:
go install -tags nomsteams,whatsappmulti github.com/42wim/matterbridge@master

when finish start command:
./matterbridge-1.23.2-linux-64bit -debug -conf matterbridge.toml

QR CODE IS OK!!!!!

mujeebcpy:
i updated everything but still getting error.
```pp..
10:50:11.337 [Client/Socket ERROR] Error reading from websocket: websocket: close 1006 (abnormal closure): unexpected EOF
[0002]  INFO whatsapp:     [Connect:/home/nivadaily/go/pkg/mod/github.com/42wim/matterbridge@v1.26.1-0.20240524210809-d16645c9527a/bridge/whatsappmulti/whatsapp.go:106] QR channel result: timeout
[...]

PR #2177, opened 2024-08-23
Update WhatsApp version number

IntGrah:
Fixes 42wim/matterbridge#2175.

Just update the WhatsApp version number.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/42wim/matterbridge
+module github.com/IntGrah/matterbridge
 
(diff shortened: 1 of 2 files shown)


IntGrah:
Probably just bump the whatsmeow version instead to `go.mau.fi/whatsmeow@v0.0.0-20240821142752-3d63c6fcc1a7` (I don't know how go packages work, but the dependencies seem to be in the `vendors` directory; someone with better go knowledge will know how to bump it)

PR #2189, opened 2024-10-06
Update README.md link for slack-go library

adlai:
From the old nlopes/slack README file:

> You're encouraged to update your import statements to use the official slack-go/slack version above.

It looks like the code was already updated and only this link was overlooked.


DAcodedBEAT (approved):

adlai:
Are the `golangci-lint` and `test-build-upload` really critical for this PR? It only modifies one link, in the human-targeted documentation. Ideally those actions would not be required for such patches, or at least would detect that the patch doesn't affect source code.

PR #2207, opened 2025-01-17
Support EditDisable and EditSuffix for Matrix.

### LINKED ISSUE #2073: Can `EditSuffix` and/or `EditDisable` be supported on the IRC bridge?
Various people on the the IRC channels I'm bridging Matrix to have requested that edits be disabled instead of sending near-duplicate messages.

Ideally `EditDisable` would work with the IRC bridge.

I don't think there are any real alternatives since there seems to be no way to identify an edit on Matrix by the time it reaches the IRC bridge.

**Additional context**
On a whim, I tried adding similar code to that present in the Matrix and Discord bridges, but it appears that `msg.ID` is not set at that point, presumably because IRC doesn't have message IDs. That said, I would have expected the Matrix message ID to be present somewhere, or some other indication that it's an edit from a service that supports them.

divVerent:
Fixes #2073.

### DIFF
--- a/bridge/matrix/matrix.go
+++ b/bridge/matrix/matrix.go
@@ -432,8 +432,13 @@ func (b *Bmatrix) handleEdit(ev *matrix.Event, rmsg config.Message) bool {
 		return false
 	}
 
+	if b.GetBool("EditDisable") {
+		return true
+	}
+
 	rmsg.ID = relation.EventID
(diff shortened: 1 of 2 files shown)


poVoq:
Would you be interested in joining a community managed fork of Matterbridge? See: https://github.com/matterbridge-org/community/issues

PR #2208, opened 2025-01-27
Improve uploads between telegram and XMPP

selfhoster1312:
Previously i would only receive empty messages without the attached images across the bridge.

Now we have a limitation which is that most XMPP clients don't support text alongside images (such as image description or in our case uploader identity). So i opted to post a previous message containing just the username and optional comment.

On telegram side the authorship (and optional comment) is added as caption to the uploaded image.

Sorry the code is ugly i'm a bad bad golang developer.


selfhoster1312:
The XMPP parts have been merged in our community fork https://github.com/matterbridge-org/matterbridge/pull/23
:fireworks:

PR #2211, opened 2025-02-13
discord: Properly paginate through all users

timschumi:
Fetching one page with 1000 users on it is not going to be enough for some servers.


poVoq:
Hi, would you be interested in contributing to a community managed fork of Matterbridge? See: https://github.com/matterbridge-org/community/issues

PR #2226, opened 2025-05-09
check for http.statusOK 200 when downloading

c-holtermann:
When downloading a file only http 200 should be considered as successful. This pull request adds a check for this.


selfhoster1312:
Hello, i just noticed the same problem and opened a PR on our side: https://github.com/matterbridge-org/matterbridge/pull/20

Since then i found your PR here and i thought i'd let you know we accept PRs on that community fork in case you have more fixes like this for others to enjoy :)

PR #2250, opened 2026-02-15
Fix xmpp URI or test chat in README.md

erebion:
This fixes the URI of the example/test chat so that users click it to join.


selfhoster1312:
Hello! Nice catch, but the correct URI for joining a group chat is `xmpp:ADDRESS?join` :-)

It's already fixed in our community fork README: https://github.com/matterbridge-org/matterbridge

erebion:
> Hello! Nice catch, but the correct URI for joining a group chat is `xmpp:ADDRESS?join` :-)

Are you sure the `?join` is really required? I actually tested it. :D

selfhoster1312:
I may be wrong, but i don't think it's defined what the client should do without `?join`. The behavior with `?join` is specified in [XEP-0045 MUC](https://xmpp.org/extensions/xep-0045.html#registrar-querytypes-join) since *at least* 20 years :)

PR #2254, opened 2026-05-01
fix(irc): include kicked nick in KICK relay

vjt:
The system message for `KICK` was constructed the same way as `JOIN`/`PART`/`QUIT` — `<source.Name> <command>s` — so it relayed as e.g. `vjt kicks` with the target nick missing.

This change special-cases `KICK` in a small `formatJoinLeaveText` helper to read `event.Params[1]` (target) and `event.Last()` (reason). Output becomes `vjt kicked alice` (or `vjt kicked alice (reason)`). Other event types unchanged. `verbosejoinpart` mode preserved.

Observed on Azzurra IRC bridging to Telegram, where peers were seeing `vjt kicks` with no target.

A small table-driven test (`bridge/irc/handlers_test.go`) covers JOIN/PART/QUIT, KICK with and without reason, and verbose mode. Happy to adjust if there's a different convention you'd prefer.

### DIFF
--- a/bridge/irc/handlers.go
+++ b/bridge/irc/handlers.go
@@ -103,13 +103,9 @@ func (b *Birc) handleJoinPart(client *girc.Client, event girc.Event) {
 		if b.GetBool("nosendjoinpart") {
 			return
 		}
-		msg := config.Message{Username: "system", Text: event.Source.Name + " " + strings.ToLower(event.Command) + "s", Channel: channel, Account: b.Account, Event: config.EventJoinLeave}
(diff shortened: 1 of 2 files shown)


soliton-:
Note that this repo is dead. Future development is done in https://github.com/matterbridge-org/matterbridge

vjt:
Re-homed to matterbridge-org/matterbridge#240 — 42wim/matterbridge is unmaintained, so moving the change to the actively maintained fork. Closing here.

PR #2258, opened 2026-08-03
Add Slack Socket Mode support and Discord MESSAGE_CONTENT intent

johnnyxmas:
## Summary
Slack is deprecating classic bot tokens/RTM in favor of Socket Mode (events delivered over a websocket, authenticated with an app-level token), and Discord now requires the privileged `MESSAGE_CONTENT` intent for bots to receive message text at all. This adds support for both, plus the follow-on fixes needed to make Socket Mode's event shapes behave correctly:

- **Slack**: implement Socket Mode via `slack-go/slack`'s `socketmode` client, vendoring the `socketmode` and `slackevents` packages (same `slack-go/slack v0.14.0` already in use, just newly using these previously-unvendored subpackages). Enabled automatically when both a bot token (`xoxb-`) and app-level token (`xapp-`) are configured; falls back to the existing webhook path otherwise.
- **Slack**: fix `skipMessageEvent`'s echo/self-message detection for Socket Mode's event shape — the "our own callback ID" check now iterates all top-level and attachment blocks (Slack can inject additional blocks, e.g. `rich_text`, ahead of ours) and also checks `ev.Message.BotID` and `ev.User` against the bot's own ID, since the `BotID` field isn't always populated depending on app configuration.
- **Slack**: fix `skipMessageEvent` to correctly handle `message_changed` events so edits don't re-echo across Discord<->Slack.
- **Slack**: fall back to a user's `RealName` before their handle when `DisplayName` is unset, for more readable relayed usernames.
- **Discord**: add the `IntentMessageContent` privileged intent, required by Discord for bots to receive message content at all.
- **Discord**: fix nick lookup to use Discord's actual display-name priority (server nick > global display name > username) via `Member.DisplayName()`, instead of only checking `Nick`.
- Update README and `matterbridge.toml.sample` with Socket Mode setup instructions (bot + app-level token, required scopes) and the new Discord intent requirement.

## Test plan
- [x] Full build succeeds with `-mod=vendor`
- [x] `go vet` clean
- [x] `gofmt` clean
- [x] Existing `bridge/slack` and `bridge/discord` tests pass
- [x] Run in production against a live Slack workspace + Discord server bridge for several weeks

### DIFF
--- a/README.md
+++ b/README.md
@@ -300,13 +300,14 @@ enable=true
 ```toml
 [slack]
 [slack.test]
-Token="yourslacktoken"
-PrefixMessagesWithNick=true
+Token="xoxb-yourslackbottoken"
+AppToken="xapp-yourapptoken"
 
 [discord]
 [discord.test]
 Token="yourdiscordtoken"
 Server="yourdiscordservername"
+AutoWebhooks=true
 
 [general]
 RemoteNickFormat="[{PROTOCOL}/{BRIDGE}] <{NICK}> "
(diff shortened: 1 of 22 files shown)


johnnyxmas:
Maintainer has been inactive on this repo since August 2024 (over two years), so I'm maintaining these fixes on my own fork (johnnyxmas/matterbridge) instead of waiting on upstream review. Closing this PR.

